PR #5, opened 2020-12-22
Controller

jmprusi:
This PR adds:

* Controller (created with operator-sdk) for reconciling a basic Authorino Service Object.
* A Basic all-in-one Authorino CRD.
* A basic cache for storing the state used by the reconcile & authorino auth service.
* Reconciler tests
* Cache tests
* A Makefile target to set up a local kubernetes environment with all the required components to run/develop/test authorino.

Note: All those changes cause Authorino to "only" work when a kubernetes cluster is available.


.gitignore
```diff
@@ -1,2 +1,23 @@
-tmp
```
guicassolato, line 1: Can we please keep this?

Dockerfile
```diff
@@ -1,21 +1,27 @@
...
+# and so that source changes don't invalidate our downloaded layer
 RUN go mod download
 
-COPY . .
-
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o authorino main.go
+# Copy the go source
+COPY main.go main.go
+COPY api/ api/
+COPY controllers/ controllers/
```
guicassolato, line 15: How about the actual code of Authorino in /pkg? No need to copy it?

Dockerfile
```diff
@@ -1,21 +1,27 @@
...
-COPY --from=build /app/authorino /
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
-EXPOSE 50051
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
+WORKDIR /
+COPY --from=builder /workspace/manager .
```
guicassolato, line 24: Why is it called "manager" instead of the name of the app? Is this a convention?

Dockerfile
jmprusi: That file was generated by the operator-sdk command. I have a pending commit with that part fixed.

Dockerfile
jmprusi: Same as the previous comment :)

controllers/service_controller_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"testing"
+)
+
+var (
+	service = v1beta1.Service{
```
didierofrivia, line 21: Maybe we could introduce in the future something like Factories to create these kind of objects

controllers/service_controller_test.go
```diff
@@ -0,0 +1,222 @@
...
+	}, &serviceCheck)
+
+	// The object we created should remain not ready
+	assert.Check(t, !serviceCheck.Status.Ready)
+
+	// Result should be empty
+	assert.DeepEqual(t, result, ctrl.Result{})
+}
+
+func TestTranslateService(t *testing.T) {
```
didierofrivia, line 220: This one left empty on purpose? Maybe a comment regarding the assertion to be made? or simply remove it

didierofrivia:
Besides being a massive PR, and for most part of the code seems to be auto-generated / vendor... the creation of the CRD thanks to the controller, the cache introduction for the state of configs and the Makefile LGTM! There are some corners that I'm not familiar with yet, so better to discuss them on a different channel. 
Great work!

PR #29, opened 2021-01-11
[wip] Remove OPA endpoint setting

guicassolato:
The option is no longer needed since OPA was integrated as a Go module instead of an external HTTP service.
Reference to OPA as a sidecard container in Kubernetes deployment also fixed.


guicassolato:
Closing in favour of #42.

PR #34, opened 2021-01-15
Concurrent authcontext calls

didierofrivia:
This PR implements a concurrent way of getting Auth objects, instead of getting them in sequential order.

Fixes https://github.com/3scale-labs/authorino/issues/27

Notes:
No tests. A following refactor of the different auth config objects is needed first, in order to avoid converting this into massive refactor PR.

### DIFF
--- a/pkg/service/auth.go
+++ b/pkg/service/auth.go
@@ -38,11 +38,7 @@ func (self *AuthService) Check(ctx context.Context, req *auth.CheckRequest) (*au
 		return self.deniedResponse(rpc.NOT_FOUND, "Service not found"), nil
 	}
 
-	authContext := AuthContext{
-		ParentContext: &ctx,
-		Request:       req,
-		API:           &apiConfig,
-	}
+	authContext := NewAuthContext(ctx, req, apiConfig)
 
 	err = authContext.Evaluate()
 	if err != nil {
--- a/pkg/service/auth.go
+++ b/pkg/service/auth.go
@@ -2,7 +2,6 @@ package service
 
 import (
 	"encoding/json"
-	"log"
 
 	"github.com/3scale-labs/authorino/pkg/cache"
 	"golang.org/x/net/context"
@@ -12,6 +11,11 @@ import (
 	envoy_type "github.com/envoyproxy/go-control-plane/envoy/type"
 	"github.com/gogo/googleapis/google/rpc"
 	rpcstatus "google.golang.org/genproto/googleapis/rpc/status"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var (
+	authServiceLog = ctrl.Log.WithName("Authorino").WithName("AuthService")
 )
 
 // AuthService is the server API for the authorization service.
@@ -26,7 +30,7 @@ func (self *AuthService) Check(ctx context.Context, req *auth.CheckRequest) (*au
 	if err != nil {
 		return self.deniedResponse(rpc.FAILED_PRECONDITION, "Invalid request"), nil
 	}
-	log.Println("[AuthService] Check()\n", string(reqJSON))
+	authServiceLog.Info("Check()", "reqJSON", string(reqJSON))
 
 	// service config
 	host := req.Attributes.Request.Http.Host
--- a/pkg/service/auth_context.go
+++ b/pkg/service/auth_context.go
@@ -4,7 +4,8 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/3scale-labs/authorino/pkg/config"
+	"github.com/3scale/authorino/pkg/config"
+	"github.com/3scale/authorino/pkg/config/internal"
 
 	auth "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
 	"golang.org/x/net/context"
@@ -23,50 +24,53 @@ type AuthContext struct {
 	Authorization map[*config.AuthorizationConfig]interface{}
 }
 
+// AuthObjectConfig provides an interface for APIConfig objects that implements a Call method
+type AuthObjectConfig interface {
+	Call(ctx internal.AuthContext) (interface{}, error)
+}
+
+func (authContext *AuthContext) getAuthObject(authObjConfig AuthObjectConfig, ch chan bool) (interface{}, error) {
+	if ret, err := authObjConfig.Call(authContext); err != nil {
+		return nil, err
+	} else {
+		ch <- true
+		return ret, nil
+	}
+}
+
 // Evaluate evaluates all steps of the auth pipeline (identity → metadata → policy enforcement)
 func (self *AuthContext) Evaluate() error {
-	self.Identity = make(map[*config.IdentityConfig]interface{})
-	self.Metadata = make(map[*config.MetadataConfig]interface{})
-	self.Authorization = make(map[*config.AuthorizationConfig]interface{})
-
 	identityCh, metadataCh, authCh := make(chan bool), make(chan bool), make(chan bool)
 
 	// identity
-
-	for _, c := range self.API.IdentityConfigs {
-		go func() error {
-			if ret, err := c.Call(self); err != nil {
-				return err
-			} else {
-				self.Identity[&c] = ret
-				identityCh <- true
-				return nil
+	for _, config := range self.API.IdentityConfigs {
+		var idAuthObjCfg AuthObjectConfig = &config
+		go func() {
+			if authObj, err := self.getAuthObject(idAuthObjCfg, identityCh); err == nil {
+				self.Identity[&config] = authObj
 			}
 		}()
 	}
 	<-identityCh
 
 	// metadata
-	for _, c := range self.API.MetadataConfigs {
-		go func() error {
-			if ret, err := c.Call(self); err != nil {
-				return err
-			} else {
-				self.Metadata[&c] = ret
-				metadataCh <- true
-				return nil
+	for _, config := range self.API.MetadataConfigs {
+		var idAuthObjCfg AuthObjectConfig = &config
+		go func() {
+			if authObj, err := self.getAuthObject(idAuthObjCfg, identityCh); err == nil {
+				self.Metadata[&config] = authObj
 			}
 		}()
 	}
 	<-metadataCh
 
 	// policy enforcement (authorization)
-	for _, c := range self.API.AuthorizationConfigs {
+	for _, config := range self.API.AuthorizationConfigs {
 		go func() error {
-			if ret, err := c.Call(self); err != nil {
+			if ret, err := config.Call(self); err != nil {
 				return err
 			} else {
-				self.Authorization[&c] = ret
+				self.Authorization[&config] = ret
 				metadataCh <- true
 				return nil
 			}
--- a/pkg/service/auth_context.go
+++ b/pkg/service/auth_context.go
@@ -11,6 +11,11 @@ import (
 
 	auth "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
 	"golang.org/x/net/context"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var (
+	authCxtLog = ctrl.Log.WithName("Authorino").WithName("AuthContext")
 )
 
 // AuthContext holds the context of each auth request, including the request itself (sent by the client),
@@ -33,6 +38,7 @@ type AuthObjectConfig interface {
 
 type configCallback = func(config AuthObjectConfig, obj interface{})
 
+// NewAuthContext creates an AuthContext instance
 func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig config.APIConfig) AuthContext {
 
 	return AuthContext{
@@ -49,11 +55,12 @@ func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig
 func (authContext *AuthContext) getAuthObject(ctx context.Context, objConfig AuthObjectConfig, cb configCallback) error {
 	select {
 	case <-ctx.Done():
-		fmt.Printf("context cancelled objConfig %v terminting\n", objConfig)
+		authCxtLog.Info("Context cancelled objConfig terminating", "config", objConfig)
 		return nil
 	default:
 		if authObj, err := objConfig.Call(authContext); err != nil {
-			return fmt.Errorf("Invalid auth object config %v ", err)
+			authCxtLog.Error(err, "Invalid auth object config")
+			return err
 		} else {
 			cb(objConfig, authObj)
 			return nil
@@ -73,7 +80,7 @@ func (authContext *AuthContext) getAuthObjects(configs []AuthObjectConfig, cb co
 	if err := errGroup.Wait(); err != nil {
 		return err
 	} else {
-		fmt.Println("Successfully fetched all auth objects.")
+		authCxtLog.Info("Successfully fetched all auth objects.", "configs", configs)
 		return nil
 	}
 }
--- a/pkg/service/auth_context.go
+++ b/pkg/service/auth_context.go
@@ -30,92 +30,90 @@ type AuthObjectConfig interface {
 	Call(ctx internal.AuthContext) (interface{}, error)
 }
 
-func (authContext *AuthContext) getAuthObject(authObjConfig AuthObjectConfig) (interface{}, error) {
-	if ret, err := authObjConfig.Call(authContext); err != nil {
-		return nil, err
-	} else {
-		return ret, nil
+type configCallback = func(config AuthObjectConfig, obj interface{})
+
+func (authContext *AuthContext) getAuthObjects(configs []AuthObjectConfig, cb configCallback) []error {
+	var wg sync.WaitGroup
+	var authObjError []error
+	for _, config := range configs {
+		wg.Add(1)
+		go func(objConfig AuthObjectConfig) {
+			defer wg.Done()
+			if authObj, err := objConfig.Call(authContext); err != nil {
+				authObjError = append(authObjError, err) // FIXME: Make error reporting better, fail as soon as we encounter maybe (?)
+				fmt.Errorf("Invalid auth object config %v", err)
+			} else {
+				cb(objConfig, authObj)
+			}
+		}(config)
 	}
+	wg.Wait()
+	return authObjError
 }
 
 // GetIDObject gets an Identity auth object given an Identity config.
-func (authContext *AuthContext) GetIDObject() error {
-	var wg sync.WaitGroup
-	var authObjError error
-	for _, config := range authContext.API.IdentityConfigs {
-		wg.Add(1)
-		var authObjCfg AuthObjectConfig = &config
-		go func() {
-			defer wg.Done()
-			if authObj, err := authContext.getAuthObject(authObjCfg); err != nil {
-				authObjError = err
-				fmt.Errorf("Invalid identity config", err)
-			} else {
-				authContext.Identity[&config] = authObj
-			}
-		}()
+func (authContext *AuthContext) GetIDObject() []error {
+	configs := make([]AuthObjectConfig, len(authContext.API.IdentityConfigs))
+	// Convert []SpecificType to []interfaceType
+	for i, conf := range authContext.API.IdentityConfigs {
+		cpConf := conf
+		configs[i] = &cpConf
 	}
-	wg.Wait()
-	return authObjError
+	return authContext.getAuthObjects(configs,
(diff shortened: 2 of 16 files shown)


pkg/service/auth_context.go
```diff
@@ -5,53 +5,78 @@ import (
 	"strings"
 
 	"github.com/3scale/authorino/pkg/config"
+	"github.com/3scale/authorino/pkg/config/internal"
 
-	"golang.org/x/net/context"
 	auth "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
+	"golang.org/x/net/context"
```
guicassolato, line 11: I had disabled this auto-sort alphabetically thing but I guess not worth fighting against the tide 😞. I'll re-enable mine.

pkg/service/auth_context.go
```diff
@@ -3,54 +3,119 @@ package service
...
+	}
+}
+
+// GetIDObject gets an Identity auth object given an Identity config.
+func (authContext *AuthContext) GetIDObject() error {
+	var wg sync.WaitGroup
+	var authObjError error
+	for _, config := range authContext.API.IdentityConfigs {
+		wg.Add(1)
+		var authObjCfg AuthObjectConfig = &config
```
eguzki, line 47: This is wrong: http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#closure_for_it_vars

the iteration variable `config` will always have the same memory address.

pkg/service/auth_context.go
```diff
@@ -3,54 +3,119 @@ package service
...
-	Identity map[*config.IdentityConfig] interface{}
-	Metadata map[*config.MetadataConfig] interface{}
-	Authorization map[*config.AuthorizationConfig] interface{}
+// AuthObjectConfig provides an interface for APIConfig objects that implements a Call method
+type AuthObjectConfig interface {
+	Call(ctx internal.AuthContext) (interface{}, error)
+}
+
+func (authContext *AuthContext) getAuthObject(authObjConfig AuthObjectConfig) (interface{}, error) {
+	if ret, err := authObjConfig.Call(authContext); err != nil {
```
eguzki, line 34: I think this method does not add any value. In any case, you could just return the value of the called method
```
func (authContext *AuthContext) getAuthObject(authObjConfig AuthObjectConfig) (interface{}, error) {  
    return authObjConfig.Call(authContext)                                                                                                 
}                                                                                                     
```

eguzki (changes requested):

pkg/service/auth_context.go
```diff
@@ -3,54 +3,119 @@ package service
...
+func (authContext *AuthContext) GetIDObject() error {
+	var wg sync.WaitGroup
+	var authObjError error
+	for _, config := range authContext.API.IdentityConfigs {
+		wg.Add(1)
+		var authObjCfg AuthObjectConfig = &config
+		go func() {
+			defer wg.Done()
+			if authObj, err := authContext.getAuthObject(authObjCfg); err != nil {
+				authObjError = err
```
eguzki, line 51: Using a shared variable to store the error is wrong. There may be concurrency issues and one "thread" could overwrite existing error to a `nil`.

Error handling in the goroutine context is done either by:
* Using an error channel where goroutines push errors and the main thread, after waiting for all to finish, checks for errors
* Use [errgroup](https://godoc.org/golang.org/x/sync/errgroup)

eguzki (changes requested):

pkg/service/auth_context.go
```diff
@@ -3,54 +3,119 @@ package service
...
 type AuthContext struct {
 	ParentContext *context.Context
-	Request *auth.CheckRequest
-	API *APIConfig
+	Request       *auth.CheckRequest
+	API           *APIConfig
+
+	Identity      map[*config.IdentityConfig]interface{}
+	Metadata      map[*config.MetadataConfig]interface{}
+	Authorization map[*config.AuthorizationConfig]interface{}
```
eguzki, line 25: I would never use struct pointers as map keys. you can have may unexpected issues.

Instead, I would try to generate some kind of ID out of the structure. I would implement `ID` method in the struct and come up with some ID.

```
func (self *IdentityConfig) ID() string {
      // compute ID
}
```
```

File: pkg/service/auth_context.go [hunk 6]
```diff
@@ -3,54 +3,119 @@ package service
...
+	} else {
+		return ret, nil
+	}
+}
+
+// GetIDObject gets an Identity auth object given an Identity config.
+func (authContext *AuthContext) GetIDObject() error {
+	var wg sync.WaitGroup
+	var authObjError error
+	for _, config := range authContext.API.IdentityConfigs {
```
eguzki, line 45: I guess the length of `IdentityConfigs` is reasonably low. Up to few tens of configs. If you expect to scale to hundreds or thousands, then consider limiting concurrency to N elements. It can be implemented using Buffered Channels (channels with max capacity) easily.

pkg/service/auth_context.go
didierofrivia: yeah, this one got slim after last refactor... it doesn't make sense now, thanks!

pkg/service/auth_context.go
```diff
@@ -3,54 +3,119 @@ package service
...
 // the auth config of the requested API and the lists of identity verifications, metadata add-ons and
 // authorization policies, and their corresponding results after evaluated
 type AuthContext struct {
 	ParentContext *context.Context
-	Request *auth.CheckRequest
-	API *APIConfig
+	Request       *auth.CheckRequest
+	API           *APIConfig
+
+	Identity      map[*config.IdentityConfig]interface{}
```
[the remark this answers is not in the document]
didierofrivia: A valid point, I'll bring this up with the team. This will be evolving, it still keeps the POC design

pkg/service/auth_context.go
didierofrivia: I'll comment this with the team, we might need to think further the context data struct.

eguzki:
Looking good. 

Why do you commit vendor deps? `go.mod` is enough to get reproducible builds. Vendoring is mainly used on networking restricted environments to have reproducible builds.

pkg/service/auth_context.go
```diff
@@ -23,44 +27,125 @@ type AuthContext struct {
...
+	return authContext.getAuthObjects(configs,
+		func(conf AuthObjectConfig, authObj interface{}) {
+			// Caution: type assertion not checked
+			v, _ := conf.(*config.IdentityConfig)
+			authContext.Identity[v] = authObj
+		})
+}
+
+// GetMDObject gets a Metadata auth object given a Metadata config.
+func (authContext *AuthContext) GetMDObject() error {
```
guicassolato, line 103:
```suggestion
func (authContext *AuthContext) GetMetadataObject() error {
```

pkg/service/auth_context.go
```diff
@@ -23,44 +27,125 @@ type AuthContext struct {
...
+	if err := errGroup.Wait(); err != nil {
+		return err
+	} else {
+		log.Info("Successfully fetched all auth objects.", configs)
+		return nil
+	}
+}
+
+// GetIDObject gets an Identity auth object given an Identity config.
+func (authContext *AuthContext) GetIDObject() error {
```
guicassolato, line 87:
```suggestion
func (authContext *AuthContext) GetIdentityObject() error {
```

pkg/service/auth_context.go
```diff
@@ -23,44 +27,125 @@ type AuthContext struct {
...
+		Request:       req,
+		API:           &apiConfig,
+		Identity:      make(map[*config.IdentityConfig]interface{}),
+		Metadata:      make(map[*config.MetadataConfig]interface{}),
+		Authorization: make(map[*config.AuthorizationConfig]interface{}),
+	}
+
+}
+
+func (authContext *AuthContext) getAuthObject(ctx context.Context, objConfig AuthObjectConfig, cb configCallback) error {
```
guicassolato, line 51:
```suggestion
func (authContext *AuthContext) getAuthorizationObject(ctx context.Context, objConfig AuthObjectConfig, cb configCallback) error {
```

pkg/service/auth_context.go
```diff
@@ -23,44 +27,125 @@ type AuthContext struct {
...
+	return authContext.getAuthObjects(configs,
+		func(conf AuthObjectConfig, authObj interface{}) {
+			// Caution: type assertion not checked
+			v, _ := conf.(*config.MetadataConfig)
+			authContext.Metadata[v] = authObj
+		})
+}
+
+// GetAuthObject gets an Authorization object given an Authorization config.
+func (authContext *AuthContext) GetAuthObject() error {
```
guicassolato, line 119:
```suggestion
func (authContext *AuthContext) GetAuthorizationObject() error {
```

pkg/service/auth_context.go
```diff
@@ -23,44 +27,125 @@ type AuthContext struct {
...
-	self.Metadata = make(map[*config.MetadataConfig]interface{})
-	self.Authorization = make(map[*config.AuthorizationConfig]interface{})
-
-	// identity (authentication)
-	identityConfigs := self.API.IdentityConfigs
-	for i := range identityConfigs {
-		c := identityConfigs[i]
-		if ret, err := c.Call(self); err != nil {
+// AuthObjectConfig provides an interface for APIConfig objects that implements a Call method
+type AuthObjectConfig interface {
```
guicassolato, line 31: I was thinking... Maybe instead of "object" we could name these something that helps one to understand better the purpose, like "AuthConfigExecutor" or "AuthConfigEvaluator" or (simpler) "AuthEvalutor".

WDYT?

pkg/service/auth_context.go
guicassolato, line 51: Also thinking about a better naming for "AuthObject" here 🤔 ... Not sure yet. "AuthPartial" perhaps? It could also be "AuthResponse" even though it's not the full response to the auth request just yet. IDK.

pkg/service/auth_context.go
```diff
@@ -23,44 +27,125 @@ type AuthContext struct {
...
+		API:           &apiConfig,
+		Identity:      make(map[*config.IdentityConfig]interface{}),
+		Metadata:      make(map[*config.MetadataConfig]interface{}),
+		Authorization: make(map[*config.AuthorizationConfig]interface{}),
+	}
+
+}
+
+func (authContext *AuthContext) getAuthObject(ctx context.Context, objConfig AuthObjectConfig, cb configCallback) error {
+	log := ctrl.Log.WithName("Authorino").WithName("AuthContext")
```
guicassolato, line 52: I think this could be in `func NewAuthContext`, so it doesn't have to be repeated in the other functions.

pkg/service/auth_context.go
```diff
@@ -23,44 +27,125 @@ type AuthContext struct {
...
+	select {
+	case <-ctx.Done():
+		log.Info("Context cancelled objConfig %v terminting", objConfig)
```
guicassolato, line 55:
```suggestion
		log.Info("Context cancelled objConfig %v terminating", objConfig)
```

pkg/service/auth_context.go
guicassolato: After reading again, I have another suggestion for the name of this function (and analogously extensible to the other 2 below):

```suggestion
func (authContext *AuthContext) EvaluateIdentity() error {
```

...because this is about splitting `Evaluate()` into the part that is specific about identity, right? `GetWhateverObject` doesn't say anything to me really.

pkg/service/auth_context.go
guicassolato, line 103:
```suggestion
func (authContext *AuthContext) EvaluateMetadata() error {
```

See https://github.com/3scale-labs/authorino/pull/34/files?file-filters%5B%5D=.go#r567739524

pkg/service/auth_context.go
guicassolato, line 119:
```suggestion
func (authContext *AuthContext) EvaluateAuthorization() error {
```

See https://github.com/3scale-labs/authorino/pull/34/files?file-filters%5B%5D=.go#r567739524

pkg/service/auth.go
```diff
@@ -38,11 +39,7 @@ func (self *AuthService) Check(ctx context.Context, req *auth.CheckRequest) (*au
 		return self.deniedResponse(rpc.NOT_FOUND, "Service not found"), nil
 	}
 
-	authContext := AuthContext{
-		ParentContext: &ctx,
-		Request:       req,
-		API:           &apiConfig,
-	}
+	authContext := NewAuthContext(ctx, req, apiConfig)
```
guicassolato, line 42: Following on https://github.com/3scale-labs/authorino/pull/34/files?file-filters%5B%5D=.go#r567730541...

```suggestion
	authContext := NewAuthContext(ctx, req, apiConfig, log)
```

pkg/service/auth_context.go
```diff
@@ -23,44 +27,125 @@ type AuthContext struct {
...
+}
+
+// GetIDObject gets an Identity auth object given an Identity config.
+func (authContext *AuthContext) GetIDObject() error {
+	configs := make([]AuthObjectConfig, len(authContext.API.IdentityConfigs))
+	// Convert []SpecificType to []interfaceType
+	for i, conf := range authContext.API.IdentityConfigs {
+		cpConf := conf
+		configs[i] = &cpConf
+	}
```
guicassolato, line 93: I wish we had another way to avoid this conversion (here and in the other analogous steps below), but I guess there's nothing we can do about it... part of the hurdles of using Golang perhaps. @jmprusi @eguzki, any ideas?

eguzki: IMO. this is a bad smell coming from the data structure design. It should be reviewed and designed better with interfaces. I ti would allow easier unittesting as well.

didierofrivia: I'll try to see how could it be done in my following refactor PRs

pkg/service/auth_context.go
didierofrivia: I think this belongs to a different context altogether, will shoot another PR to solve it.

didierofrivia: Actually, I'll push a good alternative here

pkg/service/auth.go
didierofrivia: Better to keep logging separated from the auth context

pkg/service/auth_context.go
didierofrivia: You're right, it makes better sense along the changes you propose renaming to "evaluators" instead of "getters"

pkg/service/auth_context.go
didierofrivia: I think I kind of have a good fix for this, will commit it in this PR

pkg/service/auth_context.go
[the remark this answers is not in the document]
guicassolato: The value might be needed ahead, in the authorization step. Basically whatever object is returned in Identity and Metadata will be available for Authorization, thus allowing one to write authorization policies that take into account values known to be resolved in the previous steps.

pkg/service/auth_context.go
guicassolato: That's my fault. It was inherited from when I tried to rewrite in Golang as faithful as possible to what we had before in Ruby. In Ruby, the keys of the map (`Hash`) are the actual config objects. In Golang that wouldn't be possible, so I used the pointers.

...and BTW could you please give an example of when doing that would become an issue? Not that I am against improving it obviously; just wondering. E.g. can a declared variable move position in memory in Golang? Because that would be an example of this becoming a problem.

pkg/service/auth_context.go
guicassolato: The length shouldn't even reach _one_ then of config, otherwise what the hell of an identity broker we'd have 🙂

pkg/service/auth.go
```diff
@@ -26,7 +30,7 @@ func (self *AuthService) Check(ctx context.Context, req *auth.CheckRequest) (*au
 	if err != nil {
 		return self.deniedResponse(rpc.FAILED_PRECONDITION, "Invalid request"), nil
 	}
-	log.Println("[AuthService] Check()\n", string(reqJSON))
+	authServiceLog.Info("Check()", "reqJSON", string(reqJSON))
```
guicassolato, line 33: IDK if works, but could we go straight to the value of request as JSON and skip the "reqJSON" tag?

```suggestion
	authServiceLog.Info("Check()", string(reqJSON))
```

didierofrivia: Sadly nope, the func signature expects a `msg` of type `string` then a key/value pairs as context.

> The msg argument should be used to add some constant description to the log line. The key/value pairs can then be used to add additional variable information. The key/value pairs should alternate string keys and arbitrary values

pkg/service/auth_context.go
```diff
@@ -23,99 +31,137 @@ type AuthContext struct {
...
+}
+
+// Evaluate evaluates all steps of the auth pipeline (identity → metadata → policy enforcement)
+func (authContext *AuthContext) Evaluate() error {
+	// identity
+	if err := authContext.evaluateAuthConfigs(authContext.API.IdentityConfigs,
+		func(conf common.AuthEvaluator, authObj interface{}) {
+			v, _ := conf.(*config.IdentityConfig)
+			authContext.Identity[v] = authObj
```
guicassolato, line 89: Something is broken here. We are no longer storing in the `authContext`'s maps the resulting object returned by each evaluator. Instead, the map seems now to be storing a copy of the config.

From the logs, this is the tail of "[OPA] input" for the example request to `/pets/1` with an OIDC token issued to user John.

```
[...], "metadata":{"uma":{"UserInfo":null,"UMA":{"Endpoint":"http://keycloak:8080/auth/realms/ostia","ClientID":"pets-api","ClientSecret":"523b92b6-625d-4e1e-a313-77e7a8ae4e88"}},"userinfo":{"UserInfo":{"OIDC":"keycloak","ClientID":"authorino","ClientSecret":"2e5246f2-f4ef-4d55-8225-36e725071dee"},"UMA":null}}}}
```

The "metadata" field of the JSON above should be of the form ```{ "userinfo": { the actual claims of the user John returned by keycloak }, "uma": { the actual data of the resource whose ID is 1 returned by keycloak } }```. Adding instead respectively the configs for fecthing the userinfo and the resource data is incorrect.

pkg/service/auth_context.go
```diff
@@ -23,99 +31,137 @@ type AuthContext struct {
...
-		c := metadataConfigs[i]
-		if ret, err := c.Call(self); err != nil {
-			return err
-		} else {
-			self.Metadata[&c] = ret
-		}
+	if err := authContext.evaluateAuthConfigs(authContext.API.MetadataConfigs,
+		func(conf common.AuthEvaluator, authObj interface{}) {
+			v, _ := conf.(*config.MetadataConfig)
+			authContext.Metadata[v] = authObj
```
guicassolato, line 98: Something is broken here. We are no longer storing in the `authContext`'s maps the resulting object returned by each evaluator. Instead, the map seems now to be storing a copy of the config.

From the logs, this is the tail of "[OPA] input" for the example request to `/pets/1` with an OIDC token issued to user John.

```
[...], "metadata":{"uma":{"UserInfo":null,"UMA":{"Endpoint":"http://keycloak:8080/auth/realms/ostia","ClientID":"pets-api","ClientSecret":"523b92b6-625d-4e1e-a313-77e7a8ae4e88"}},"userinfo":{"UserInfo":{"OIDC":"keycloak","ClientID":"authorino","ClientSecret":"2e5246f2-f4ef-4d55-8225-36e725071dee"},"UMA":null}}}}
```

The "metadata" field of the JSON above should be of the form ```{ "userinfo": { the actual claims of the user John returned by keycloak }, "uma": { the actual data of the resource whose ID is 1 returned by keycloak } }```. Adding instead respectively the configs for fecthing the userinfo and the resource data is incorrect.

guicassolato (changes requested):

pkg/service/auth_context.go
didierofrivia: Taking a closer look into this. Thanks for reporting!

didierofrivia: Fixed :) Thanks for your help :)

guicassolato (approved):
Nice work, @didierofrivia! I've tested it too again and it worked.

When I first noticed https://github.com/3scale-labs/authorino/pull/34/files#r572380595 I worried about some more serious issue after the refactoring, but it turns out it was just that silly thing that you spotted in the `AuthContext.GetMetadata()` method. Nice catch and thank you!

Thank you as well for fixing sooooo many other things, apart from just making authcontext calls concurrent :-)

eguzki (approved):

PR #35, opened 2021-01-19
enhancement concurrent authcontext calls

eguzki:
Proposal to avoid DRY for concurrent authcontext calls. Working out further your data structures and working with interfaces may improve further this implementation. But with the current data structures, this is the best I could come up with.

IMO, right now, the additional structure copies and the deteriorated code legibility does not pay off. But up to you. With better data structures, it can be interesting. 

PS: Sorry about all the formatting "noise". That is the output of `go fmt`, which is the standard one. My IDE executes `go fmt` every now and then. Focus on the refactor of `GetIDObject`, `GetMDObject` and `GetAuthObject`.

### DIFF
--- a/pkg/service/auth_context.go
+++ b/pkg/service/auth_context.go
@@ -5,8 +5,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/3scale/authorino/pkg/common"
 	"github.com/3scale/authorino/pkg/config"
-	"github.com/3scale/authorino/pkg/config/internal"
 
 	auth "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
 	"golang.org/x/net/context"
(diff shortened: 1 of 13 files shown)


didierofrivia:
Great! That's what I was trying with the callback, but couldn't make it work with types! 
If you don't mind, I could cherry pick only the auth_context, and leave the linting and renaming of the `internal` module that are a bit out of scope for this PR. Most probably these last changes will be part of another one ;)
Thanks again!

pkg/service/auth_context.go
```diff
@@ -27,78 +27,76 @@ type AuthContext struct {
...
 }
 
+// GetIDObject gets an Identity auth object given an Identity config.
+func (authContext *AuthContext) GetIDObject() error {
+	configs := make([]AuthObjectConfig, len(authContext.API.IdentityConfigs))
+	// Convert []SpecificType to []interfaceType
+	for i, conf := range authContext.API.IdentityConfigs {
+		cpConf := conf
+		configs[i] = &cpConf
+	}
```
didierofrivia, line 61: This can't be like...

```suggestion
	for i, conf := range authContext.API.IdentityConfigs {
		configs[i] = &conf
	}
```

?

eguzki: No, you cannot do that. You need to copy the data: https://github.com/3scale-labs/authorino/pull/34#discussion_r560129840

eguzki: anyway, having a map with type pointers as keys has a bad smell that needs reviewing

didierofrivia:
Thanks for the work and suggestions! closing this, will link to keep the changes and comments handy for the next refactors

PR #40, opened 2021-02-08
Github Action for Golang unit tests.

jmprusi:


guicassolato (dismissed):
Screenshot 2021-02-09 at 18 10 29

Now all we need to do is to actually add tests 😆

jmprusi:
> Screenshot 2021-02-09 at 18 10 29
> 
> Now all we need to do is to actually add tests 😆

It is actually running make test.. :) 

image

I guess i should improve that, so it shows "make test" instead of "set -x"

guicassolato:
It looks like some checks moved from "Code Style" to "Test". Is this expected, @jmprusi?

Screenshot 2021-02-09 at 18 26 52

jmprusi:
@guicassolato  honestly, no idea why that is happening...

guicassolato (approved):

PR #41, opened 2021-02-10
Testing AuthContext

didierofrivia:
This is a following PR of #34 

It also introduces a fix, similar to https://github.com/3scale-labs/authorino/pull/45

Notes: 
Don't pay attention at the moment to the mocked responses in the tests, those will be used in future tests.

### DIFF
--- a/pkg/service/auth_context.go
+++ b/pkg/service/auth_context.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -15,7 +16,7 @@ import (
 )
 
 var (
-	authCxtLog = ctrl.Log.WithName("Authorino").WithName("AuthContext")
+	authCtxLog = ctrl.Log.WithName("Authorino").WithName("AuthContext")
 )
 
 // AuthContext holds the context of each auth request, including the request itself (sent by the client),
@@ -31,7 +32,7 @@ type AuthContext struct {
 	Authorization map[*config.AuthorizationConfig]interface{}
 }
 
-type configCallback = func(config common.AuthConfigEvaluator, obj interface{})
+type evaluateCallback = func(config common.AuthConfigEvaluator, obj interface{}, err error) error
 
 // NewAuthContext creates an AuthContext instance
 func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig config.APIConfig) AuthContext {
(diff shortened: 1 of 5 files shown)


pkg/service/auth_context.go
```diff
@@ -47,71 +48,121 @@ func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig
...
-		func(conf common.AuthConfigEvaluator, authObj interface{}) {
+	successChannel := make(chan bool)
+
+	go func() {
+		defer close(successChannel)
+		authContext.evaluateAuthConfigs(identityConfigs,
+			func(conf common.AuthConfigEvaluator, cancelCtx context.CancelFunc, authObj interface{}, err error) error {
+				if err != nil {
+					authCtxLog.Error(err, "error evaluating identity config", "config", conf)
+					return nil
```
guicassolato, line 95: So when there's an error, we return no error? because in this case the error should not be blocking, right?

pkg/service/auth_context.go
```diff
@@ -47,71 +48,121 @@ func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig
...
+	metadataConfigs := authContext.API.MetadataConfigs
+	if len(metadataConfigs) <= 0 {
+		authCtxLog.Info("No metadata configs found", "Configs", metadataConfigs)
+		return
+	}
+	authContext.evaluateAuthConfigs(metadataConfigs,
+		func(conf common.AuthConfigEvaluator, cancelCtx context.CancelFunc, authObj interface{}, err error) error {
+			if err != nil {
+				authCtxLog.Error(err, "error evaluating metadata config", "config", conf)
+				return nil
```
guicassolato, line 124: Same as https://github.com/3scale-labs/authorino/pull/41/files#r576698099

pkg/service/auth_context.go
```diff
@@ -47,71 +48,121 @@ func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig
...
-	if err := authContext.evaluateAuthConfigs(authContext.API.AuthorizationConfigs,
-		func(conf common.AuthConfigEvaluator, authObj interface{}) {
+// EvaluateAuthorization evaluates the authorization config, along with any metadata and identity and authorizes it.
+func (authContext *AuthContext) EvaluateAuthorization() error {
+	authorizationConfigs := authContext.API.AuthorizationConfigs
+	if len(authorizationConfigs) <= 0 {
+		return errors.New("no authorization configs found")
+	}
+	return authContext.evaluateAuthConfigs(authContext.API.AuthorizationConfigs,
+		func(conf common.AuthConfigEvaluator, cancelCtx context.CancelFunc, authObj interface{}, err error) error {
```
guicassolato, line 141: Apparently the reason for this function to return an `error` now is only because of the evaluation of the autorization configs, where an error is blocking and should close the conext. Why not making this function a routine that does not return anything and use `cancelCtx` here?

guicassolato (commented):
@didierofrivia, to me it almost looks like we are adapting our code so we can keep using https://pkg.go.dev/golang.org/x/sync@v0.0.0-20201207232520-09787c993a3a/errgroup, which may add some abstraction to the control of concurrency but doesn't really help much in terms of readability IMO; apart from being another dependency.

I also see some coupling between the different policies of error treatment and each step of the pipeline. Those things could be slightly separate and maybe allow us in the future to add a config in the middle to offer that control as an option to the user. E.g. in some cases, treat identity evaluation as _all-or-nothing_ instead of _at-least-one_, or treat metadata as _at-least-one_ instead _any/optional_ and so on.

Honestly, I would favour https://github.com/3scale-labs/authorino/pull/45, which I believe it's more straightforward and doesn't add any extra dependency. Let's wait for more opinions and decide.

pkg/service/auth_context.go
didierofrivia: non blocking error, exactly

pkg/service/auth_context.go
didierofrivia: yup, the same, it seems non blocking is kind of default with the new behaviour

guicassolato: The "new" part is on me. It is something that was obvious to me from start but only actually realised that the code was not right (and it wasn't even before concurrency was introduced!) when I started working on https://github.com/3scale-labs/authorino/pull/46.

pkg/service/auth_context.go
```diff
@@ -47,71 +48,122 @@ func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig
 
 }
 
-func (authContext *AuthContext) evaluateAuthConfig(ctx context.Context, config common.AuthConfigEvaluator, cb configCallback) error {
+func (authContext *AuthContext) evaluateAuthConfig(ctx context.Context, config common.AuthConfigEvaluator, cb evaluateCallback) error {
 	select {
```
eguzki, line 52: What is the purpose of this select? 

It will not cancel the job if any other job returns error.

pkg/service/auth_context.go
```diff
@@ -47,71 +48,122 @@ func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig
...
 	}
 
-	// metadata
-	if err := authContext.evaluateAuthConfigs(authContext.API.MetadataConfigs,
-		func(conf common.AuthConfigEvaluator, authObj interface{}) {
+	ctx, cancelCtx := context.WithCancel(context.Background())
+
+	successChannel := make(chan bool)
+
+	go func() {
```
eguzki, line 90: `authContext.evaluateAuthConfigs` will create goroutines, why create new goroutine only to wait for success?

pkg/service/auth_context.go
```diff
@@ -47,71 +48,122 @@ func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig
...
+		authContext.evaluateAuthConfigs(ctx, identityConfigs,
+			func(conf common.AuthConfigEvaluator, authObj interface{}, err error) error {
+				if err != nil {
+					authCtxLog.Error(err, "error evaluating identity config", "config", conf)
+					return nil // Non blocking error
+				}
+				// Convert from interfaceType to SpecificType
+				v, _ := conf.(*config.IdentityConfig)
+				authCtxLog.Info("Identity", "Config", conf, "AuthObj", authObj)
+				authContext.Identity[v] = authObj
```
eguzki, line 101: this data structure is, potentially, going to be updated concurrently by several goroutines. Needs to be protected

eguzki (changes requested):

pkg/service/auth_context.go
```diff
@@ -47,71 +48,122 @@ func NewAuthContext(parentCtx context.Context, req *auth.CheckRequest, apiConfig
...
 
-	// metadata
-	if err := authContext.evaluateAuthConfigs(authContext.API.MetadataConfigs,
-		func(conf common.AuthConfigEvaluator, authObj interface{}) {
+	ctx, cancelCtx := context.WithCancel(context.Background())
+
+	successChannel := make(chan bool)
+
+	go func() {
+		defer close(successChannel)
```
eguzki, line 91: this is very risky. You could be closing some channel and after that try to read it in the main goroutine

pkg/service/auth_context.go
eguzki: I see, this is for "early finish". 

I am not sure this is the correct way to do this. 

If the approach is err == nil is evaluation ok and err != nil means evaluation failed, the errgroup approach does not seem to be the best fit for this purpose. In understand that overall evaluation succeeds when at least one evaluation succeeds.

didierofrivia:
Closing this PR, favouring possibly https://github.com/3scale-labs/authorino/pull/49. Keeping the `errGroup` with the new desired behaviour does not make sense.

PR #44, opened 2021-02-11
make local-deploy

guicassolato:
Make target to ease rebuilding and redeploying Authorino to the local Kubernetes cluster

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,3 +145,8 @@ endif
 .PHONY: local-setup
 local-setup: kustomize kind
 	utils/local-setup.sh
+
+# Rebuild and push the docker image and redeploy authorino to the local k8s cluster
+.PHONY: local-deploy
+local-deploy: kind
+	utils/local-deploy.sh
--- a/utils/local-deploy.sh
+++ b/utils/local-deploy.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -euo pipefail
+
+export AUTHORINO_NAMESPACE="authorino"
+export KIND_CLUSTER_NAME="authorino-integration"
+
+echo "Building Authorino"
+docker build -t authorino:devel ./
+kind load docker-image authorino:devel --name ${KIND_CLUSTER_NAME}
+
+echo "Deploying Authorino"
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -145,3 +145,8 @@ endif
 .PHONY: local-setup
 local-setup: kustomize kind
 	utils/local-setup.sh
+
+# Rebuild and push the docker image and redeploy authorino to the local k8s cluster
+.PHONY: local-deploy
```
guicassolato, line 150: Happy to rename it in case you think it's confusing with `local-setup`.

utils/local-deploy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -euo pipefail
+
+export AUTHORINO_NAMESPACE="authorino"
+export KIND_CLUSTER_NAME="authorino-integration"
+
+echo "Building Authorino"
+docker build -t authorino:devel ./
+kind load docker-image authorino:devel --name ${KIND_CLUSTER_NAME}
```
guicassolato, line 10: I know. Some DRY opportunity here: https://github.com/3scale-labs/authorino/blob/3cae60bd3305ec98e1d1f3c4664c040d0c731042/utils/local-setup.sh#L10-L12

Makefile
didierofrivia: Let's review both tasks after some use

didierofrivia:
You might want to add the task description to readme? maybe we could include a separate Development.md file

Makefile
guicassolato: `make local-image-reload`?

guicassolato: `make local-image`?

jmprusi: what about `make build-to-kind` ? 

This script will end up, building the images and loading them into the local Kind cluster, not only making images.

guicassolato: > what about make build-to-kind ?

I think better keeping it consistent with the other one, `local-setup`. It seems that "local" already means "to kind".

didierofrivia (approved):
LGTM, let's merge to aid development, and then we can add steps to the readme along with any improvement

PR #45, opened 2021-02-13
Fix identity and metadata config evaluation context

guicassolato:
Unlike the authorization step, the evaluation of identity configs does not require all identity sources to return a valid response. Only one identity source configured is required to verify the identity of the requestor (upon token/credentials presented) for this step to be considered successful and, once that happens, the concurrent context can be closed – i.e. the go routine that succeeds should tell the others that the evaluation is already done.

Metadata config evaluation on the other hand should require no config to be successful to continue.

This PR fixes these two behaviours described, while keeping the current behaviour of the authorization config evaluation step – i.e. all configs or nothing for this one.


pkg/service/auth_context.go
```diff
@@ -63,53 +62,119 @@ func (authContext *AuthContext) evaluateAuthConfig(ctx context.Context, config c
...
-		})
+		go func() {
+			defer waitGroup.Done()
+
+			err := authContext.evaluateAuthConfig(ctx, objConfig, cb)
+			if err != nil {
+				errorChannel <- err
+			} else {
+				successChannel <- true
+				cancel() // cancels the context if at least one thread succeeds
```
eguzki, line 90: this marks the context as cancelled. Are other jobs checking if the context is cancelled?

pkg/service/auth_context.go
```diff
@@ -63,53 +62,119 @@ func (authContext *AuthContext) evaluateAuthConfig(ctx context.Context, config c
...
+			}
+		}()
 	}
-	if err := errGroup.Wait(); err != nil {
-		return err
-	} else {
-		authCxtLog.Info("Successfully fetched all auth objects.", "authConfigs", authConfigs)
+
+	success := <-successChannel
```
eguzki, line 95: With this implementation:
* If at least one evaluation succeeds,  other potential errors are not taken into account
* If all evaluations fail, only one err is reported.

This is the expected behavior?

pkg/service/auth_context.go
```diff
@@ -63,53 +62,119 @@ func (authContext *AuthContext) evaluateAuthConfig(ctx context.Context, config c
...
+	size := len(authConfigs)
+	waitGroup := new(sync.WaitGroup)
+	waitGroup.Add(size)
+
+	for _, authConfig := range authConfigs {
+		objConfig := authConfig
+		go func() {
+			defer waitGroup.Done()
+
+			_ = authContext.evaluateAuthConfig(ctx, objConfig, cb)
```
eguzki, line 146: evaluation success == (err == nil)
evaluation failed == (err != nil)

Shouldn't be different reporting? like :  
a) evaluation: true/false
b) err (if some unexpected error happens)

pkg/service/auth_context.go
guicassolato: Yes. Whenever there are more than one config of the same auth step (i.e. more than one identify config, more than one metadata config, more than one authorization config), those will run concurrently to each other, and line 51 of this same file incurs of signalling to other when the context has been cancelled.

pkg/service/auth_context.go
guicassolato: This is exactly the expected behaviour and what "OneAuthConfig" is supposed to mean.

pkg/service/auth_context.go
guicassolato: I was talking about that with @didierofrivia this morning. Having a non-boolean response as such would be an alternative and a better fit to using https://pkg.go.dev/golang.org/x/sync@v0.0.0-20201207232520-09787c993a3a/errgroup again. We would need to change all evaluators though and give them this new semantics. I'd suggest doing it as an improvement/refactoring in a future iteration. Right now, this is more about fixing behaviour though.

pkg/service/auth_context.go
guicassolato: _Correction:_ more than one err may be reported. However, it doesn't matter. What matters is that if at least one success is reported, it's good enough to continue.

didierofrivia (approved):
Behavioural wise, it's good, there are some stuff to improve, but I think it's OK to leave it for a following or future refactor. 🎖️ 

If @eguzki thinks the use of channels in the way proposed is good/safe, I think it's good to go

guicassolato:
> Behavioural wise, it's good, there are some stuff to improve, but I think it's OK to leave it for a following or future refactor. 🎖️
> 
> If @eguzki thinks the use of channels in the way proposed is good/safe, I think it's good to go

Agreed, @didierofrivia. And in fact that one comment from @eguzki about unsafe writing to the map is critical and requires a fix ASAP. He actually commented in #41. I've tested with `-race` flag and Golang does mark it as a problem.

```
==================
WARNING: DATA RACE
Write at 0x00c00025a450 by goroutine 105:
  runtime.mapassign_fast64()
      /usr/local/go/src/runtime/map_fast64.go:92 +0x0
  github.com/3scale-labs/authorino/pkg/service.(*AuthContext).Evaluate.func3()
      /Users/guilherme.cassolato/ws/authorino/pkg/service/auth_context.go:179 +0x1f9
  github.com/3scale-labs/authorino/pkg/service.(*AuthContext).evaluateAuthConfig()
      /Users/guilherme.cassolato/ws/authorino/pkg/service/auth_context.go:60 +0x499
  github.com/3scale-labs/authorino/pkg/service.(*AuthContext).evaluateAllAuthConfigs.func2()
      /Users/guilherme.cassolato/ws/authorino/pkg/service/auth_context.go:124 +0xd0
```

Although I believe the issue wouldn't ever be activated in production due to the way we are using the map (i.e. different indexes in each thread and storing pointers in the value), it is still risky in the long run. #49 will take care of it.

Another one to tackle in the future maybe is the semantics of returning an error.

Thanks.

PR #47, opened 2021-02-15
Upgrade Envoy ext authz implementation to v3 API

### LINKED ISSUE #4: Upgrade envoy ext auth to v3
Authorino's current implementation is based on the deprecated version 2 of the Envoy's ext authz gRPC API, forcing Envoy to be bootstrapped in backward compatibility mode: https://github.com/3scale-labs/authorino/blob/2b6a6f8016a5837650506cd125751f8d95ea4197/examples/envoy/Dockerfile#L10

The implementation should be upgraded to https://github.com/envoyproxy/envoy/blob/master/api/envoy/service/auth/v3/external_auth.proto

guicassolato:
Closes #4

----

### Notes to the reviewer:

- Envoy ext auth config now requires `transport_api_version` so it doesn't default to v2 ([ref](https://www.envoyproxy.io/docs/envoy/latest/faq/api/envoy_v3#faq-api-v3-config))
- Envoy container image used in the example bumped to [v1.17-latest](https://hub.docker.com/r/envoyproxy/envoy/tags?page=1&ordering=last_updated&name=v1.17-latest) – Specification of the API version in the Envoy config only supported starting in v1.15 ([ref](https://github.com/envoyproxy/envoy/commit/96920250a05e62095a68fbaf4f77612179e5af32))
- Imported packages related to the Envoy auth service now all labeled with the "envoy_" prefix for clarity
- Timeout setting of the gRPC service increased to 1s (default is 200ms) to avoid some 403's experienced in the demo
- The PR contains commits of #45


pkg/config/metadata/user_info.go
```diff
@@ -18,6 +19,11 @@ type UserInfo struct {
 func (self *UserInfo) Call(ctx common.AuthContext) (interface{}, error) {
 	// find oidc config and the userinfo endpoint
 	idConfig, _ := ctx.FindIdentityByName(self.OIDC)
+
+	if idConfig == nil {
+		return nil, fmt.Errorf("Null OIDC object for config %v. Skipping related UserInfo metadata.", self.OIDC)
```
jmprusi, line 24: we should propagate a logger here

jmprusi (approved):

pkg/config/metadata/user_info.go
guicassolato: Agreed. Out of scope of this PR though.

I was planning to work on the User Info evaluator anyway. Not happy with how it works now.

Thanks.

didierofrivia (approved):

PR #49, opened 2021-02-16
Concurrency control enhanced

guicassolato:
- [x] Writing safely to the auth context map of configs and evaluated objects
- [x] Concurrency context propagated to the `Call()` function of the evaluators
- [x] Append evaluators context to auth parent context (i.e. the grpc call)
    
    Screenshot 2021-03-01 at 16 28 08

- [x] Add more returns inside the evaluators when the context is finished (especially identity/oidc and metadata/uma; leaving metadata/user_info for #53)

----

Based on https://github.com/3scale-labs/authorino/pull/45 – MERGED


pkg/common/functions.go
```diff
@@ -0,0 +1,62 @@
+package common
```
didierofrivia, line 1: Generally the file name would be the same name of the package, in this case, `common.go`. When we start having more related to context, json, etc... we could split them into `context.go`, `json-utils.go`. etc..

pkg/service/auth_context.go
```diff
@@ -18,6 +18,29 @@ var (
...
+}
+
+func newEvaluationResponseFailure(evaluator common.AuthConfigEvaluator, err error) EvaluationResponse {
+	return EvaluationResponse{
+		Evaluator: evaluator,
+		Success:   false,
+		Error:     err,
+	}
+}
+
```
didierofrivia, line 43: I think here we could simplify  a bit just removing the `Success` from the `EvaluationResponse` struct. Then we could have just one func `newEvaluationResponse` like:
```go
func newEvaluationResponse(evaluator common.AuthConfigEvaluator, obj interface{}, err error) EvaluationResponse {
	return EvaluationResponse{
		Evaluator: evaluator,
                Object: obj,
		Error:     err,
	}
}

```

later it could be called depending on success or failure like:

```go
newEvaluationResponse(evaluator, nil, err) # error found
newEvaluationResponse(evaluator, obj, nil) # as cool as kim deal
```

later on we would just compare the `EvaluationResponse.Error` to `nil` to flow as intended.

pkg/common/functions.go
```diff
@@ -0,0 +1,62 @@
...
+//   if err := common.CheckContext(ctx); err != nil {
+//   	 return err
+//   } else {
+//     doSomething()
+//   }
+// }
+func CheckContext(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("Context aborted")
```
didierofrivia, line 25: Probably we will want an scenario where the context was cancelled and it didn't return an error, the cancellation was intended, part of the execution.

pkg/config/authorization.go
```diff
@@ -17,12 +18,12 @@ type AuthorizationConfig struct {
 	JWT *authorization.JWTClaims `yaml:"jwt"`
 }
 
-func (config *AuthorizationConfig) Call(ctx common.AuthContext) (interface{}, error) {
+func (config *AuthorizationConfig) Call(authContext common.AuthContext, ctx context.Context) (interface{}, error) {
```
didierofrivia, line 21: We might want to look for a different name for the `AuthContext` interface and struct... it's not a golang context per se, and it gets weird when one understands that in fact we are passing just one context instead of two.

pkg/common/functions.go
guicassolato: We've already started having more. I'll try to be more domain domain-oriented than type-oriented then. Thanks.

pkg/common/functions.go
guicassolato: It wouldn't make any sense to use this helper function in such case, I think. However, you can still just ignore the response, no?

```go
  if err := common.CheckContext(ctx); err != nil {
    // Yeah, it looks like context is over
    // Do something cool then
  }
```

pkg/config/authorization.go
guicassolato: I thought about it as well. I see the confusion and also would like to avoid it. Thing is Golang has appropriated the word "context" which has a way richer meaning in English than what one can do with it in Go. I couldn't find a better name for ours. Maybe "AuthRequest"? (Though it is more than just "request".) Suggestion?

didierofrivia: Let's put a pin on this, for a future refactor.

pkg/common/functions.go
didierofrivia: I think it's not worth the abstraction.. specially when we most probably we'll take different actions when it's cancelled. Maybe @eguzki can enlighten us ?

eguzki: If there will be different actions when it is cancelled,  I would do something like this

```
func IsContextDone(ctx context.Context) bool {
	select {
	case <-ctx.Done():
		return true
	default:
		return false
	}
}
```
But still, this approach might not be valid for all scenarios. In fact, the most common scenario would be the one documented in their [source code](https://golang.org/src/context/context.go), wait  for many things to happen and if the context is cancelled first, then react:
```
func Stream(ctx context.Context, out chan<- Value) error {
	for {
		v, err := DoSomething(ctx)
		if err != nil {
			return err
		}
		select {
		case <-ctx.Done():
			return ctx.Err()
		case out <- v:
		}
	}
}
```

guicassolato: Guys, I agree replacing

```ruby
		return fmt.Errorf("Context aborted")
```

with
```ruby
		return ctx.Err()
```

But I think having the helper method still makes sense. A few things to have in mind:
1. I wouldn't want to have to write a `select..case ` statement every time for something that a simple `if` would do.
2. Most of the places where it's used (if not all of them!) returns an error to the caller. Making this check to also return an error is an easy way to avoid dozens of different error messages, spread all over the code, for something that is essentially the same. So, I wouldn't make it return a `bool`.
3. The example of usage [pointed out](https://github.com/3scale-labs/authorino/pull/49#discussion_r584582801) by @eguzki is one where, _first_, the code does something (potentially something heavy), then if the context is not done yet, uses the result of that something; otherwise, returns the context error. Maybe that's because it is inside of a `for` loop and trying to depict that the task would usually be performed at least once before the context could be cancelled. Here, however, I mostly implemented the check _before_ running the (potentially heavy) task, not _after_ it. I guess there's no ideal solution for this type of problem actually. If it's after the task, it means all concurrent jobs will do something and some of them later realize it wasn't needed to spend time on it after all; if it's before, then there is the risk of, due to concurrency, all jobs to misjudge that performing the task is needed, when actually it isn't or it won't be soon but you cannot tell now. Hopefully, this difference won't matter much because we are adding here not one single check per evaluator, per possibly several of them, interspersed in middle of multiple steps performed by each evaluator.

didierofrivia: I'm not totally convinced but I'll keep an open mind and see the future usage of the helper. Let's roll with it

didierofrivia (approved):
Some minor details pointed out, that we will address in the future. It's a good enhancement, much clearer with the response structs and flow overall

pkg/common/functions.go
eguzki: > Here, however, I mostly implemented the check before running the (potentially heavy) task, not after it.

That depends on the "cancelling" approach you want. Let's say you want to do something that takes seconds or minutes. In heavy load, do you want to have the long running task finished before returning back to handle the next job? That is one way. Another way is to cancel the long running job as soon as the cancel signal arrives. For that approach, you need the example of usage pointed out. 

Anyway, as said earlier, I think it is to soon to have this kind of concerns. They can be tackled later on.

PR #50, opened 2021-02-17
Removing vendor dependencies from codebase

didierofrivia:
Following the thread about vendoring where the team decided to remove it because Golang state of modules are better than in the past, the annoyance that caused in every PR and that we are doing "vendoring" in productization pipelines...

Screenshot 2021-02-17 at 09 10 13


guicassolato (commented):
Some work needed on the Makefile as well perhaps? Possibly in the README too.

Makefile
```diff
@@ -62,6 +62,10 @@ deploy: manifests kustomize
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
+# Download vendor dependencies
+vendor:
```
guicassolato, line 66: Maybe add this target to `local-setup`

README.md
```diff
@@ -465,5 +465,7 @@ kind delete clusters authorino-integration
     - [Golang](https://golang.org)
     - [Operator SDK](https://sdk.operatorframework.io/)
 - Fork the repo
+- Download dependencies
+  `go mod download`
```
guicassolato, line 469:
```suggestion
  `make vendor`
```

guicassolato:
To test this locally, you may want to

```
rm -rf vendor
go clean -modcache
make vendor
```

Makefile
didierofrivia: It's added to `all`, that should work, no?

guicassolato (approved):

PR #52, opened 2021-02-17
upgrade dex to v2.26

guicassolato:
This is so we can use dex's userinfo endpoint ([introduced](https://github.com/dexidp/dex/commit/8b4dbb9fe70de468ebb6301d20ae64344dfa2846) in v2.17).

~The PR also changes dex storage type to `kubernetes`, which requires adding a new CRD and corresponding RW permissions for the dex service account.~

It also changes dex's db volume to `emptyDir` type so it doesn't require any special config for Kind to create the dir in the host. Actually, no dir at the host is needed at all. The trade-off was that now dex stored data is truly ephemeral. If the pod goes down, all issued access tokens should be lost. This is obviously not good for production but enough for the demo.

----

Updated README:
https://github.com/3scale-labs/authorino/tree/upgrade-dex/examples/dex


examples/dex/README.md
```diff
@@ -40,7 +44,8 @@ kubectl -n authorino apply -f ./examples/echo-api-protection.yaml
 
 Add dex to the cluster:
 ```shell
-kubectl -n authorino create -f examples/dex.yaml
+kubectl create -f examples/dex/crd.yaml
```
eguzki, line 47: Shouldn't installing the CRD responsibility of the owner of the CRD (usually an operator)? Shouldn't be required the operator to be installed up  and running?

guicassolato: There's no operator in this case. This CRD is the data structure dex uses to store its state in etcd, alternatively to an actual database system.

eguzki: It is very weird, but ok

guicassolato: Yeah. I went with dex v2.26 because the version we were using before (v2.10) didn't have the OAuth2 UserInfo endpoint. dex introduced it in v2.17.

When the example was based still on v2.10, dex was configured to store data in a sqlite3 db. Unfortunately that didn't work when I upgraded to v2.26. I think it's related to our Kind cluster and permissions. Therefore I switched dex's storage to "kubernetes", which means it will store its data in CRs.

Right now, I have no plans to deploy any dex operator only for this "Trying Authorino with dex" example. Unless it was something super easy to do, but I honestly couldn't even find anything about a "dex operator" in their documentation, nor in OperatorHub. So we're hosting the CRDs, whose version should be paired with image version of dex that we are using.

For reference:
- https://dexidp.io/docs/kubernetes/#deploy-the-dex-server
- https://dexidp.io/docs/storage/#kubernetes-custom-resource-definitions-crds

eguzki: For devtesting purposes, it should not be a big deal.  :+1:

examples/dex/crd.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: 2017-09-13T19:56:28Z
+  name: authcodes.dex.coreos.com
```
jmprusi, line 5: You can leave only the `name`, all the other info is not necessary and could cause some issues.

examples/dex/README.md
```diff
@@ -40,7 +44,8 @@ kubectl -n authorino apply -f ./examples/echo-api-protection.yaml
...
+sed -e 's/\${NAMESPACE}/authorino/g' examples/dex.yaml | kubectl -n authorino create -f -
```
jmprusi, line 48: as pointed before, you can get rid of the `sed` and just create it.

jmprusi (changes requested):

examples/dex/crd.yaml
guicassolato: ⌘+C, ⌘+V

examples/dex/crd.yaml
```diff
@@ -0,0 +1,34 @@
...
+spec:
+  group: dex.coreos.com
+  names:
+    kind: AuthCode
+    listKind: AuthCodeList
+    plural: authcodes
+    singular: authcode
+  scope: Namespaced
+  version: v1
+status:
```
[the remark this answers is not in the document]
guicassolato: Same https://github.com/3scale-labs/authorino/pull/52#discussion_r579187535

examples/dex.yaml
```diff
@@ -85,3 +80,36 @@ spec:
...
   selector:
     app: dex
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app: dex
+  name: dex
+  namespace: ${NAMESPACE}
```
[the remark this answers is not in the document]
guicassolato: We need to make sure it matches https://github.com/3scale-labs/authorino/blob/7bcef0622092cc10635774d9cc43e952f5a85ef1/examples/dex.yaml#L115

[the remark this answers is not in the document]
guicassolato: Tested. Unfortunately not possible. Creation of the ClusterRoleBinding will fail with:

```
The ClusterRoleBinding "dex" is invalid: subjects[0].namespace: Required value
```

jmprusi (approved):

PR #53, opened 2021-02-19
Proper OAuth/OIDC UserInfo metadata

### LINKED ISSUE #51: Missing proper OAuth2 UserInfo metadata
Current implementation of the `userinfo` metadata is more a Keycloak-specific token introspection than OAuth2 UserInfo.

**Expected behaviour**
- `userinfo` metadata works like normal OAuth2/OIDC [UserInfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo)
    ```
    GET /userinfo HTTP/1.1
    Host: server.example.com
    Authorization: Bearer SlAV32hkKG
    ```
- Token introspection implemented in another metadata type.

guicassolato:
Makes userinfo metadata compliant to openid connect userinfo endpoint [spec](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo), instead of current Keycloak-specific token_introspection implementation.

- [x] Use OIDC userinfo_endpoint
- [x] Check context before sending the request to the fetch userinfo endpoint
- [x] Resuse same credentials config from the matching OIDC identity config
- [x] Add a 'name' attribute to the metadata type, so the auth objects are identified by this instead of the protocol name (i.e. oidc, uma, etc)
- [x] Move field `OIDC.Name` to the `IdentityConfig` type (following the CRD)
- [x] OIDC ref of the UserInfo config as a pointer
- [x] Refactor the service controller's `translateService` function
- [x] Update the examples (default README and dex)
- [x] Tests

**Future iteration:**
- Re-introduce Keycloak-specific token introspection metadata (?)

----

Closes #51


pkg/config/metadata/user_info.go
```diff
@@ -17,26 +19,58 @@ type UserInfo struct {
 	ClientSecret string `yaml:"client_secret"`
 }
 
-func (self *UserInfo) Call(authContext common.AuthContext, ctx context.Context) (interface{}, error) {
-	// find oidc config and the userinfo endpoint
-	idConfig, _ := authContext.FindIdentityByName(self.OIDC)
+func (userinfo *UserInfo) Call(authContext common.AuthContext, ctx context.Context) (interface{}, error) {
+	// find associated oidc identity config
+	var oidcIdentityConfig *identity.OIDC
+	identityConfig, err := authContext.FindIdentityConfigByName(userinfo.OIDC)
```
didierofrivia, line 25: Another way would be assigning the pointer to the Identity when building the `UserInfo` struct

pkg/config/metadata/user_info_test.go
```diff
@@ -0,0 +1,167 @@
...
+      client_id: my_client
+      client_secret: do_not_tell_anyone
+`, authServerHost)
+
+	wellKnownOIDCConfig string = fmt.Sprintf(`{
+		"issuer": "http://%s",
+		"userinfo_endpoint": "http://%s/userinfo"
+	}`, authServerHost, authServerHost)
+
+	authContext *__AuthContext
```
didierofrivia, line 45: Maybe in the near future we could introduce factories for mock objects... specially with the `AuthContext` one (which we might want to choose a better name then)

didierofrivia (commented):
There are a couple of things that I'd rather build different, since we have the data to create the  `UserInfo` instance, we might avoid calling external methods. Still, it's nothing critical and I reckon this will keep changing with new features

pkg/config/metadata/user_info.go
guicassolato: And I think that other way will actually be a better one, but I'll leave it for another PR.

In this next iteration, I want to point the UserInfo struct directly to the OIDC one or allow a user_info endpoint directly in the UserInfo struct alternatively. The tricky part is that we'll soon introduce OAuth identity type which technically would admit a pointing UserInfo struct too.

Either way, if/when the UserInfo struct points to an OIDC identity struct, the identity struct should be caching the discovered OIDC configuration, making possible to save that step done again by `UserInfo.Call`.

guicassolato: And I think that other way will actually be a better one, but I'll leave it for another PR.

In this next iteration, I want to point the UserInfo struct directly to the OIDC one or, alternatively, allow a user_info endpoint to be set directly in the UserInfo struct. The tricky part of this is that soon we'll be introducing OAuth identity type which should accept UserInfo structs pointing to as well, so this pointer would have to be to either an OIDC struct or to an OAuth struct. While as a string, we don't have such problem.

Either way, if/when the UserInfo struct points to an identity struct (OIDC or OAuth), this identity struct should hold the endpoints. If the identity config is OIDC, then it will be caching the entire OIDC config actually and this will make possible to save a step in `UserInfo.Call`.

controllers/service_controller.go
```diff
@@ -102,116 +103,96 @@ func (r *ServiceReconciler) translateService(ctx context.Context,
...
 		case configv1beta1.IdentityOidc:
-			authCred := &auth_credentials.AuthCredential{
-				KeySelector: identity.Credentials.KeySelector,
-				In:          identity.Credentials.In,
-			}
-			translatedIdentity = config.IdentityConfig{
-				OIDC: authorinoIdentity.NewOIDCIdentity(identity.Name, identity.Oidc.Endpoint, authCred),
+			translatedIdentity.OIDC = &authorinoIdentity.OIDC{
+				Endpoint:    identity.Oidc.Endpoint,
+				Credentials: authCred, // FIXME: this should not be here as the identity struct is supposed to hold the creds, not the OIDC struct
```
didierofrivia, line 121: Yes, it should be at Identity level... I take the blame, forgot to hoist it hehe.

guicassolato: For now, let's keep copying or moving this struct into the lower level of the identity config, i.e. in the actual evaluator. It will make the code simpler, without any fancier (messier?) design using interfaces and pointers to workaround Golang's no-circular refs constraint.

We can revisit this decision in the future, say when/if other things currently stored in the generic/upper level of the identity config start being requested by the specific evaluators (e.g., the `Name` of the identity config or any other field that we end up adding). If that happens and we are forced to re-think the structures of configs and evaluators, perhaps then one option would be getting rid of the middle layer altogether (i.e. `IdentityConfig`, `MetadataConfig` and `AuthorizationConfig`) and working straight with the evaluator structs (`OIDC`, `APIKey`, `UserInfo`, `UMA`, `OPA`, etc), always copying to the lower level what is needs (thus making it self-contained, good for concurrency).

didierofrivia: On a second thought, this actually makes sense, it's a classic composite pattern (structural). After the call we had it was a bit clearer. We could talk about it later today.

didierofrivia (approved):
LGTM! 🎖️

PR #55, opened 2021-03-02
API Key auth method

didierofrivia:
This PR introduces the authentication through API Key feature.

Closes https://github.com/3scale-labs/authorino/issues/7

## Use case:
The developer of an API wants to provide some API consumer with an API key so the latter have access to the developer's service (authenticated through Authorino).

To make it happen, the developer will create opaque k8s secrets and configure Authorino identity configs with a specific `key_selector` and lookup labels (i.e. `label_selectors`), where `key_selector` tells where Authorino will get the supplied API key from, in the request sent by the consumer, and `label_selectors` tells what subset of secrets created in Authorino's namespace Authorino will effectively consider while looking up for the supplied API key.

The developer creates the secret in the cluster:

```bash
kubectl -n authorino create secret generic lesecret --from-literal=api_key=123456

kubectl -n authorino label secret lesecret target=echo-api
```

Then she adds to the Authorino-protected `Service` the following identity config:

```yaml
...
identity:
    - name: friends # user defined
      credentials:
         in: custom_header # options include: custom_header, authorization_header and query
         key_selector: X-API-KEY # user defined
      api_key:
        label_selectors:
          target: echo-api # user defined
...
```

The consumer can then start `curl`ing the service like this:

**Valid API key:**
```bash
curl -H 'Host: echo-api' -H 'X-API-KEY: 123456'  http://localhost:8000/api/ohai -v  # 200
```

**Invalid API key:**
```bash
curl -H 'Host: echo-api' -H 'X-API-KEY: nonono'  http://localhost:8000/api/ohai -v  # 403
```

## TODO:
* [x] Define APIKey yaml config
* [x] Get secrets from cluster 
* [x] Get APIKey from request headers https://github.com/3scale-labs/authorino/pull/58
* [x] Get APIKey from request query params https://github.com/3scale-labs/authorino/pull/58
* [x] Tests


pkg/config/identity/api_key.go
```diff
@@ -1,13 +1,88 @@
...
+	}
+
+	reqHeaders := authCtx.GetRequest().Attributes.Request.Http.Headers
+	fmt.Println("Le reqHeaders", reqHeaders)
+	reqPath := authCtx.GetRequest().GetAttributes().GetRequest().GetHttp().GetPath()
+	fmt.Println("Le reqPath", reqPath)
+	reqQuery := authCtx.GetRequest().GetAttributes().GetRequest().GetHttp().GetQuery()
+	fmt.Println("Le reqQuery", reqQuery)
+	reqFragment := authCtx.GetRequest().GetAttributes().GetRequest().GetHttp().GetFragment()
+	fmt.Println("Le reqFragment", reqFragment)
```
guicassolato, line 71: I think we'll be using the `Location` field for this (probably in a `switch..case` statement), right?

didierofrivia: Yup. Also, don't trouble yourself with review just now since it's super WIP hehe

guicassolato:
Looking good so far, @didierofrivia.

I think we could also consider moving the responsibility of fetching the matching secrets to the reconciliation cycle, done upfront and caching them, instead of doing it in every request. Or in evaluation-time on every request as you did it, but caching the matches to save this step next time. Anyway, ideas for improvement in a next iteration.

pkg/service/auth_context.go
```diff
@@ -251,7 +251,7 @@ func (authContext *AuthContext) GetMetadata() map[string]interface{} {
 
 func (authContext *AuthContext) FindIdentityByName(name string) (interface{}, error) {
 	for identityConfig := range authContext.Identity {
-		if identityConfig.OIDC.Name == name {
+		if identityConfig.OIDC != nil && identityConfig.OIDC.Name == name {
```
guicassolato, line 254: Hmm. I think this nil-check should be done in the client, not here. E.g. https://github.com/3scale-labs/authorino/blob/6a9c800929e7a48d90e6a2a791d26883582ec90c/pkg/config/metadata/user_info.go#L25-L29

Reasons:
1. it is more consistent with the name of the function which suggests finding a config and nothing else.
2. Because the function returns the config, not the evaluated identity, after all.

guicassolato: BTW, even fetching only OIDC identity configs here is wrong. That's why I'm proposing to change it in #53.

didierofrivia: It makes sense, I'll change it. However, there's something that smells, I think that relying on `OIDC` not being `nil` is a bit flawed since it's possible that we don't want `OIDC` as an identity method in the first place (as in this PR case). Maybe the problem relies on how we are storing the `Identity` config, we should find a better way and not being implicit about those methods.

didierofrivia: Didn't see your second comment until after reloading the page hehe. Yes, that proposing change is more or less what I'm talking about

pkg/config/identity/api_key.go
```diff
@@ -2,14 +2,59 @@ package identity
...
-	SecretKey string `yaml:"secret_key"`
+	Name           string            `yaml:"name"`
+	Location       string            `yaml:"location"`
+	KeySelector    string            `yaml:"key_selector"`
+	LabelSelectors map[string]string `yaml:"label_selectors"`
+	secrets        []string
+}
+
+const (
+	locationHeader     = "header"
```
guicassolato, line 20: [This page](https://www.golangprograms.com/go-language/constants.html) claims uppercase to be the convention for constant names in Golang, while [this answer](https://stackoverflow.com/a/22688926/841699) in StackOverflow points to camel case just like other symbols (variables, function names, etc).

I know the choice affects the visibility of the symbol and also that so far we seem to have adopted camel case. Still wondering if you guys are happy with this convention or not. Particularly, I'd love to use uppercase but at the same time I don't it to feel weird for other more experienced Golangers reading the code later. What do you guys think?

didierofrivia: The official [Golang blog](https://blog.golang.org/constants) follows the [StackOverflow answer](https://stackoverflow.com/a/22688926/841699) you found.

controllers/service_controller.go
```diff
@@ -264,3 +278,17 @@ func (r *ServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&configv1beta1.Service{}).
 		Complete(r)
 }
+
+func (r *ServiceReconciler) getClusterSecrets(ctx context.Context, apiKeyConf *authorinoIdentity.APIKey) ([]string, error) {
```
didierofrivia, line 282: I could place this function within the ApiKey module in https://github.com/3scale-labs/authorino/pull/55/files#diff-25d3e852acdfdbd26bb9a6f3f0e2941b2a54b32d508483b3054790187e2079ac , maybe would be nicely encapsulated and tested within that package. WDYT @guicassolato ?

guicassolato: Not sure if the APIKey identity should be aware of Secrets. For all purposes, the APIKey identity needs a list of authorized identities. If that list comes from Secrets, that's a details that can be abstracted IMO. In fact, even the `SetSecrets` function (as well as internal fields of the struct perhaps) should reflect that abstraction.

pkg/config/identity/api_key.go
```diff
@@ -2,14 +2,61 @@ package identity
...
+	auth_credentials.AuthCredentials
+	apiKeyDetails
+}
+
+var (
+	apiKeyLog = ctrl.Log.WithName("Authorino").WithName("ApiKey")
+)
+
+func NewApiKeyIdentity(name string, labelSelectors map[string]string, authCred auth_credentials.AuthCredentials) (*APIKey, error) {
+	if credentials, err := authCred.GetCredentialsFromCluster(context.TODO(), labelSelectors); err != nil {
```
guicassolato, line 34:
```suggestion
func NewApiKeyIdentity(name string, labelSelectors map[string]string, authCred auth_credentials.AuthCredentials, ctx context.Context) (*APIKey, error) {
	if credentials, err := authCred.GetCredentialsFromCluster(ctx, labelSelectors); err != nil {
```

guicassolato, line 34: Hmm. What about k8s secrets that are created in the cluster _after_ Authorino loads the service config?

didierofrivia: We'll need to define when this will happen, like.. would it be triggered by a stale state? user interaction? a regular worker? the service controller should do it when it detects a change from the cluster?

guicassolato: I don't think it's really Authorino's controller job to reconcile k8s secrets. Yet, we should expect developers to create/delete/update secrets at anytime (as this is literally the equivalent of managing access to the protected service in this authn method). Finally, querying the cluster on every request to the protected service would obviously imply an overhead. Not sure if there is any good solution for this.

pkg/common/auth_credentials/auth_credentials.go
```diff
@@ -38,14 +56,29 @@ func (c *AuthCredential) GetCredentialsFromReq(httpReq *envoyServiceAuthV3.Attri
...
+func (c *AuthCredential) GetCredentialsFromCluster(ctx context.Context, labelSelectors map[string]string) ([]string, error) {
+	var matchingLabels client.MatchingLabels = labelSelectors
+	var secretList = &v1.SecretList{}
+	if err := c.k8sClient.List(ctx, secretList, matchingLabels); err != nil {
+		return nil, err
 	}
+	var parsedSecrets []string
+
+	for _, secret := range secretList.Items {
+		parsedSecrets = append(parsedSecrets, string(secret.Data[c.KeySelector]))
```
guicassolato, line 72: Wee need to highlight this in the instructions somehow. BTW, the key of the k8s secret is case sensitive.

pkg/config/identity/api_key.go
```diff
@@ -2,14 +2,61 @@ package identity
...
-func (self *APIKey) Call(authContext common.AuthContext, ctx context.Context) (interface{}, error) {
-	return "Authenticated with API key", nil // TODO: implement
+func (apiKey *APIKey) Call(authCtx common.AuthContext, _ context.Context) (interface{}, error) {
+	if reqKey, err := apiKey.GetCredentialsFromReq(authCtx.GetHttp()); err != nil {
+		apiKeyLog.Error(err, noApiKeysFoundMsg)
+		return nil, err
+	} else {
+		for _, secret := range apiKey.authorizedCredentials {
+			if secret == reqKey {
+				return authSuccessfulMsg, nil
```
guicassolato, line 55: I think here we could return something from the matched secret instead of a fixed string. I understand that this would involve storing more in the `AuthCredentials` struct, but it might be a worthy trade-off. I can see some usage for the secret's annotations, for example, later on in the authorization policies.

examples/echo-api-key-simple.yaml
```diff
@@ -0,0 +1,27 @@
...
+        in: custom_header # options include: custom_header, authorization_header and query
+        key_selector: X-API-KEY # user defined
+      api_key:
+        label_selectors:
+          target: echo-api # user defined
+  metadata:
+    - userInfo:
+        identitySource: keycloak
+        credentialsRef:
+          name: userinfosecret
```
guicassolato, line 20:
```suggestion
  metadata: []
```

pkg/common/auth_credentials/auth_credentials.go
```diff
@@ -1,34 +1,52 @@
...
 
 	envoyServiceAuthV3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+	v1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type AuthCredentials interface {
 	GetCredentialsFromReq(*envoyServiceAuthV3.AttributeContext_HttpRequest) (string, error)
+	GetCredentialsFromCluster(context.Context, map[string]string) ([]string, error)
```
guicassolato, line 17: I think we're mixing two different things here, one being the data structure that tells how request credentials are expected to be supplied and a tool to extract such value from a given instance of request, and another thing which is a service to fetch an array of strings from somewhere.

pkg/config/identity/api_key.go
didierofrivia: I left that message knowing we still don't have any usage for any format returned. The idea is whenever we have some specific flow needing more than an error check to change it accordingly.

pkg/config/identity/api_key.go
didierofrivia: Me neither TBH, that's why I raised those questions. I'd put a pin on this until we find out the right way to "refresh" those authorized credentials

pkg/common/auth_credentials/auth_credentials.go
didierofrivia: Yeah, you are right, this is kind of key to make it work

pkg/common/auth_credentials/auth_credentials.go
didierofrivia: One could see it as a module in charge of managing credentials, since those strings are not any simple string. Anyway, where would you see this making more sense?

didierofrivia (commented):
There are some minor changes on config, some guidelines/documentation to add and a lot of answers that we might need time and usage to understand how to change those interfaces.

didierofrivia:
I don't understand why GH forced me to do a review on my own and disabled replying the previous comments :suspect:

pkg/common/auth_credentials/auth_credentials.go
didierofrivia: It's case sensitive, but we are forced to make it lowercase when working with headers. Envoy stores them lowercase.

pkg/common/auth_credentials/auth_credentials.go
guicassolato: For now, getting secrets from the cluster is an APIKey auth identity config-specific thing. It could be part of `APIKey`. Otherwise inside yet another separate module that is only about getting resources from the cluster.

I think in the end practically everything that happens in "identity" could be seen as about managing/handling credentials. Soon, those will be: openid, api_key, oauth, openshift, hmac, etc. IMO, better not opening this door of accommodating in this AuthCredentials module anything that has to do with credentials because that can end up being a lot. In fact, I think AuthCredentials should actually be about "CredentialsConfig", i.e. the struct to store that part of the config that holds where the credentials are carried in the request.

pkg/common/auth_credentials/auth_credentials.go
guicassolato, line 72:
```suggestion
		parsedSecrets = append(parsedSecrets, string(secret.Data["api_key"]))
```

pkg/common/auth.go
```diff
@@ -9,6 +9,7 @@ import (
 type AuthContext interface {
 	GetParentContext() *context.Context
 	GetRequest() *envoy_auth.CheckRequest
+	GetHttp() *envoy_auth.AttributeContext_HttpRequest
```
guicassolato, line 12: Thank you for this 🙂

pkg/config/identity/api_key.go
```diff
@@ -2,14 +2,94 @@ package identity
...
+const (
+	apiKeySelector              = "api_key"
+	invalidApiKeyMsg            = "the API Key provided is invalid"
+	noApiKeysFoundMsg           = "no API Keys were found on the request"
+	authSuccessfulMsg           = "Successfully authenticated with the provided API key"
+	credentialsFetchingErrorMsg = "Something went wrong fetching the authorized credentials"
 )
 
+// APIKeyIdentityEvaluator interface represents the API Key Identity evaluator
+type APIKeyIdentityEvaluator interface {
```
guicassolato, line 24: We have another one already defined at https://github.com/3scale-labs/authorino/blob/e0290e73d6963ce8b5b03e4a7e7edd0618913608/pkg/common/auth.go#L20, which makes it easier to be imported without circular refs. However, let's leave this one here for now and we can move it with the other interfaces of the kind in another PR.

guicassolato: Hmm. I see we're not using the interface for anything just yet. So it's just for the good practice of communicating the contract.

guicassolato (approved):
Excellent, @didierofrivia! Thank you

PR #58, opened 2021-03-05
Auth Credentials

didierofrivia:
This PR will unified the way credentials/secrets are managed through all auth methods.

One needs to declare within the identity configuration (besides other specific attributes), the location and key selector as follows:
```yaml
...
identity:
    - name: default
      oidc:
        endpoint: http://identity.provider/auth
     credentials:
        in: authorization_header
        key_selector: Bearer

    - name: vip
      api_key:
        selectors:
          app: authorino
          target: echo-api
     credentials:
        in: custom_header
        key_selector: X-API-KEY
```

Notes:

* The `in` key will target the secret location, valid values are: `custom_header`, `authorization_header` and `query` at the moment.
* The `key_selector` will target the designated "name" for a custom header or query param, and the "scheme" (i.e. Bearer, APIKEY, etc...) in the http authorization header.


pkg/common/auth_credentials/auth_credentials.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+	envoyServiceAuthV3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+	"regexp"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"strings"
+)
+
+type AuthCredentials interface {
+	GetCredentialsFromReq(*envoyServiceAuthV3.AttributeContext_HttpRequest) (string, error)
```
guicassolato, line 12: This `envoyServiceAuthV3.AttributeContext_HttpRequest` is spreading within our code like covid. On another PR, I was thinking about replacing with sth like: https://github.com/projectcontour/contour-authserver/blob/main/pkg/auth/convert.go

guicassolato (approved):

PR #63, opened 2021-03-16
Terminology doc

andrewdavidmackenzie:
Add an initial terminology.md document in ./docs/

### DIFF
--- a/docs/terminology.md
+++ b/docs/terminology.md
@@ -2,7 +2,7 @@
 Here we define some terms that are used in the project, with the goal of avoiding confusion and facilitating more
 accurate conversations related to `Authorino`.
 
-If you see terms used that are not here (or are used in place of terms here) please considre contributing a definition
+If you see terms used that are not here (or are used in place of terms here) please consider contributing a definition
 to this doc with a PR, or modifying the use elsewhere to align with these terms.
 
 ## Terms
--- a/docs/terminology.md
+++ b/docs/terminology.md
@@ -1,0 +1,65 @@
+# Terminology
+Here we define some terms that are used in the project, with the goal of avoiding confusion and facilitating more
+accurate conversations related to `Authorino`.
+
+If you see terms used that are not here (or are used in place of terms here) please considre contributing a definition
+to this doc with a PR, or modifying the use elsewhere to align with these terms.
+
+## Terms
+* **Access token**: a type of temporary security token, tied to an authenticated identity, that is issued by an auth 
+  server under request of registered auth clients of this auth server, and that delegates power of operating on 
+  behalf of that identity to a party that presents this token on requests to a resource server, including and 
+  especially to the auth client itself; it is typically formatted as a JWT
+* **Auth**: usually employed as a short for authentication and authorization (AuthN/AuthZ)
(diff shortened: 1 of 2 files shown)


docs/terminology.md
```diff
@@ -0,0 +1,72 @@
+# Terminology
+Here we define some terms that are used in the project, with the goal of avoiding confusion and facilitating more
+accurate conversations related to `Authorino`.
+
+If you see terms used that are not here (or are used in place of terms here) please considre contributing a definition
```
guicassolato, line 5:
```suggestion
If you see terms used that are not here (or are used in place of terms here) please consider contributing a definition
```

guicassolato (approved):
Thank you!

docs/terminology.md
```diff
@@ -0,0 +1,44 @@
+# Terminology
+Here we define some terms that are used in the project, with the goal of avoiding confusion and facilitating more
+accurate conversations related to `Authorino`.
+
+If you see terms used that are not here (or are used in place of terms here) please consider contributing a definition
+to this doc with a PR, or modifying the use elsewhere to align with these terms.
+
+## Terms
+* **Access token**: a type of temporary security token, tied to an authenticated identity, that is issued by an auth server under request of registered auth clients of this auth server, and that delegates power of operating on behalf of that identity to a party that presents this token on requests to a resource server, including and especially to the auth client itself; it is typically formatted as a JWT
```
didierofrivia, line 9: Maybe instead of...

```suggestion
* **Access token**: a type of temporary security credential, tied to an authenticated identity, that is issued by an auth server under request of registered auth clients of this auth server, and that delegates power of operating on behalf of that identity to a party that presents this token on requests to a resource server, including and especially to the auth client itself; it is typically formatted as a JWT
```

didierofrivia: Just to avoid defining it with the same word (token) and also because below we are saying that the AuthN process verifies credentials belonging to an identity.

docs/terminology.md
```diff
@@ -0,0 +1,44 @@
...
+* **Auth**: usually employed as a short for authentication and authorization (AuthN/AuthZ)
+* **Auth client**: an application client that uses an auth server, either to authenticate or to authorize identities who want to consume resources from a resources server; it can represent a user application like a SPA or a mobile app, or a resource server
+* **Auth server**: a server where auth clients, users, roles, scopes, resources, policies and permissions are stored and managed
+* **Authentication (AuthN)**: the process of verifying that a given credential belongs to a claimed-to-be identity; usually resulting in the issuing of an access token
+* **Authorization (AuthZ)**: the process of grating (or denied) access over a resource to a party based on the set of policies and permissions enforced and
```
didierofrivia, line 14:
```suggestion
* **Authorization (AuthZ)**: the process of granting (or denying) access over a resource to a party based on the set of policies and permissions enforced
```

didierofrivia (changes requested):
Some minor changes needed, besides that would be good to choose to start with capital or lowercase and end with (or not) a point the sentences, just in order to keep consistency. Thanks for the contribution!

didierofrivia (approved):

PR #64, opened 2021-03-17
Update CRD spec with API key auth option

guicassolato:
Missing in #55

PR #66, opened 2021-03-18
[Fix] Removes `Location` and `KeySelector` from APIKey Identity

didierofrivia:
Both are part of `Credentials` type.


guicassolato (approved):

guicassolato (changes requested):
Please run `make manifests`... and we can close #64

guicassolato (approved):

Issue #56, opened 2021-03-02
UUID of the OPA policy should be auto-generated


guicassolato:
Users shouldn't be required to define a UUID for their OPA authorization policies.

https://github.com/3scale-labs/authorino/blob/ce5bbf4e0db0a37f470723c6f6cfff439dcd1fd6/examples/echo-api-protection.yaml#L23

Authorino should auto-generate those, maybe by hashing some unique data of the `Service` custom resource, e.g. name or selfLink of the resource, plus the index of the policy in the array of `authorization` configs.

guicassolato:
Done in #67

PR #67, opened 2021-03-18
Simple JSON pattern matching authorization

### LINKED ISSUE #11: Simple pattern matching authZ
There's a [placeholder](https://github.com/3scale-labs/authorino/blob/2b6a6f8016a5837650506cd125751f8d95ea4197/pkg/config/authorization/jwt.go) in the code for "JWT claim pattern matching", but perhaps that can be merged into one single "authz rules pattern matching" mechanism.

guicassolato:
Closes #11 and #56

----

**Usage:**

```yaml
identity: [...]
metadata: [...]
authorization:
  - name: admins-only
    json:
      # pattern: object → composed of:
      #   selector: string → can be any path accepted by https://pkg.go.dev/github.com/tidwall/gjson
      #   operator: string → enum: eq, neq, incl, excl, matches (regex)
      #   value: string → expected value to be compared with the obtained one in each request

      # conditions (optional): array of patterns → the policy will only apply if all the conditions match
      conditions:
        - selector: "context.request.http.method"
          operator: "neq"
          value: "OPTIONS" # (rules do not apply when the HTTP method is OPTIONS)

      # rules: array of patterns → the request will be authorized if all rules match
      rules:
        - selector: "auth.identity.role"
          operator: eq
          value: admin
        - selector: "context.source.address.Address.SocketAddress.address"
          operator: matches
          value: "99.99.99.(\\d)+" # (an authorized range of IPs)
```

Other changes that come with this PR:

- [x] Remove any trace of the old `JWTClaimSet` authn mode (Authorino's internal structs and CRD)
- [x] Rename the `OPAPolicy` authn mode as `opa` in the CRD
- [x] Remove the `Enabled` field from the API config and authorization evaluators (they were not being used for anything)
- [x] Add a field `Name` to `AuthorizationConfig` (mainly for the user to identify each authz policy in the config, but also used in an internal hash generated to uniquely identify each OPA policy)
- [x] Move the `OPAInput` struct from `OPA` to `AuthContext` and rename it for general usage →  it is now `AuthContext.ToData()`
- [x] Rename some JSON attributes of the old `OPAInput` structure (now thatthe struct is more general, used by at least 2 authn modes)
    - “attributes” → “context”
    - “context” → “auth”
- [x] Remove the `uuid` field from the OPA authn config in the CRD →  redefined as “OPA.policyUID” and now auto-generated (hash of name of the policy + content of `inlineRego` + nonce, where nonce is the index of the config in the authorization config array)
- [x] Define a function `NewOPAAuthorization` →  builds the struct, generates the policy UID and precompiles the Rego policy
- [x] Remove all logs from the OPA evaluator


pkg/config/authorization/opa.go
```diff
@@ -2,122 +2,86 @@ package authorization
...
+	if regoPolicy, err := rego.New(regoQuery, regoModule).PrepareForEval(opa.opaContext); err != nil {
+		return err
+	} else {
+		opa.policy = &regoPolicy
+		return nil
+	}
+}
+
+func generatePolicyUID(policyName string, policyContent string, nonce int) string {
+	data := []byte(fmt.Sprint(nonce) + policyUIDHashSeparator + policyName + policyUIDHashSeparator + policyContent)
```
guicassolato, line 85: We should include as well the name of the service.

guicassolato: Done

pkg/common/auth.go
```diff
@@ -13,6 +13,7 @@ type AuthContext interface {
 	GetAPI() interface{}
 	GetResolvedIdentity() (interface{}, interface{})
 	GetResolvedMetadata() map[interface{}]interface{}
+	ToData() interface{}
```
didierofrivia, line 16: I'm a bit afraid of what the AuthContext is becoming. I'd like to step back and review the design (and specially the name). Probably after we cover the Tier 1 and desired examples.

pkg/service/auth_context.go
```diff
@@ -249,3 +249,25 @@ func (authContext *AuthContext) GetResolvedMetadata() map[interface{}]interface{
 	}
 	return m
 }
+
+type authContextData struct {
+	Request *envoy_auth.AttributeContext `json:"context"`
+	Context map[string]interface{}       `json:"auth"`
+}
+
+func (authContext *AuthContext) ToData() interface{} {
```
didierofrivia, line 258: I'm not sure what this `ToData` func transforms to. Maybe we could define with a type the desired shape of Data?

pkg/service/auth_context.go
```diff
@@ -249,3 +249,25 @@ func (authContext *AuthContext) GetResolvedMetadata() map[interface{}]interface{
 	}
 	return m
 }
+
+type authContextData struct {
+	Request *envoy_auth.AttributeContext `json:"context"`
+	Context map[string]interface{}       `json:"auth"`
+}
```
didierofrivia, line 256: I'd expect some different structure holding "data", it's a bit puzzling constructing this with a `Request` and a `Context`, specially this last one not being of the type of `context`

pkg/service/auth_context_test.go
```diff
@@ -256,3 +277,19 @@ func TestEvaluateAnyAuthConfigsWithoutError(t *testing.T) {
...
+	identityConfigs = append(identityConfigs, &successConfig{})
+	request := envoy_auth.CheckRequest{}
+	_ = json.Unmarshal([]byte(rawRequest), &request)
+	authContext := newAuthContext(identityConfigs, &request)
+	data := authContext.ToData()
+	if dataJSON, err := json.Marshal(&data); err != nil {
+		t.Error(err)
+	} else {
+		requestJSON, _ := json.Marshal(request.GetAttributes())
+		expectedJSON := fmt.Sprintf(`{"context":%s,"auth":{"identity":null,"metadata":{}}}`, requestJSON)
```
didierofrivia, line 292: So it seems it's AuthN data, or an extended Identity Data (Identity + Metadata)

api/v1beta1/service_types.go
```diff
@@ -93,11 +93,11 @@ type Metadata_UMA struct {
 
 type Authorization struct {
 	Name        string                     `json:"name"`
-	OPAPolicy   *Authorization_OPAPolicy   `json:"OPAPolicy,omitempty"`
+	OPA         *Authorization_OPA         `json:"opa,omitempty"`
```
didierofrivia: 👍

pkg/service/auth_context.go
guicassolato: I couldn't find any better name for it. That is what it does, i.e. it extracts from the auth context only that data that matters.

Maybe we could name it "ToAuthorizationData", but I see how even this can be confusing.

pkg/common/auth.go
guicassolato: I renamed it to `GetDataForAuthorization()`. Hopefully it is clearer now.

pkg/service/auth_context.go
guicassolato: Renamed to `GetDataForAuthorization()`

pkg/service/auth_context.go
guicassolato: I hid this structure as implementation detail of `AuthContext.GetDataForAuthorization()` and it is now:

```go
Context  *envoy_auth.AttributeContext `json:"context"`
AuthData map[string]interface{}       `json:"auth"`
```

I hope it's clearer.

pkg/service/auth_context_test.go
guicassolato: It is the data to perform the authorization. In other words, the strict input that each authorization policy can count on to evaluate its rules.

didierofrivia (approved):
🎖️

PR #68, opened 2021-03-19
K8s secret reconciler

### LINKED ISSUE #65: API key secrets reconciliation
**Describe the bug**
Authorino currently doesn't acknowledge changes made to k8s secrets that represent API keys after the service is deployed, meaning any new secret, secret update or secret deletion will not be reflected in Authorino's internal cache of API keys until the `config.authorino.3scale.net/Service` 

**Help us Reproduce it**

1. Create a k8s secret that can work as an API key (i.e. opaque, containing an "api_key" key and label selectors)
2. Deploy a `config.authorino.3scale.net/Service` with API key auth enabled
3. Check that requests to the service using API key created in (1) works
4. Delete the k8s secret created in (1) – i.e. mimicking revocation
5. Check again requests to the service using API key created in (1) and see that it still works (i.e. revocation not acknowledged by Authorino)

**Expected behaviour**

Changes to matched k8s secrets reflected in Authorino – i.e.
- adding a new secret with an "api_key" key and matching label selectors creates a new credential to access the service
- deleting a secret revokes access to the service
- - updating an existing secret revokes previous corresponding API key and creates a new one

All environments

**Solution proposal**
A second reconciler, only for secrets, could be introduced. This second reconciler should trigger an update on the main reconciler whenever it acknowledges changes to the watched secrets.

To solve the problem of matching the right secrets to watch independently of reconciliating service configs first, a new Authorino-specific label could be defined and required to be added to every secret.

guicassolato:
Closes #65

----

Introduces the requirement for k8s secrets meant for being used as API keys to always include a label `authorino.3scale.net/managed-by` (with whatever non-null value). Consequently, all services configured for API key authentication must as well include in its `label_selectors` of the `api_key` identity config this new label `authorino.3scale.net/managed-by` (with value matching the one used in the secrets).

A limitation of the implementation here presented is that deleted secrets trigger the reconciler but cannot be checked for matching labels and corresponding affected services. This makes a bit harder for the revoking of access to services by deleting the secrets. Two possible workarounds to this limitation are:
- i) first update the secret before deleting it (i.e. remove the `api_key` entry or change its value)
- ii) enhance with authorization policies that check attributes of the resolved secret (e.g. use the secret's `creationTimestamp` for short-lived API keys working as access tokens)

### DIFF
--- a/controllers/secret_controller_test.go
+++ b/controllers/secret_controller_test.go
@@ -1,0 +1,150 @@
+package controllers
+
+import (
+	"testing"
+
+	"github.com/3scale-labs/authorino/api/v1beta1"
+	"gotest.tools/assert"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type fakeReconciler struct {
+	Reconciled bool
+}
+
+func (r *fakeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	r.Reconciled = true
+	return ctrl.Result{}, nil
+}
+
+type secretReconcilerTest struct {
+	secret            v1.Secret
+	service           v1beta1.Service
+	serviceReconciler *fakeReconciler
+	secretReconciler  *SecretReconciler
+}
+
+func newSecretReconcilerTest(secretLabels map[string]string) secretReconcilerTest {
+	secret := v1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Secret",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "bill",
+			Namespace: "authorino",
+			Labels:    secretLabels,
+		},
+		Data: map[string][]byte{
+			"api_key": []byte("123456"),
+		},
+	}
+
+	service := v1beta1.Service{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "config.authorino.3scale.net/v1beta1",
+		},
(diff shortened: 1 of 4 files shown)


controllers/secret_controller_test.go
```diff
@@ -0,0 +1,160 @@
...
+func (r *FakeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	r.Reconciled = true
+	return ctrl.Result{}, nil
+}
+
+func (r *FakeReconciler) Reset() {
+	r.Reconciled = false
+}
+
+func TestReconciler(t *testing.T) {
```
didierofrivia, line 31: Maybe we could do the creation of Service and Secrets outside the `TestReconciler` func, and separate the assertions addressing the type of secrets and if the service was reconciled or not? That way it would be easier to debug and to keep track of coverage

examples/echo-api-key-simple.yaml
```diff
@@ -12,12 +12,18 @@ spec:
...
+        uuid: e314f9a7-3381-4fe9-bc74-4bc02c7a666f
+        inlineRego: |
+          allow {
+            creationTimestampStr := object.get(object.get(identity, "metadata", {}), "creationTimestamp", {})
+            creationTimestamp := time.parse_rfc3339_ns(creationTimestampStr)
+            durationNs := time.now_ns() - creationTimestamp
+            durationSec := durationNs/1000000000
+            durationMin := durationSec/60
+            durationMin < 5
+          }
```
didierofrivia, line 29: As we discussed previously, maybe it would be better to extract this rego expiring key policy to a different file, and leave this one simpler with the non expiring key?

controllers/secret_controller_test.go
guicassolato: I ended up creating a `<Subject>Test` struct. Our former approach of declaring variables that are reused across multiple tests in the same test file easily run into naming conflicts with other files of the same package.

controllers/secret_controller_test.go
```diff
@@ -0,0 +1,154 @@
...
+func (r *FakeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	r.Reconciled = true
+	return ctrl.Result{}, nil
+}
+
+func (r *FakeReconciler) Reset() {
+	r.Reconciled = false
+}
+
+type SecretReconcilerTest struct {
```
didierofrivia, line 32: maybe no need to export this, since it's quite specific to these batch of tests.

```suggestion
type secretReconcilerTest struct {
```

didierofrivia (approved):

PR #69, opened 2021-03-22
Mocking interfaces

didierofrivia:
In order to avoid repetition and insane boilerplate, a good practice is to adopt a mock factory to create mock structs of our interfaces.

This PR introduces the [Go Mock Framework](https://github.com/golang/mock), it can help test and isolate the business logic it depends on, enabling it to compile, link, and run independently. Also, it use the Go's built-in testing package, so no extra dependencies are needed.

The command to create the mocks is the following:

```bash
mockgen -source=pkg/common/auth.go -destination=pkg/common/mocks/mock_common.go
```

`-source`: targets the file where the interfaces one wants to mock are.
`-destination`: will tell the path and file the mocks will be created. Otherwise it will be printed on the screen

The directory where the mocks are placed, usually it's the same one the package is residing, you create a `mocks`directory there. Gomocks doesn't suggest, but other libs like `mockery` use it as a default.


pkg/config/metadata/user_info_test.go
```diff
@@ -169,21 +80,49 @@ func mockHTTPServer() *httptest.Server {
...
+	idConfEval := mock_common.NewMockIdentityConfigEvaluator(ctrl)
+	idConfEval.EXPECT().GetOIDC().Return(newOIDC)
+	authContextMock.EXPECT().GetHttp().Return(nil)
+	authContextMock.EXPECT().GetResolvedIdentity().Return(idConfEval, nil)
+
+	obj, err := userInfo.Call(authContextMock, ctx)
 	assert.NilError(t, err)
 
 	claims := obj.(map[string]interface{})
+	fmt.Println("claims", claims)
```
guicassolato, line 97:
```suggestion
```

guicassolato:
@didierofrivia, what was the exact command you used to generate the mocks? Shall we include a make target for it (e.g. `make mocks`)?

I was also thinking about the destination of the mock package. WDYT about `github.com/3scale-labls/authorino/pkg/common/testing`?

guicassolato (approved):

PR #70, opened 2021-03-24
Credentials in cookies

didierofrivia:
This PR extends the feature introduced by https://github.com/3scale-labs/authorino/pull/58 with the option of sending credentials in a Cookie header as described in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie

The config for this would look like:

```yaml
...
identity:
    - name: vip
      api_key:
        label_selectors:
          authorino.3scale.net/managed-by: authorino
          target: echo-api
     credentials:
        in: cookie
        key_selector: API-KEY
...
```

### DIFF
--- a/examples/echo-api-key-simple.yaml
+++ b/examples/echo-api-key-simple.yaml
@@ -8,7 +8,7 @@ spec:
   identity:
     - name: friends # user defined
       credentials:
-        in: custom_header # options include: custom_header, authorization_header and query
+        in: custom_header # options include: custom_header, authorization_header, cookie and query
         key_selector: X-API-KEY # user defined
       api_key:
         label_selectors:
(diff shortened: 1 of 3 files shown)


examples/echo-api-key-simple.yaml
```diff
@@ -8,7 +8,7 @@ spec:
   identity:
     - name: friends # user defined
       credentials:
-        in: custom_header # options include: custom_header, authorization_header and query
+        in: custom_header # options include: custom_header, authorization_header, cookie and query
```
guicassolato, line 11: We need to document this in the OpenAPI spec for the CRD. Let's make it in another PR because there are other fields and options missing in there as well (e.g. the whole part on "json" authorization).

guicassolato (approved):
🏅

PR #73, opened 2021-03-24
Kubernetes authn

### LINKED ISSUE #72: Kubernetes auth
Authorino could leverage Kubernetes authentication resources ([Service Account tokens](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens) in particular) to authenticate client applications ("identity verification" step).

One possible implementation before Kubernetes's [Service Account Issuer Discovery](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery) becomes stable (beta in v1.20) could be by creating `TokenReview` resources (beta in v1.12 and stable since v1.20; [ref](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1)).

With newer versions of Kubernetes, _Service Account Issuer Discovery_ would still be an alternative (v1.20+), leveraging Authorino's already existing OIDC identity verification, however only for JWTs and with more strict constraints regarding the issuer.

guicassolato:
Introduces identity verification based on Kubernetes authentication with Service Account tokens.

Closes #72

----

The implementation leverages Kubernetes [TokenReview API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1) – included in v1.10 (alpha), promoted to beta in v1.12 and stable (GA) in v1.20.

It supports authentication with JWTs (e.g. Service Account tokens) as well as with opaque tokens (other types of Kubernetes access tokens), as long as the token matches the intended "audiences".

Authorino's default service account token is used to create the `TokenReview` resources, but any user with proper permissions to the cluster can request tokens for the clients, including kubelet as part of Kubernetes [Service Account Token Volume Projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection). Client applications running in the cluster can read service account tokens projected into the pods from file (`/var/run/secrets/kubernetes.io/serviceaccount/token`), and then use those to authenticate to APIs protected with Authorino. The service account tokens can be requested or configured to include custom values in the "audiences", such as the required host name of the protected service, thus limiting the scope of the tokens.

**Identity config:**

```yaml
identity:
  - name: trustworthy-service-accounts
    credentials:
      in: authorization_header
      key_selector: Bearer
    kubernetes:
      audiences: # defaults to the host name of the service if omitted
        - echo-api
```

**Resolved identity object:**

The resolved identity object, when authentication succeeds, can be either the claims in the JWT (when the token is a JWT), e.g.

```json
{
  "aud": [
    "echo-api"
  ],
  "exp": 1616592091,
  "iat": 1616591491,
  "iss": "kubernetes.default.svc",
  "kubernetes.io": {
    "namespace": "authorino",
    "serviceaccount": {
      "name": "api-consumer",
      "uid": "eb5bcf55-6bdb-4b5b-a7ae-fbb8ad1119cb"
    }
  },
  "nbf": 1616591491,
  "sub": "system:serviceaccount:authorino:api-consumer"
}
```

...or the Kubernetes "user" returned (when the token is not a valid JWT), e.g.

```json
{
  "username": "system:serviceaccount:authorino:api-consumer-sa",
  "uid": "0632367e-2455-44fb-bd01-3cf5b5f4a34b",
  "groups": [
    "system:serviceaccounts",
    "system:serviceaccounts:authorino",
    "system:authenticated"
  ]
}
```

**Try it out**

Here's an example to try this PR out with a token scoped for the "custom-audience" audience, issued by a "token issuer" service account with permission to use Kubernetes's [TokenRequest API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1).

1. Load the env and deploy the service protection with Kubernetes-auth:
```sh
make local-setup
sed -e "s/\${AUTHORINO_NAMESPACE}/authorino/g" ./examples/echo-api-kubernetes-auth.yaml | kubectl -n authorino apply -f -
```

Expect the following objects created:

```
service.config.authorino.3scale.net/echo-api created                   # the api protection with kubernetes-auth identity verification
serviceaccount/sa-token-issuer created                                 # a service account to request access tokens for other service accounts
clusterrolebinding.rbac.authorization.k8s.io/sa-token-issuer created   # cluster-admin role bound to the sa-token-issuer service account
serviceaccount/api-consumer created                                    # a service account to consume the protected api
```

2. Get a Service Account token issued including 'custom-audience' among the audiences of the token:
```sh
export KUBERNETES_API=$(kubectl cluster-info | head -n 1 | awk '{print $7}' | sed -r "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g")
export TOKEN_ISSUER_TOKEN=$(kubectl -n authorino get secret/$(kubectl -n authorino get sa/sa-token-issuer -o json | jq -r '.secrets[0].name') -o json | jq -r '.data.token' | base64 -d)
export API_CONSUMER_TOKEN=$(curl -k -X "POST" "$KUBERNETES_API/api/v1/namespaces/authorino/serviceaccounts/api-consumer/token" \
     -H "Authorization: Bearer $TOKEN_ISSUER_TOKEN" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{ "apiVersion": "authentication.k8s.io/v1", "kind": "TokenRequest", "spec": { "audiences": ["custom-audience"], "expirationSeconds": 600 } }' | jq -r '.status.token')
```

3. Forward the requests to port 8000 in your localhost to the envoy service running in the cluster

```sh
```

4. Send a request to the protected service using the token to authenticate:

```sh
curl -H 'Host: echo-api' -H "Authorization: Bearer $API_CONSUMER_TOKEN" http://localhost:8000/hello -v
```


pkg/config/identity/kubernetes_auth.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+type KubernetesAuth struct {
+	auth_credentials.AuthCredentials
+	kubernetesAuthDetails
+}
+
+func NewKubernetesAuthIdentity(authCred auth_credentials.AuthCredentials) *KubernetesAuth {
+	config, _ := rest.InClusterConfig()
+	k8sClient, _ := kubernetes.NewForConfig(config)
```
didierofrivia, line 31: What are the chances of hitting a nasty error?

guicassolato: The exact chance I wouldn't know, but it's possible. If somebody messes with some configs of the cluster, happens to override some cluster-reserved env vars within Authorino container, etc...

I will add a safe-guard clause here so it doesn't panic in line 37.

Thanks.

guicassolato: Done

pkg/config/identity/kubernetes_auth.go
```diff
@@ -0,0 +1,76 @@
...
+		} else {
+			resultStatus := result.Status
+			if resultStatus.Authenticated {
+				claims := jwt.MapClaims{}
+				// returns the jwt claims (if the token is in fact a jwt); otherwise, returns the "user" in the TokenReview status
+				if token, _, err := new(jwt.Parser).ParseUnverified(reqToken, claims); err != nil { // we don't care about verifying the jwt because it has been authenticated ("reviewed") already
+					return resultStatus.User, nil
+				} else {
+					return token.Claims, nil
+				}
```
didierofrivia, line 77: Extract to func `tokenInfo` or `parseTokenInfo` or `identityInfo`, etc ?

pkg/config/identity/kubernetes_auth_test.go
```diff
@@ -0,0 +1,182 @@
...
+				},
+			},
+			Audiences: audiences,
+		},
+	}, nil
+}
+
+type k8sAuthenticationClientMock struct {
+	tokenReviewData
+}
```
didierofrivia: Another way would be to define a local interface with the exact method we require from the k8s authenticator, initialize it with the real one in the package and then replace it in the tests. It could be done in a following refactor ;)

pkg/config/identity/kubernetes_auth.go
guicassolato: I guess we could if you think it would improve readability, though this is the core function of the identity Call, i.e. translating a valid token into a JSON-represented identity object.

didierofrivia: Not essential really, it could be done when refactoring the authenticator interface

didierofrivia (approved):
Minor coding improvements that could be addressed in a future refactor. Great to have yet another AuthN method, good job!

didierofrivia (approved):
Great! Thanks for including the changes suggested in this PR <3 Great work! 🥇

PR #74, opened 2021-03-25
Auth Credentials: Default values

didierofrivia:
Closes https://github.com/3scale-labs/authorino/issues/71

* default KeySelector is `Bearer`
* default In (location) is `authorization_header`

### DIFF
--- a/controllers/service_controller.go
+++ b/controllers/service_controller.go
@@ -106,10 +106,7 @@ func (r *ServiceReconciler) translateService(ctx context.Context,
 			Name: identity.Name,
 		}
 
-		authCred := &auth_credentials.AuthCredential{
-			KeySelector: identity.Credentials.KeySelector,
-			In:          identity.Credentials.In,
-		} // TODO: prepare for when missing credentials field
+		authCred := auth_credentials.NewAuthCredential(identity.Credentials.KeySelector, identity.Credentials.In)
 
 		switch identity.GetType() {
 		// oidc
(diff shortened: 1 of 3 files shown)


guicassolato:
@didierofrivia, do we need to change something here as well maybe?

https://github.com/3scale-labs/authorino/blob/578b9a7f7e6ad1ce8c8fcaff4f98699a83adbacc/controllers/service_controller.go#L109-L112

controllers/service_controller.go
```diff
@@ -106,10 +106,8 @@ func (r *ServiceReconciler) translateService(ctx context.Context,
 			Name: identity.Name,
 		}
 
-		authCred := &auth_credentials.AuthCredential{
-			KeySelector: identity.Credentials.KeySelector,
-			In:          identity.Credentials.In,
-		} // TODO: prepare for when missing credentials field
+		// TODO: prepare for when missing credentials field
+		authCred := auth_credentials.NewAuthCredential(identity.Credentials.KeySelector, identity.Credentials.In)
```
guicassolato, line 110: I think the problem is that `identity.Credentials` can be `nil`, no?

didierofrivia: Even if the struct `Identity` has `	Credentials Credentials 'json:"credentials,omitempty"'` the Crendentials will never be `nil`, since in Golang structs don't have a zero value (pointers do).

I'll remove the TODO

guicassolato (approved):

PR #76, opened 2021-03-26
UMA metadata independent from OIDC identity config

### LINKED ISSUE #75: Make UMA metadata independent from OIDC identity config
There is no need for Authorino's UMA metadata evaluator to depend on an OIDC identity config. In cases when they both point to the same auth server, the only actual config shared between them is the endpoint, thus adding no real benefit the coupling of these two configs. Instead, UMA metadata config should accept `endpoint` and become completely independent from the OIDC one.

The UMA metadata evaluator will still discover the UMA config from the well-known endpoint (which should be done in creation time and cached by the way; not in request time), and continues with rest of the resource data fetching from the UMA resource registry, using the client credentials stored to authenticate, etc.

guicassolato:
Changes the UMA metadata config from this:

```yaml
identity:
  - name: auth-server
    oidc:
      endpoint: http://auth-server
metadata:
  - uma:
    identitySource: auth-server
    credentialsRef:
      name: umacredentialssecret
```

to this:

```yaml
metadata:
  - uma:
      endpoint: http://auth-server
      credentialsRef:
        name: umacredentialssecret
```

Closes #75

----

We didn't have tests for the UMA metadata module yet, so the PR also introduces some. Together with the new tests, there is also a new HTTP server mock helper that makes it easy to map expected HTTP request paths and correspondig responses.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	google.golang.org/genproto v0.0.0-20191108220845-16a3f7862a1a
 	google.golang.org/grpc v1.27.0
 	gopkg.in/square/go-jose.v2 v2.5.1 // indirect
-	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gotest.tools v2.2.0+incompatible
 	k8s.io/api v0.18.6
 	k8s.io/apimachinery v0.18.6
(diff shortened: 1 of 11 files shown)


go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	google.golang.org/genproto v0.0.0-20191108220845-16a3f7862a1a
 	google.golang.org/grpc v1.27.0
 	gopkg.in/square/go-jose.v2 v2.5.1 // indirect
-	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v2 v2.4.0 // indirect
```
didierofrivia, line 20: I've seen this change going back and forth in other PRs

didierofrivia (approved):

go.mod
guicassolato: Yeah 😒

PR #80, opened 2021-03-30
make local-setup speed up

guicassolato:
Improves the performance of `make local-setup` by preloading images into the cluster whenever available in the local container registry. Otherwise, images such as keycloak image, envoy image and echo-api image will all be pulled again from the remote registries. In some cases (especially with bad internet connection), pulling those images from remote can take a long time, even causing the waiting for "all deployments to be up" to time out. If the tags are available locally, it is easier and faster just to load them into the cluster container registry. On the other hand, if any image is not available in the local registry, the script will not preload that image and kind will pull it (as before) from remote into the cluster registry.

The PR also introduces a new environment variable parameter `SKIP_BUILD`, which allows setting up the local cluster without re-building the authorino image. This in case all you want is to setup the environment with a version of the `authorino:devel` image that you have already tagged locally from before. Of course, if the local tag is missing, there is no fallback in this case and you should re-run the `make local-setup` command without setting the `SKIP_BUILD` env.

### DIFF
--- a/examples/dex.yaml
+++ b/examples/dex.yaml
@@ -1,3 +1,36 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: dex
+data:
+  config.yaml: |
+    issuer: http://dex:5556
+    storage:
+      type: sqlite3
+      config:
+        file: /etc/dex/db/dex.db
+    web:
+      http: 0.0.0.0:5556
+    oauth2:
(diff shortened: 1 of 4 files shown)


guicassolato:
I've revisited this and it turns out we don't really need it. It doesn't always "speed up" the deployment and it starts complicating things unnecessarily. If we really want to skip a few steps,  we can always run each step manually.

PR #95, opened 2021-05-13
Remove the port part from the host header when looking for a cache service config.

jmprusi:
When trying to authorize a request with HOST:PORT (ex: petstore.127.0.0.1.nip.io:8080) Authorino fails to find a service config that matches. 

This PR tries to retrieve the service config with the incoming Host header, and if that fails, removes the port part (if it exists) from the host header and tries to get a new config.

### DIFF
--- a/pkg/service/auth.go
+++ b/pkg/service/auth.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"encoding/json"
+	"strings"
 
 	"golang.org/x/net/context"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


guicassolato (changes requested):
We will need to fix at least one of the tutorials: https://github.com/Kuadrant/authorino/tree/master/docs/tutorials/showcase

pkg/service/auth.go
```diff
@@ -47,8 +48,15 @@ func (self *AuthService) Check(ctx context.Context, req *envoy_auth.CheckRequest
 	}
 	authServiceLog.Info("Check()", "reqJSON", string(reqJSON))
 
-	// service config
-	host := req.Attributes.Request.Http.Host
+	// If the host header contains a port, remove it, we only want to match the host part.
+	var host string
+	if strings.Contains(req.Attributes.Request.Http.Host,":") {
```
guicassolato, line 53: Should we avoid caching the port number as well (if included in `hosts`)?

jmprusi: yup, I will add it to this PR.

guicassolato:
I wish we had some more input on this. According to [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-5.4), the `Host` header can include the port number, and we are using the host info to lookup the API protection. I am a bit concerned that this change could remove from developers a legit way to distinguish between resources.

pkg/service/auth.go
```diff
@@ -47,8 +48,10 @@ func (self *AuthService) Check(ctx context.Context, req *envoy_auth.CheckRequest
 	}
 	authServiceLog.Info("Check()", "reqJSON", string(reqJSON))
 
-	// service config
-	host := req.Attributes.Request.Http.Host
+	// If the host header contains a port, remove it, we only want to match the host part.
+	splitHost := strings.Split(req.Attributes.Request.Http.Host, ":")
```
guicassolato, line 52: What if we rolled back the change in the Service controller (keeping the key as stated by the developer), and here we first attempted a lookup with the exact host info passed by the proxy (including port number if present); if it doesn't find one, we then tried again without the port number?

guicassolato (approved):
Thanks, @jmprusi!

PR #96, opened 2021-05-17
Makes the secret label configurable via ENV var

jmprusi:


main.go
```diff
@@ -93,11 +94,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	secretLabel := os.Getenv("AUTHORINO_SECRET_LABEL")
```
guicassolato, line 97: We have a func to fetch env vars with default value (see line 147).

jmprusi: Ok, just take into account that if the value is set, but empty, LookupEnv will return Ok. I can reuse this function if you want.

guicassolato (approved):

PR #101, opened 2021-05-31
Better support for multiple Authorino instances

### LINKED ISSUE #105: Leader election breaks the reconciliation of the cache
**Describe the bug**
When leader election is enabled and multiple instances of Authorino are running, only the elected leader instance will watch the `(config.authorino.3scale.net) Service` and `(v1) Secret` resources, causing only this instance to update the internal cache of auth configs in the reconcile loop. Consequently, whenever another instance (no-leader) receives traffic on the external authorization endpoint, the cache is empty and the instance returns with a cache miss (404).

**Help us Reproduce it**

1. Deploy authorino having the number of replicas in the `authorino-controller-manager` `Deployment` greater than 1 (e.g. 2) and the leader election flag enabled;
2. Create a `(config.authorino.3scale.net) Service` custom resource;
3. Check the logs of the manager containers and verify that only one instance reconciled the resource;
4. Starting enough traffic to the protected API to verify that, whenever the traffic is directed to the manager that reconciled the resource, the instance kicks the auth pipeline as expected, whilst on requests directed to the other instances always result in a `404 Service not found` response.

**Expected behavior**

All Authorino instances should reconcile the cache and trigger the auth pipeline, regardless of the state of the leader election flag.

**Additional context**

The observed behaviour is part of how leader election works in the Kubernetes operator. See https://docs.openshift.com/container-platform/4.7/operators/operator_sdk/osdk-leader-election.html for more information.

### LINKED ISSUE #106: Support for cluster-wide and namespaced instances
By default, Authorino instances watch CRs and Service resources in all namespaces (i.e. cluster-wide deployment). However, some users may prefer limiting the scope of the Authorino instance to a given namespace where the resources are defined and no other. To do that, right now there's no support for such case. Even if users manually edit the Authorino `Deployment` and RBAC definitions before deploying an instance, the manager will still not set the `Namespace` option.

A usual pattern to implement operators that are scoped to a given namespace is to define an environment variable, e.g. `WATCH_NAMESPACE`, that is used to set the `Namespace` option of the manager. Of course, the role bindings must follow, giving to the `ServiceAccount` permissions that are as well limited to specified namespace.

### Problem #&#8203;1
One problem that may arise from supporting simultaneously both modes to deploy Authorino (i.e. cluster-wide and namespaced) is that cluster-wide instances will watch all resources, including the ones also watched by other namespaced instances running in the same cluster.

Of course, information cached by any instance of Authorino will only take effect if that instance is in the route for the external authorization of the specified hosts. In other words, in a envoy reverse-proxy integration-scenario, a cluster-wide Authorino instance may cache an auth config, but it won't ever enforce such config to protect a service, unless Envoy sets up that very same instance for the external authorization filter.

### Problem #&#8203;2
Another thing to take into account is that, with support for having multiples instances of Authorino running and scoped in different namespaces, the version of the CRD installed by the cluster admin is always cluster-wide. Users must be careful not to mismatch Authorino instance version and CRD version, as well as cluster admins must be aware of this whenever rolling out updates on the CRD.

### LINKED ISSUE #107: User can surpersede the protection of an API by using the same host
On cluster-wide deployments of Authorino, i.e. where the Authorino controller watches CRs and Secrets defined across all the cluster rather than in a single namespace, a user with permissions to create Authorino `Service` custom resources can define an API protection listing amongst the `hosts` a host already taken by a different `Service` CR. The Authorino cluster-wide instance that reconciles the newly created CR, will update the cache of auth configs replacing the entry previously taken by the formerly created CR with the auth config stated in the new CR.

### Steps to reproduce

1. Deploy Authorino (cluster-wide)
2. Create an Authorino `Service` CR `x` that includes a host, e.g., `abc.api`
3. Sets Envoy to filter traffic sent to `abc.api` using the Authorino instance as external authorization service
4. Create another Authorino `Service` CR `y`, with a different auth spec (e.g. different identity sources, diff authz policies), but that also includes the host `abc.api`
5. Sends request to `abc.api` and verifies that Authorino `Service` CR `y` was enforced rather than Authorino `Service` CR `x`

### Expected behavior

Authorino should enforce some protection against host collision. E.g. `Service` CRs including host names already taken could be rejected. Perhaps this could be some policy enforced only when different namespaces are involved, as a way to protect user definition from the actions of other users in other user spaces.

#### Example

Consider the following resources:

```yaml
# authorino controller deployed cluster-wide in the `system` namespace
apiVersion: v1
kind: Pod
metadata:
  namespace: system
  name: authorino-controller-1
```

```yaml
# authorino CR `foo` declared in the `x` namespace using host `example.com`
apiVersion: config.authorino.3scale.net/v1beta1
kind: Service
metadata:
  namespace: x
  name: foo
spec:
  hosts:
    - example.com
```

```yaml
# authorino CR `bar` declared in the `y` namespace using host `example.com`
apiVersion: config.authorino.3scale.net/v1beta1
kind: Service
metadata:
  namespace: y
  name: bar
spec:
  hosts:
    - example.com
```

Then (in order):
1. User applies `x/foo` which is **accepted** by the Authorino controller that caches `example.com → x/foo`
2. User applies `y/bar` which is **rejected** by the Authorino controller (otherwise, the owner of `y/bar` would take control over the protection for `example.com`)

#### Problems that may still arise from the proposed approach (based on the example above)

**Problem #&#8203;1:** What if `y/bar` is the legit owner of `example.com`?

This is not an enormous problem because, in order for Authorino to apply the protection over `example.com`, requests to this host must be proxied to Authorino which can only be set by the legit owner of the ingress route.

The only actual problem here is the locking of a host with a stale `Service` config. In case of host collision, the illegitimate holder of a host with a stale `Service` config will have to be requested to release the lock.

**Problem #&#8203;2:** What if instance A of Authorino caches `x/foo` and instance B of Authorino caches `y/bar` in a race condition?

This lead to an inconsistent state where each instance of Authorino knows about a different protection to apply over `example.com`, and the outcome of requests to the protected API is random, as we cannot cannot assume about the order of reconciliation.

A possible solution could be implementing a **centralised cache** although that would introduce a bottleneck. Another possible solution could be through **syncing the cache** across Authorino instances → hard to implement and no actual guarantee of a consistent state.

> Refer to #105 for a discussion about problems in Authorino related to cache reconciliation under multiple instances of the external authorization service and leader election enabled.

**Problem #&#8203;3:** Depending on how Authorino "rejects" a `Service` resource (as for restrictions of the OPerator SDK), this might only mean that the corresponding spec will not be reconciled (i.e. it will not be added to the cache), however the `Service` resource might still be accepted by the API server and remain in the data plane (without status).

> Check out Kubernetes [**ValidatingAdmissionWebhook**](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) to prevent that.

guicassolato:
This PR gathers changes for better support for multiple Authorino instances running on the same cluster, including the following:
- Reconciliation of the cache by all instances regardless of leader election (closes #105)
- Support for cluster-wide and namespaced instances (closes #106)
- Protection against host collision (closes #107)

----

### Cluster-wide vs. namespaced Authorino deployments

With this PR, users can chose between 2 supported modes to deploy Authorino:
1. **cluster-wide:** instances watch Authorino `Service` and `Secret` resources in all namespaces;
2. **namespaced (default):** instances watch Authorino `Service` and `Secret` resources defined in one particular user space.

> **Note:** _Deployment_ here refers to running an Authorino pod that watches and reconciles Authorino `Service` CRs and associated `Secrets`, exposes a `/metrics` endpoint and endpoints of the Festival Wristband OIDC configuration. Deploying may or may not require admin privileges over the Kubernetes cluster, depending on the mode chosen. It differs from _installing_ Authorino, which always requires admin privileges over the cluster, and refers to the step of defining the Authorino CRD and base RBAC cluster roles. Installing Authorino to the cluster is done only once (by cluster admin), while deploying can be done any number of times (by cluster admins and cluster users).
>
> In the near future, I plan to improve this with operator bundles, implementing for the [Operator Lifecycle Manager (OLM)](https://olm.operatorframework.io).

For cluster-wide deployment, the controller will reject `Service` resources containing at least one host already taken by another `Service` resource in a different namespace, and therefore already cached by the Authorino instance for a different tenant of the cluster. This will prevent host collision across different user spaces, thus enabling safe multi-tenancy on such mode of deploying of Authorino. In this deployment mode, `ClusterRoleBinding`s are used to provider the `default` `ServiceAccount` with the required permissions (leader election, reconciler, metrics endpoint, CR management), therefore only admins of the Kubernetes cluster will be able to deploy Authorino in cluster-wide mode. 

For namespaced deployments, the controller expects a `WATCH_NAMESPACE` environment variable with value set to the name of the namespace of the Authorino instance, which must be also the same namespace where the user will define the Authorino `Service` CRs and associated `Secret`s. In this deployment mode, `RoleBinding`s are used to provider the `default` `ServiceAccount` with the required permissions (leader election, reconciler, metrics endpoint, CR management).

### Protection against host collision

Consider the following resources:

```yaml
# authorino controller deployed cluster-wide in the `system` namespace
apiVersion: v1
kind: Pod
metadata:
  namespace: system
  name: authorino-controller-1
```

```yaml
# authorino CR `foo` declared in the `x` namespace using host `example.com`
apiVersion: config.authorino.3scale.net/v1beta1
kind: Service
metadata:
  namespace: x
  name: foo
spec:
  hosts:
    - example.com
```

```yaml
# authorino CR `bar` declared in the `y` namespace using host `example.com`
apiVersion: config.authorino.3scale.net/v1beta1
kind: Service
metadata:
  namespace: y
  name: bar
spec:
  hosts:
    - example.com
```

Then (in order):
1. User applies `x/foo` which is **accepted** by the Authorino controller that caches `example.com → x/foo`
2. User applies `y/bar` which is **rejected** by the Authorino controller (otherwise, the owner of `y/bar` would take control over the protection for `example.com`)

The are, however, still a few problems with this approach:

**Problem #&#8203;1:** What if `y/bar` is the legit owner of `example.com`?

This is not an enormous problem because, in order for Authorino to apply the protection over `example.com`, requests to this host must be proxied to Authorino which can only be set by the legit owner of the ingress route.

The only actual problem here is the locking of a host with a stale `Service` config. In case of host collision, the illegitimate holder of a host with a stale `Service` config will have to be requested to release the lock.

**Problem #&#8203;2:** What if instance A of Authorino caches `x/foo` and instance B of Authorino caches `y/bar` in a race condition?
- This can lead to an inconsistent state where each instance of Authorino knows about a different protection to apply over `example.com`, and the outcome of requests to the protected API is random
- We cannot assume about the order of reconciliation
- A possible solution could be implementing a **centralised cache** → introduces a bottleneck
- Another possible solution could be through **syncing the cache** across Authorino instances → hard to implement and no actual guarantee of a consistent state
- Perhaps the leader could help avoiding the conflict (?)

**Problem #&#8203;3:** Authorino "rejecting" a `Service` resource only means that the corresponding spec will not be reconciled (i.e. it will not be added to the cache), however the `Service` resource is still accepted by the API server and remains in the data plane (without status).
- We could use Kubernetes [**ValidatingAdmissionWebhook**](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) instead → requires another HTTP service/endpoint.

----

### TODOs
- [x] Split manifests in `install` and `deploy` so it's clear what belongs to cluster admins and what belongs to cluster users
- [x] `cluster-wide` and `namespaced` kustomize overlays
- [x] Support for the `WATCH_NAMESPACE` env var in the controller
- [x] Remove Operator SDK bundle-related stuff
- [x] Host collision protection across different namespaces
- [x] Refactor `Cache` struct (replace it with an interface for better testing)
- [x] Fix make targets to build, deploy and try in a local cluster
- [x] Fix resource status
    Screenshot 2021-06-07 at 19 17 21
- [x] Remove leader election for the controllers/reconcilers that update the cache (otherwise when leader election is enabled only the leader will update the cache)
- [x] Define a separate controller to take care of resource status update (option to enable leader election)
- [x] Update the examples and the docs


guicassolato:
This might be useful for https://github.com/operate-first/support/issues/253.

Makefile
```diff
@@ -129,8 +134,8 @@ endif
 .PHONY: bundle
 bundle: manifests kustomize
 	operator-sdk generate kustomize manifests -q
-	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
-	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
+	cd deploy/base && $(KUSTOMIZE) edit set image authorino=$(AUTHORINO_IMAGE)
+	$(KUSTOMIZE) build install/crd | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
```
miguelsorianod, line 138: I understand your point about separating the "install" from the "deploy".
The `config` directory is part of the standard layout provided by the `operator-sdk` tool.
Potential issues I can see with separating the directories is that operator-sdk tool might have some assumptions on the directory and/or all configs being in the same parent directory. For example, this same `bundle` Make target is not properly generating the bundle now: you will see for example that the role permissions are not included in the generated bundle CSV, nor the CRDs. To fix this a workaround would need to be figured out.
There might be also another issues as a consequence of this that we are unaware of at the moment. For example quickly thinking I know there are CSV annotations specific to operator-sdk and I wonder if it might make assumptions on directory locations from that side.

In general in an operator-sdk based project moving away from the standard layout seems problematic.

Makefile
```diff
@@ -153,29 +158,73 @@ else
...
+	echo "KUSTOMIZE=$(KUSTOMIZE)"; \
+	echo "AUTHORINO_IMAGE=$(AUTHORINO_IMAGE)"; \
+	echo "AUTHORINO_NAMESPACE=$(AUTHORINO_NAMESPACE)"; \
+	echo "AUTHORINO_DEPLOYMENT=$(AUTHORINO_DEPLOYMENT)"; \
+	}
+
+# Creates a namespace where to deploy Authorino
+.PHONY: namespace
+namespace:
+	kubectl create namespace $(AUTHORINO_NAMESPACE)
```
miguelsorianod, line 175: This was originally handled by the `deploy` Makefile task as the `config` directory included the autogeneration of the namespace

Makefile
```diff
@@ -153,29 +158,73 @@ else
...
 .PHONY: local-push
 local-push: kind
-	kind load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
+	kind load docker-image $(AUTHORINO_IMAGE) --name $(KIND_CLUSTER_NAME)
+
+# Builds the image, pushes to the local cluster and deployes Authorino.
+# Sets the imagePullPolicy to 'IfNotPresent' so it doesn't try to pull the image again (just pushed into the server registry)
+.PHONY: deploy
+local-deploy: docker-build local-push deploy
+	kubectl -n $(AUTHORINO_NAMESPACE) patch deployment authorino-controller-manager -p '{"spec": {"template": {"spec":{"containers":[{"name": "manager", "imagePullPolicy":"IfNotPresent"}]}}}}'
```
miguelsorianod, line 210: Has this been tried?
The issue with setting `ImagePullPolicy` to `IfNotPresent` is that if the cluster is already running the new image will not be used because the image will have the same tag as before.

Makefile
guicassolato: Not really. Although the `config` included a `Namespace` resource, due to the use of kustomize – particularly kustomize's `namespace` and `namePrefix` directives –, the `Namespace` resource included in `config` was completely useless. `deploy` ended up creating a namespace `authorino-system` and adding metadata of all resources `namespace: authorino`. The tools we added afterwards to deploy locally had to create the `authorino` namespace and `authorino-system` was never used.

Makefile
guicassolato: It has been tried and it works. This is because `local-deploy` builds the image and pushes it to the authorino kind cluster. As the image is tagged `latest` by default, with `ImagePullPolicy`, Kubernetes tries to pull the image, which obviously fails, despite the image being present within the cluster already. So we patch setting the policy to `IfNotPresent`.

Makefile
miguelsorianod: Wouldn't changing the `namespace` value in `config/default/kustomization.yaml` would have fixed the issue you comment

Makefile
miguelsorianod: :+1: . I guess it fails because it tries to actually go to the repository. on the case of the default `Always` pullPolicy

Makefile
guicassolato: I think we need a concrete reason why we cannot customize our file tree to make it more obvious for the users (who are not necessarily Operator SDK experts, nor they should be required to be IMO) and to make our extended tools simpler and deterministic for the different deployment modes we want to support. If it's all adjustable by adapting the Makefile, then I don't see why not.

Right now, `make generate`, `make manifests` and pretty much all other necessary targets to build, install, deploy, locally or on whatever other Kubernetes server, all work. If `make bundle` is broken and there's no way to fix it with the manifests output to a dir other than `/config`, then this would be a concrete reason. Before assuming that though, I'll try to fix the `bundle` make target. If I cannot fix it, then I'll rollback to having all manifests under `/config`; otherwise, I still prefer separating install and deploy parts, until we are forced to rethink this, when we start using Operator bundles.

miguelsorianod: :+1:

miguelsorianod:
As a side note,

as another deployment method, OLM kind of covers this kind of deploy / install scenarios.
With `ClusterServiceVersion` and `OperatorGroups` some annotations are automatically set in the CSV when installing the operator and then the `WATCH_NAMESPACE` environment variable can be set to reference that annotation. In case of wanting to offer all kinds of deployments I'm guessing different bundles / set of files might be needed however because in one case a ClusterRole is used and in another a Role is used. It would be interesting to see how would that be implemented.

Of course, this implies that you need OLM installed on the K8S cluster and install the operator using an index image as it's usually done when working with OLM.

miguelsorianod:
> Problem #​2: What if instance A of Authorino caches x/foo and instance B of Authorino caches y/bar in a race condition?

I'm thinking about this issue.

Why would there be two instances of Authorino with the current implementation (where all the authorino logic is part of the operator pod itself)?
If Authorino is deployed to watch all namespaces by definition there should be only one Authorino deployment.
If Authorino is deployed to watch a single namespace there should only be one deployment per namespace.

A "cluster-wide" Authorino deployment should never be deployed together with a "namespace" specific Authorino deployment because the cluster-wide Authorino deployment will deal with all namespaces.

Makefile
guicassolato: I guess if we set `namespace: authorino-system` in `config/default/kustomization.yaml`, that would do the trick. However, having a new namespace created to deploy Authorino not always makes sense.

For cluster-wide deployment, I'd say having an "authorino" (or "authorino-system") namespace is desirable. For namespaced deployment, on the other hand, that makes less sense, as the namespace will usually be something defined by the user, eventually created beforehand with the APIs to be protected already running in there.

Anyway, this dichotomy made me want to remove the `Namespace` resource and leave it for the user to decide.

miguelsorianod: To cover the use case of having a custom namespace name I think the `kustomize edit set` subcommand might fill that purpose.

miguelsorianod:
Does adding the in-memory cache solve the issue?

I understand that if the pod is restarted the cache is lost, as it is in-memory. Thus the following could happen:
* User applies x/foo which is accepted by the Authorino controller that caches example.com → x/foo
* Controller pod is restarted (due to failure or upgrade for example)
* user applies y/bar which would be accepted by the Authorino controller because x/foo was not in the cache anymore

guicassolato:
> Why would there be two instances of Authorino with the current implementation (where all the authorino logic is part of the operator pod itself)?

I meant 2+ Authorino pods.

> A "cluster-wide" Authorino deployment should never be deployed together with a "namespace" specific Authorino deployment in the same cluster because the cluster-wide Authorino deployment will deal with all namespaces.

You are right about the part where the cluster-wide instance will deal with CRs created in all namespaces. However, in order for those CRs to be useful, Authorino still has to be hooked as the external authorization filter on requests to the corresponding hosts. I don't see why one would want this, but it is technically possible to have a scenario, for example, where a cluster-wide Authorino instance _X_ reconciles (caches) CRs _A_, _B_ and _C_, and another namespaced Authorino instance _Y_ that only knows about CR _C_. An instance of Envoy can be configured to use Authorino _X_ on requests to APIs whose hosts are listed in CRs _A_ and _B_, and another instance of Envoy can be configured to use Authorino _Y_ on requests to API whose host(s) are listed  in CR _C_. In such case, the caching of CR _C_ by Authorino _A_ is useless, but the overall topology would still work.

miguelsorianod:
> > Why would there be two instances of Authorino with the current implementation (where all the authorino logic is part of the operator pod itself)?
> 
> I meant 2+ Authorino pods.

On an operator controller deployment even when there are more than 1 pods only one pod, the leader, performs reconciliation

> > A "cluster-wide" Authorino deployment should never be deployed together with a "namespace" specific Authorino deployment in the same cluster because the cluster-wide Authorino deployment will deal with all namespaces.
> 
> You are right about the part where the cluster-wide instance will deal with CRs created in all namespaces. However, in order for those CRs to be useful, Authorino still has to be hooked as the external authorization filter on requests to the corresponding hosts. I don't see why one would want this, but it is technically possible to have a scenario, for example, where a cluster-wide Authorino instance _X_ reconciles (caches) CRs _A_, _B_ and _C_, and another namespaced Authorino instance _Y_ that only knows about CR _C_. An instance of Envoy can be configured to use Authorino _X_ on requests to APIs whose hosts are listed in CRs _A_ and _B_, and another instance of Envoy can be configured to use Authorino _Y_ on requests to API whose host(s) are listed in CR _C_. In such case, the caching of CR _C_ by Authorino _A_ is useless, but the overall topology would still work.

In that case only the cluster-wide instance of authorino should be used, and different envoys should be deployed all pointing to the cluster-wide authorino. Having a cluster-wide Authorino instance and a namespace-specific one will be problematic as both will compete to reconcile the same resources.

miguelsorianod:
> > Does adding the in-memory cache solve the issue?
> 
> The in-memory cache is rebuilt when the pod starts again, from listing again all existing CRs in the watched namespace/all namespaces and reconciling them all over.

If you are referring to https://github.com/Kuadrant/authorino/pull/101/files#diff-00e81e75f6d3a8e448c8698e8838e914e508f7bb66e569da6fdd3d394baf96e5R92, that part adds the received Service in a reconcile event to the cache.
That part adds the Service that has been received as part of a reconcile event of a given `Service`, but if there's no event for the Service then it's not added to the cache.
Maybe when the controller is restarted the operator library automatically triggers an event for each existing CR? has this been observed? If that happens, even in that case there would be no guarantee of the order so I guess my scenario might still happen right? the order of events is not guaranteed

guicassolato:
> Maybe when the controller is restarted the operator library automatically triggers an event for each existing CR?

I don't know about the details of how it works TBH, but it has been observed. In fact, we have been using this since always in the development cycle, for example, when we use the `make local-rollout` target. This target rebuilds the authorino image, pushes it to the cluster and rolls out the deployment. When the new pod is up, it rebuilds the cache automatically

miguelsorianod:
> I don't know about the details of how it works TBH, but it has been observed. In fact, we have been using this since always in the development cycle, for example, when we use the make local-rollout target. This target rebuilds the authorino image, pushes it to the cluster and rolls out the deployment. When the new pod is up, it rebuilds the cache automatically

The only scenario I can think of is that it triggers events for each CR when a pod is restarted.
But again, in that case, the following race condition could happen:

1. User applies x/foo which is accepted by the Authorino controller that caches example.com → x/foo
1. Controller pod is restarted (due to failure, upgrade or change of leader pod for example)
1. User applies y/bar which could be accepted by the Authorino controller before the event for x/foo arrives that makes it being added to the cache

Probably the race condition is unlikely but it could happen.

Issue #116, opened 2021-06-15
Filter `Secret` resources watched by Authorino `SecretReconciler`


guicassolato:
Use https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate to filter the events related to Kubernetes `Secret` resources, watched by `SecretReconciler`, only to resources labelled with the Authorino "managed-by" label:

https://github.com/Kuadrant/authorino/blob/6b44f9552fc97b7836137954fa26c19da71aac0c/main.go#L54

eguzki:
This example used in [kuadrant-controller](https://github.com/Kuadrant/kuadrant-controller) is doing pretty much the same, but with services instead of the secrets.

```go
// SetupWithManager sets up the controller with the Manager.
func (r *ServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		WithEventFilter(onlyLabeledServices()).
		For(&corev1.Service{}).
		Complete(r)
}

func onlyLabeledServices() predicate.Predicate {
	return predicate.Funcs{
		CreateFunc: func(e event.CreateEvent) bool {
			// Lets filter for only Services that have the kuadrant label and are enabled.
			if val, ok := e.Object.GetLabels()[kuadrantDiscoveryLabel]; ok {
				enabled, _ := strconv.ParseBool(val)
				return enabled
			}
			return false
		},
		UpdateFunc: func(e event.UpdateEvent) bool {
			// In case the update object had the kuadrant label set to true, we need to reconcile it.
			if val, ok := e.ObjectOld.GetLabels()[kuadrantDiscoveryLabel]; ok {
				enabled, _ := strconv.ParseBool(val)
				return enabled
			}
			// In case that service gets update by adding the label, and set to true, we need to reconcile it.
			if val, ok := e.ObjectNew.GetLabels()[kuadrantDiscoveryLabel]; ok {
				enabled, _ := strconv.ParseBool(val)
				return enabled
			}

			return false
		},
		DeleteFunc: func(e event.DeleteEvent) bool {
			// If the object had the Kuadrant label, we need to handle its deletion
			_, ok := e.Object.GetLabels()[kuadrantDiscoveryLabel]
			return ok
		},
	}
}
```

Hopefully it is useful.

[...]

PR #124, opened 2021-07-13
Adds a eventFilter to the SecretReconciler to filter events by label

jjaferson:
The SecretReconciler was reconciling at every event fired by any secret in the cluster or in a namespace (depending on the scope authorino was installed) to avoid these unnecessaries loops this PR includes a eventFilter that only fires an SecretReconciler loop when there is a change on a labelled secret.

Addresses issue https://github.com/Kuadrant/authorino/issues/116


guicassolato:
Thanks so much for the PR, @jjaferson! I can try to help you with the tests.

eguzki:
What if I remove the label from the secret? The secret reconciller will not get a request and internal cache will not be updated. 

 Is this use case interesting to tackle?

guicassolato:
> What if I remove the label from the secret? The secret reconciller will not get a request and internal cache will not be updated.

This is a problem we already have today, @eguzki. It's a known bug.

> 
> Is this use case interesting to tackle?

In time, we sure need to fix this. For now, the recommendation for one who wants to revoke access of an API key (uncache it) is to delete the `Secret` altogether.

eguzki:
Good, I just underlined the issue for awareness.

jjaferson:
> Thanks so much for the PR, @jjaferson! I can try to help you with the tests.

I managed to test it yesterday on my cluster and all worked as expected

Screenshot 2021-07-14 at 12 27 51

guicassolato:
> I managed to test it yesterday on my cluster and all worked as expected
> 

That's great, @jjaferson. Let's see if we can come up with some unit testing before merging then.

guicassolato:
To share here another example of this working, in this case, on the local cluster started with `make local-setup`, I've added an info log message to https://github.com/jjaferson/authorino/blob/c61f00f3188d39e3ad00b28ecb9264ad8e30dfed/controllers/secret_controller.go#L67

And then I created a service and 2 secrets, where one of the secrets (`friend-1-api-key-1`) was an Authorino-managed secret (i.e., it included Authorino secret label key) and the other (`friend-1-api-key-2`) was not.

Both secrets showed up in the logs of the controller as "Successfully Reconciled", but there was nothing about `friend-1-api-key-2` being skipped (my added info log message), thus confirming that it never reached the reconciler.

Screenshot 2021-07-15 at 13 15 06

guicassolato (approved):

jjaferson:
> To share here another example of this working, in this case, on the local cluster started with `make local-setup`, I've added an info log message to https://github.com/jjaferson/authorino/blob/c61f00f3188d39e3ad00b28ecb9264ad8e30dfed/controllers/secret_controller.go#L67
> 
> And then I created a service and 2 secrets, where one of the secrets (`friend-1-api-key-1`) was an Authorino-managed secret (i.e., it included Authorino secret label key) and the other (`friend-1-api-key-2`) was not.
> 
> Both secrets showed up in the logs of the controller as "Successfully Reconciled", but there was nothing about `friend-1-api-key-2` being skipped (my added info log message), thus confirming that it never reached the reconciler.
> 
> Screenshot 2021-07-15 at 13 15 06

Thanks for adding the unit testing and verifying the PR again @guicassolato I am happy to get it merged.

PR #135, opened 2021-07-19
(no-)TLS deployment flavor

### LINKED ISSUE #128: Authorino requires cert-manager CRDs even when TLS secrets are present beforehand
**Describe the bug**
If you don't have [cert-manager](https://cert-manager.io/docs/installation/kubernetes/)'s CRDs applied to the cluster, Authorino deployment will fail with the `Issuer` and `Certificate` cert-manager custom resources rejected to be applied. This shouldn't be the case when you are planning to create the TLS `Secret` for Authorino's OIDC service beforehand and therefore wouldn't need cert-manager at all.

**Help us Reproduce it**

Please give us a list of steps we can follow, in order to reproduce the issue:
e.g.
1. `make deploy` Authorino to a cluster without cert-manager CRDs installed
2. see Authorino deployment failing right after `deployment.apps/authorino-controller-manager created` with
    ```
    unable to recognize "STDIN": no matches for kind "Certificate" in version "cert-manager.io/v1"
    unable to recognize "STDIN": no matches for kind "Certificate" in version "cert-manager.io/v1"
    unable to recognize "STDIN": no matches for kind "Issuer" in version "cert-manager.io/v1"
    unable to recognize "STDIN": no matches for kind "Issuer" in version "cert-manager.io/v1"
    ```

**Expected behavior**

Authorino to skip creation of `Issuer` and `Certificate` cert-manager custom resources and proceed with the deployment.

If the `authorino-oidc-server-cert` TLS `Secret` is not present, then the authorino-controller-manager pod will not start and that's expected. However, if the `Secret` is present, then Authorino should start.

guicassolato:
Since TLS has been enabled for Authorino’s built-in OIDC server for wristband tokens (#120), [cert-manager](https://cert-manager.io/) became a requirement. This PR changes it so users who do not want to use cert-manager to manage Authorino TLS certificates can create corresponding TLS secrets beforehand. Moreover, it adds a deployment flag to disable TLS on the endpoints served by Authorino.

Closes #128


main.go
```diff
@@ -204,10 +204,26 @@ func startOIDCServer(serviceCache cache.Cache) {
 			Cache: serviceCache,
 		})
 
-		logger.Info("starting oidc service", "port", oidcHTTPPort)
```
eguzki, line 207: Serving TLS is ruled by the existence of the certificate files. If I do some mistake providing the certificates, authorino will serve plan HTTP instead of failing. Not a big issue, though.

Ideally, I would control serving plain or TLS based HTTP using a env var pointing to the certificate file(s). If the env var does not exist, serve plain HTTP. If it exists, try to serve TLS and the service will fail if referred cert files are not found or not valid.

guicassolato: Adding the env var to control is no big deal. I'll send another PR.
Thanks for suggesting, @eguzki!

eguzki: It was just food for thought for future PR's :)

PR #137, opened 2021-07-20
TLS for the external authorization service

guicassolato:
Adds support for env vars `TLS_CERT` and `TLS_CERT_KEY` to serve external authorization over TLS.

Deploys the example Envoy proxy config according to the TLS state (enabled/disabled) in Authorino.


main.go
```diff
@@ -175,16 +179,31 @@ func startExtAuthServerGRPC(serviceCache cache.Cache) {
...
+		grpcServerOpts := []grpc.ServerOption{grpc.MaxConcurrentStreams(gRPCMaxConcurrentStreams)}
 
-		envoy_auth.RegisterAuthorizationServer(s, &service.AuthService{Cache: serviceCache})
-		healthpb.RegisterHealthServer(s, &service.HealthService{})
+		tlsEnabled := tlsCertPath != "" && tlsCertKeyPath != ""
+		logger.Info("starting grpc service", "port", extAuthGRPCPort, "tls", tlsEnabled)
 
-		logger.Info("starting grpc service", "port", extAuthGRPCPort)
+		if tlsEnabled {
+			if tlsCert, err := tls.LoadX509KeyPair(tlsCertPath, tlsCertKeyPath); err != nil {
```
guicassolato, line 188: We're going with `crypto/tls` here instead of using `openssl` simply because that's what Golang's HTTP server uses and we are already using it for the OIDC wristband validation endpoints. So no two ways of doing the same thing.

Issue #134, opened 2021-07-19
Upgrade envtest


guicassolato:
[controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) introduced a new Envtest setup tool in v0.9.0 (see release note [here](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.9.0)), so Authorino's `make test` should stop downloading `hack/setup-envtest.sh` and switch to `sigs.k8s.io/controller-runtime/tools/setup-envtest`. 

Docs: https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest

jjaferson:
@guicassolato Do we want to use always the latest version of the setup-envtest or stick to an specific version?

guicassolato:
> @guicassolato Do we want to use always the latest version of the setup-envtest or stick to an specific version?

I think it's ok to install latest, and, in case we need a specific version, we can set it in `setup-envtest use`.

PR #140, opened 2021-07-25
Upgrade envtest

jjaferson:
Addresses issue #134 to replace the download of hack/setup-envtest.sh with go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest in the Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,12 +20,9 @@ AUTHORINO_REPLICAS ?= 1
 
 all: manager
 
-# Run tests
-ENVTEST_ASSETS_DIR = $(shell pwd)/testbin
-test: generate fmt vet manifests
-	mkdir -p $(ENVTEST_ASSETS_DIR)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -155,6 +152,22 @@ else
...
 endif
 
+setup-envtest:
+ifeq (, $(shell which setup-envtest))
+	@{ \
+	set -e ;\
+	ENVTEST_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$ENVTEST_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	GO111MODULE="on" go get sigs.k8s.io/controller-runtime/tools/setup-envtest@latest ;\
```
guicassolato, line 162: Can we use `go install` and save the creation of the temp dir maybe?

jjaferson: Sure, I've changed that and specified a version to be used as well

guicassolato (approved):
Nice! Thanks!

PR #142, opened 2021-08-19
Use latest version of Kind

jsmadis:
Upgrades the version of `Kind` to the latest one. This upgrade removes the necessity of the configuration file for the TokenRequest feature.

I successfully tested it with the [example](https://github.com/Kuadrant/authorino/blob/371d0408998f6223b3d7be170704688901647772/examples/README.md#kubernetes-authentication).


Makefile
```diff
@@ -144,7 +144,7 @@ ifeq (, $(shell which kind))
 	KIND_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$KIND_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	GO111MODULE="on" go get sigs.k8s.io/kind@v0.9.0 ;\
+	GO111MODULE="on" go get sigs.k8s.io/kind@v0.11.1 ;\
```
guicassolato, line 147: What do you think about changing the whole target to something like this?

```Makefile
KIND_VERSION=v0.11.1

kind:
ifneq ($(KIND_VERSION), $(shell kind version | cut -d' ' -f2))
	go install sigs.k8s.io/kind@$(KIND_VERSION)
KIND=$(GOBIN)/kind
else
KIND=$(shell which kind)
endif
```

guicassolato: If you have kind installed other than with Go (e.g. with Homebrew in macOS) and the version mismatches the expected one, then this will not help, but neither it would that other implementation that just checks for `which kind`.

jsmadis: It looks good to me, I will change it and test it locally.

guicassolato (approved):

PR #144, opened 2021-08-23
Adds external URL for the OPA authorization policy

### LINKED ISSUE #115: Rego policies pulled from external registry
Right now, Authorino only supports inline Rego policies. The API should be extended to allow users alternatively to indicate in the custom resource a URL of the policy instead. Authorino would then fetch the Rego content from that URL, and proceed with the compilation ("PrepareForEval") of the policy and caching as usual.

jsmadis:
Closes #115 

Adds the possibility to download the Rego document from the external registry.

If the user fills both `InlineRego` and `ExternalURL` we will use only the `InlineRego` document and ignore the `ExternalURL`.


pkg/config/authorization/opa.go
```diff
@@ -19,14 +21,24 @@ default allow = false
 	policyUIDHashSeparator = "|"
 
 	opaPolicyPrecompileErrorMsg = "Failed to precompile OPA policy"
+	regoDownloadErrorMsg        = "Failed to download Rego data"
 	invalidOPAResponseErrorMsg  = "Invalid response from OPA policy evaluation"
 )
 
 var (
-	opaLog = ctrl.Log.WithName("Authorino").WithName("ApiKey")
+	opaLog = ctrl.Log.WithName("Authorino").WithName("OPA")
```
guicassolato, line 29: Nice catch!

pkg/config/authorization/opa.go
```diff
@@ -19,14 +21,24 @@ default allow = false
...
 )
 
-func NewOPAAuthorization(policyName string, rego string, nonce int) *OPA {
+func NewOPAAuthorization(policyName string, rego string, url string, nonce int) *OPA {
+	if rego == "" && url != "" {
+		downloadedRego, err := downloadRegoDataFromUrl(url)
+		if err != nil {
+			opaLog.Error(err, regoDownloadErrorMsg, "secret", policyName)
+			return nil
```
guicassolato, line 37: I think we should return the error in this case, as fetching the policy is essential.

pkg/config/authorization/opa.go
```diff
@@ -81,3 +93,21 @@ func generatePolicyUID(policyName string, policyContent string, nonce int) strin
 	data := []byte(fmt.Sprint(nonce) + policyUIDHashSeparator + policyName + policyUIDHashSeparator + policyContent)
 	return fmt.Sprintf("%x", md5.Sum(data))
 }
+
+func downloadRegoDataFromUrl(url string) (string, error) {
+	if resp, err := http.Get(url); err != nil {
```
guicassolato, line 98: We may need to support some basic authentication between Authorino and the policy registry. HTTP external metadata protocol can be an example of hot to define and implement that, but we should also check with potential policy registry services (e.g. Red Hat Registry), to see the types of authentication that can work out.

pkg/config/authorization/opa.go
jsmadis: What we will do with the error in the service controller? Fail the reconciliation? 
https://github.com/Kuadrant/authorino/blob/9b9b6d7fe742ba90a2af54fbdb2bfffc81d6141f/controllers/service_controller.go#L232

guicassolato: It's either that or trying again before processing the first request.

For OIDC identity, for example, if the OpenID Connect config cannot be fetched in reconciliation time, we try again in request-time. We could do the same here.

...Or as I suggested before. We modify `NewOPAAuthorization` to return `(*OPA, error)` and let the controller to treat it when there's an error (e.g. retry reconciling the resource).

jsmadis: > For OIDC identity, for example, if the OpenID Connect config cannot be fetched in reconciliation time, we try again in request-time. We could do the same here. 

Do you mean inside the `call` function?  


I would go for both options because I was missing the error when my `InlineRego` was faulty.

guicassolato: > Do you mean inside the `call` function?

Yes.

guicassolato: Returning the error is acceptable, but it's either one thing or the other.

If we reject creating the config (because we cannot fetch the policy, or because compiling the policy failed), then so the reconciler should fail and fall back to its default, which is retrying to reconcile the resource later.

On the other hand, if we chose for accepting the config silently, the user will only realize something is wrong when a first request is sent. In the case of OIDC identity, this was my only option because I wanted to support it in combination with wristband configs in the same custom resource.

pkg/config/authorization/opa.go
guicassolato: Example of authentication if the external registry is an OPA server: https://www.openpolicyagent.org/docs/latest/rest-api/#authentication

pkg/config/authorization/opa.go
```diff
@@ -81,3 +93,21 @@ func generatePolicyUID(policyName string, policyContent string, nonce int) strin
...
+		return "", fmt.Errorf("failed to fetch Rego config: %v", err)
+	} else {
+		defer resp.Body.Close()
+
+		body, err := ioutil.ReadAll(resp.Body)
```
guicassolato, line 103: We need to specify very careful what external registries must comply with.

OPA's REST API, for example, returns an `application/json` to the Get Policy endpoint: https://www.openpolicyagent.org/docs/latest/rest-api/#get-a-policy. We don't have to support that specific data structure, but it is a reference to consider before we commit to any implementation.

pkg/config/authorization/opa.go
jsmadis: Then I would go for failing the reconciliation. It will useful for both options.

api/v1beta1/service_types.go
```diff
@@ -235,6 +235,11 @@ type Authorization_OPA struct {
 	// The Rego document must include the "allow" condition, set by Authorino to "false" by default (i.e. requests are unauthorized unless changed).
 	// The Rego document must NOT include the "package" declaration in line 1.
 	InlineRego string `json:"inlineRego,omitempty"`
+
+	// External URL that contains Rego document which will be downloaded and used.
+	// The Rego document must include the "allow" condition, set by Authorino to "false" by default (i.e. requests are unauthorized unless changed).
+	// The Rego document must NOT include the "package" declaration in line 1.
+	ExternalRegoUrl string `json:"externalRegoUrl,omitempty"`
```
guicassolato, line 242:
```suggestion
	ExternalRegoUrl string `json:"endpoint,omitempty"`
```

pkg/config/authorization/opa.go
```diff
@@ -81,3 +98,51 @@ func generatePolicyUID(policyName string, policyContent string, nonce int) strin
 	data := []byte(fmt.Sprint(nonce) + policyUIDHashSeparator + policyName + policyUIDHashSeparator + policyContent)
 	return fmt.Sprintf("%x", md5.Sum(data))
 }
+
+type ResponseOpaJson struct {
```
guicassolato, line 102: No need to export these types if they're only used for unmarshalling the response.

```suggestion
type responseOpaJson struct {
```

pkg/config/authorization/opa.go
```diff
@@ -19,24 +24,36 @@ default allow = false
...
+	if rego == "" && url != "" {
+		downloadedRego, err := downloadRegoDataFromUrl(url)
+		if err != nil {
+			opaLog.Error(err, regoDownloadErrorMsg, "secret", policyName)
+			return nil, err
+		}
+		rego = downloadedRego
+	}
+
+	rego = cleanUpRegoDocument(rego)
```
guicassolato, line 45: I like this. We "clean it up" even when it's defined inline. 👍

pkg/config/authorization/opa.go
```diff
@@ -81,3 +98,51 @@ func generatePolicyUID(policyName string, policyContent string, nonce int) strin
...
+func cleanUpRegoDocument(rego string) string {
+	r, _ := regexp.Compile("(\\s)*package.*")
+
+	lines := strings.Split(rego, "\n")
+	for i, line := range lines {
+		if r.MatchString(line) {
+			lines[i] = ""
+		}
+	}
+	return strings.Join(lines, "\n")
```
guicassolato, line 147: Would this works perhaps?

```suggestion
	r, _ := regexp.Compile("(\\s)*package.*[;\\n\\s]+")
	return r.ReplaceAllString(s, "")
```

jsmadis: Good catch, but I would remove the whitespace at the end of the regex, we want to catch the end of the line 
```regex
"(\\s)*package.*[;\\n]+"
```

install/crd/config.authorino.3scale.net_services.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.6.1
+    controller-gen.kubebuilder.io/version: v0.3.0
```
guicassolato, line 7: You might need to update your controller-gen. Upgraded by @eguzki in #143.

jsmadis: I will rebase and force push when we are both happy with the PR (so we don't lose any comments).

Do you mind having a lot of commits from development or should I squash it?

guicassolato: I usually try to keep a clean history. It doesn't have to be linear, but without "broken" commits as much as possible.

api/v1beta1/service_types.go
```diff
@@ -235,6 +235,20 @@ type Authorization_OPA struct {
...
+	// The expected JSON structure is the one offered by the OPA REST API (https://www.openpolicyagent.org/docs/latest/rest-api/#get-a-policy).
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// Reference to a Secret key whose value will be passed by Authorino in the request.
+	// The HTTP service can use the shared secret to authenticate the origin of the request.
+	SharedSecret *SecretKeyReference `json:"sharedSecretRef,omitempty"`
+
+	// Defines where client credentials will be passed in the request to the service.
+	// If omitted, it defaults to client credentials passed in the HTTP Authorization header and the "Bearer" prefix expected prepended to the secret value.
+	Credentials Credentials `json:"credentials,omitempty"`
```
guicassolato, line 251: How about...

```go
type ExternalRegistry struct {
  Endpoint string `json:"endpoint,omitempty"`
  SharedSecret *SecretKeyReference `json:"sharedSecretRef,omitempty"`
  Credentials Credentials `json:"credentials,omitempty"`
}

type Authorization_OPA struct {
  InlineRego string `json:"inlineRego,omitempty"`
  ExternalRegistry *ExternalRegistry `json:"externalRegistry,omitempty"`
}
`
?

(Omitting comments here for simplicity.)

File: api/v1beta1/service_types.go [hunk 11]
```diff
@@ -236,6 +236,12 @@ type Authorization_OPA struct {
 	// The Rego document must NOT include the "package" declaration in line 1.
 	InlineRego string `json:"inlineRego,omitempty"`
 
+	// External registry that specifies endpoint with OPA policy.
+	ExternalRegistry ExternalRegistry `json:"external_registry,omitempty"`
```
guicassolato, line 240:
```suggestion
	ExternalRegistry ExternalRegistry `json:"externalRegistry,omitempty"`
```

controllers/service_controller.go
```diff
@@ -229,23 +229,24 @@ func (r *ServiceReconciler) translateService(ctx context.Context, service *confi
 		case configv1beta1.AuthorizationOPA:
 			policyName := service.GetNamespace() + "/" + service.GetName() + "/" + authorization.Name
 			opa := authorization.OPA
+			extRegistry := opa.ExternalRegistry
```
guicassolato, line 232:
```suggestion
			externalRegistry := opa.ExternalRegistry
```

jsmadis:
@guicassolato I will clean up the commit history when you are satisfied with the PR 
:)

api/v1beta1/service_types.go
```diff
@@ -229,12 +229,32 @@ func (a *Authorization) GetType() string {
...
+}
+
 // Open Policy Agent (OPA) authorization policy.
 type Authorization_OPA struct {
 	// Authorization policy as a Rego language document.
 	// The Rego document must include the "allow" condition, set by Authorino to "false" by default (i.e. requests are unauthorized unless changed).
 	// The Rego document must NOT include the "package" declaration in line 1.
 	InlineRego string `json:"inlineRego,omitempty"`
+
+	// External registry that specifies endpoint with OPA policy.
```
guicassolato, line 261:
```suggestion
	// External registry of OPA policies.
```

api/v1beta1/service_types.go
```diff
@@ -229,12 +229,32 @@ func (a *Authorization) GetType() string {
 	return TypeUnknown
 }
 
+//ExternalRegistry specifies external source of data (i.e. OPA policy registry)
+type ExternalRegistry struct {
+	// Endpoint of the HTTP external registry.
+	// the endpoint contains Rego document which will be downloaded and used.
+	// The endpoint must return plain text with rego document or JSON with following structure: {result:{raw:"rego document"}}
+	// The expected JSON structure is the one offered by the OPA REST API (https://www.openpolicyagent.org/docs/latest/rest-api/#get-a-policy).
```
guicassolato, line 242:
```suggestion
	// Endpoint of the HTTP external registry.
	// The endpoint must respond with either plain/text or application/json content-type.
	// In the latter case, the JSON returned in the body must include a path `result.raw`, where the raw Rego policy will be extracted from. This complies with the specification of the OPA REST API (https://www.openpolicyagent.org/docs/latest/rest-api/#get-a-policy).
```

api/v1beta1/service_types.go
```diff
@@ -234,12 +234,31 @@ func (a *Authorization) GetType() string {
 	return TypeUnknown
 }
 
+//ExternalRegistry specifies external source of data (i.e. OPA policy registry)
+type ExternalRegistry struct {
+	// Endpoint of the HTTP external registry.
+	// The endpoint must respond with either plain/text or application/json content-type.
```
guicassolato, line 240: Just for the record, we say responses in `plain/text` or `application/json`, but only the latter is checked. This is fine and can be improved in the future if a third content-type (returned with a 200 OK status code) ever becomes problem when assumed to be a plain/text Rego policy.

guicassolato (approved):
Nice job, @jsmadis! Thank you!

api/v1beta1/service_types.go
jsmadis: I agree. Should I add a comment with it or is it enough to leave it as a comment on the PR?

PR #146, opened 2021-08-25
Generic HTTP metadata can be downloaded without any authentication

jsmadis:
This PR removes the necessity of authentication to the external metadata server. It enables to define CRD with following metadata:
```yaml
metadata:
  - name: echo-api
    http:
      endpoint: http://example.com/metadata
```

### DIFF
--- a/pkg/common/auth_credentials/auth_credentials_test.go
+++ b/pkg/common/auth_credentials/auth_credentials_test.go
@@ -1,6 +1,7 @@
 package auth_credentials
 
 import (
+	"context"
 	"testing"
 
 	envoyServiceAuthV3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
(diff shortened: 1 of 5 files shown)


pkg/common/auth_credentials/auth_credentials_test.go
```diff
@@ -182,3 +183,20 @@ func TestGetCredentialsFromQueryFail(t *testing.T) {
 
 	assert.Error(t, err, "credential not found")
 }
+
+func TestBuildRequestWithCredentials(t *testing.T) {
+	creds := NewAuthCredential("", "")
+	req, err := creds.BuildRequestWithCredentials(context.TODO(), "http://example.com", "GET", "123", nil)
+
+	assert.Check(t, err == nil)
+	assert.Check(t, len(req.Header.Values("Authorization")) == 1)
```
guicassolato, line 192:
```suggestion
	assert.Equal(t, len(req.Header.Values("Authorization")), 1)
```

guicassolato (approved):
Nice job! Thanks!

PR #147, opened 2021-08-30
Rename `Service` CRD to `AuthConfig`

### LINKED ISSUE #108: Rename Authorino `Service` CRD
Authorino's `Service` Custom Resource Definition does not refer to a "service" _per se_, neither to the service eventually being protected, even less to the Authorino external authorization service itself. Rather, it declares the state of authentication and authorization policies used to protect a host. Moreover, named as such, it often causes confusion with Kubernetes native `Service` resource kind. 

Therefore Authorino's `Service` CRD would better be renamed to something else such as “APIProtection” or “AuthPipeline” or “AuthConfig”.

This might also be an opportunity to review group name `config.authorino.3scale.net`.

guicassolato:
`services.configs.authorino.3scale.net/v1beta1` becomes `authconfigs.authorino.3scale.net/v1beta1`.

Closes #108

----

Dear reviewers, 

This is the only change you need to know, really:

Screenshot 2021-08-30 at 17 42 14


gsaslis:
@guicassolato +1 on the renaming. 

(even though `AuthStuff` sounds much more fun 😜 )

api/v1beta1/service_types.go
```diff
@@ -49,11 +49,11 @@ type SecretKeyReference struct {
 	Key string `json:"key"`
 }
 
-// Specifies the desired state of the Service resource, i.e. the authencation/authorization scheme to be applied to protect the matching HTTP services.
```
didierofrivia, line 52: It's good that "HTTP services" was removed, it was a bit limited and wasn't representing the full spectrum of where Authorino is applied 👍🏼

api/v1beta1/service_types.go
```diff
@@ -359,23 +359,23 @@ type ServiceStatus struct {
 // +kubebuilder:printcolumn:name="Authz policies",type=integer,JSONPath=`.status.numAuthorizationPolicies`,description="Number of authorization policies",priority=2
 // +kubebuilder:printcolumn:name="Response items",type=integer,JSONPath=`.status.numResponseItems`,description="Number of items added to the client response",priority=2
 // +kubebuilder:printcolumn:name="Wristband",type=boolean,JSONPath=`.status.festivalWristbandEnabled`,description="Whether issuing Festival Wristbands",priority=2
-type Service struct {
+type AuthConfig struct {
```
didierofrivia, line 362: <3

controllers/auth_config_controller.go
```diff
@@ -42,25 +42,25 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// ServiceReconciler reconciles a Service object
-type ServiceReconciler struct {
+// AuthConfigReconciler reconciles a AuthConfig object
```
didierofrivia, line 45:
```suggestion
// AuthConfigReconciler reconciles an AuthConfig object
```

controllers/auth_config_controller.go
```diff
@@ -42,25 +42,25 @@ import (
...
 
-	service := configv1beta1.Service{}
-	err := r.Get(ctx, req.NamespacedName, &service)
+	authConfig := configv1beta1.AuthConfig{}
+	err := r.Get(ctx, req.NamespacedName, &authConfig)
 	if err != nil && errors.IsNotFound(err) {
 
 		// As we can't get the object, that means it was deleted.
-		// Delete all the services related to this k8s object.
+		// Delete all the auth configs related to this k8s object.
```
didierofrivia, line 63: I'm not sure about this, because if we are talking about "authorization configs" it's separated, but when talking about the CRD, it'd be `authconfigs`, no?. Not suggesting a change, just pointing it out, see if it makes more sense or not :)

didierofrivia: This one applies for any appearance of the term

didierofrivia (commented):
Really minor things, the refactor is sound and very welcomed <3

controllers/auth_config_controller.go
guicassolato: It does make sense, I guess. Thanks!

didierofrivia (approved):
❤️

PR #150, opened 2021-09-14
Makefile: pin controller-gen and kustomize tool versions

eguzki:
Enforce repeatable builds by pinning `kustomize` and `controller-gen` tools to specific version.

The tools will be installed and used from `${PROJECT_DIR}/bin` directory. Note that if version is updated, then installed binary should be manually removed. Then the make target will install the new version.

Signed-off-by: Eguzki Astiz Lezaun <eastizle@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+
 # Image URL to use all building/pushing image targets
 DEFAULT_AUTHORINO_IMAGE = quay.io/3scale/authorino:latest
 AUTHORINO_IMAGE ?= $(DEFAULT_AUTHORINO_IMAGE)
@@ -37,11 +39,11 @@ run: generate fmt vet manifests
 	go run ./main.go
 
 # Install CRDs into a cluster
-install: manifests kustomize
+install: manifests $(KUSTOMIZE)
 	$(KUSTOMIZE) build install | kubectl apply -f -
 
 # Uninstall CRDs from a cluster
-uninstall: manifests kustomize
+uninstall: manifests $(KUSTOMIZE)
 	$(KUSTOMIZE) build install | kubectl delete -f -
 
 # Requests TLS certificates for services if cert-manager.io is installed, the secret is not already present and TLS is enabled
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,11 +39,11 @@ run: generate fmt vet manifests
 	go run ./main.go
 
 # Install CRDs into a cluster
-install: manifests kustomize
+install: manifests $(KUSTOMIZE)
```
guicassolato, line 42: “kustomize” here is the name of a target. No need to change this.

Makefile
```diff
@@ -65,8 +67,37 @@ else
 	echo "tls disabled."
 endif
 
+CONTROLLER_GEN=$(PROJECT_DIR)/bin/controller-gen
+# find or download controller-gen
+# download controller-gen if necessary
+$(CONTROLLER_GEN):
```
guicassolato, line 73: Why do we need a make target with a dynamic name?

eguzki: well, actually it is only the base path set to the current project base directory.

I guess it will also work like this:

```
bin/controller-gen:
     ....
```

But I would use the absolute path, instead of relative just in case current working directory is not the same as the authorino project directory

eguzki: What do you think about this? Makefiles usually assume your CWD is the one which has the makefile, so I guess, it should be safe to do targets like `bin/controller-gen`

eguzki: targets defined as 
```
bin/kustomize:
     ....
bin/controller-gen:
     .....
```

guicassolato (changes requested):
It breaks at https://github.com/Kuadrant/authorino/blob/06c66733af557fc58a813d54bdd532d5119d0f39/Makefile#L55

with:

```
cd deploy/base/certmanager && bin/kustomize edit set namespace authorino
/bin/bash: bin/kustomize: No such file or directory
make[1]: *** [certs] Error 127
make: *** [deploy] Error 2
```

The same will probably happen at https://github.com/Kuadrant/authorino/blob/06c66733af557fc58a813d54bdd532d5119d0f39/Makefile#L71

guicassolato:
> Oh, I see.
> 
> Maybe a clean solution would be to add $[PROJECT_PATH}/bin to the $PATH and then execute `kustomize` instead of `../../../bin/kustomize`. $[PROJECT_PATH}/bin should be the first path in the list to be sure the right version is used.

I was trying this, but then I stumbled into the same as https://stackoverflow.com/questions/8941110/how-i-could-add-dir-to-path-in-makefile#comment110557115_8942216 on macOS. Works on Linux though.

Issue #110, opened 2021-06-14
Authorino External Authorization protocol


guicassolato:
Authorino currently implements [Envoy proxy's external authorization](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter) protocol as the only integration option, requiring clients interested in using Authorino behind other reverse proxies or as a simple external authorization service integrated directly by the client application to comply (i.e. translate the payload in terms of) to the Envoy protocol.

We would like to have a proper "Authorino  external authorization protocol" under which the implementation-specific details and different types of integratation can be abstracted and supported, either being for the reverse-proxy use case or other (direct integration, Kafka brokers, etc).

PR #154, opened 2021-09-29
Kubernetes authorization (SubjectAccessReview)

### LINKED ISSUE #153: Add support for Kubernetes `SubjectAccessReview`-based authorization
Authorino already implements Kubernetes [TokenReview API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/) to support the verification of Kubernetes-valid tokens in the identity ("authentication") phase.

To be able to leverage Kubernetes authorization (RBAC) features, Authorino could implement as well Kubernetes [SubjectAccessReview API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#subjectaccessreview-v1-authorization-k8s-io) in the authorization phase.

Of course, one whose wants to combine Authorino's Kubernetes Authentication and this proposed authorization feature based on `SubjectAccessReview` (and nothing else), would end up with some auth config easily replaceable with [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy), making in such case Envoy+Authorino a bit overkill. On the other hand, other features of Authorino could eventually be added to the stack, e.g. OPA policies, JSON pattern-matching authorization policies, HTTP metadata fetching, wristband tokens, etc.

Suggested in #152

guicassolato:
Adds access control enforcement based on rules defined in the Kubernetes authorization system (e.g. `ClusterRole` and `ClusterRoleBinding` resources of Kubernetes RBAC authorization).

Authorino issues a [SubjectAccessReview](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#subjectaccessreview-v1-authorization-k8s-io) inquiry checking with the underlying Kubernetes cluster whether the user can access the requested resource. It can be used with `resourceAttributes` or `nonResourceAttributes` (the latter inferring HTTP verb and method from the original request).

`User` and `resourceAttributes` of the `SubjectAccessReview` can be specified as a fixed values or as patterns to fetch from the Authorization JSON. A static array of required groups can as well be specified.

`conditions` is also supported and it works exactly as in the JSON  pattern-matching authorization policies.

**Example using Kubernetes authentication and API keys combined with Kubernetes RBAC for authorization:**

```yaml
apiVersion: authorino.3scale.net/v1beta1
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
    - talker-api
  identity:
    - name: service-accounts
      kubernetes:
        audiences:
          - talker-api
      extendedProperties:
        - name: username
          valueFrom:
            authJSON: auth.identity.sub
    - name: api-key-users
      apiKey:
        labelSelectors:
          authorino.3scale.net/managed-by: authorino
      extendedProperties:
        - name: username
          valueFrom:
            authJSON: auth.identity.metadata.annotations.userid
      credentials:
        in: authorization_header
        keySelector: APIKEY
  authorization:
    - name: kubernetes-rbac
      kubernetes:
        conditions: # Comment to enforce this policy for users authenticating with api-key as well
          - selector: auth.identity.iss
            operator: eq
            value: https://kubernetes.default.svc.cluster.local
        user:
          valueFrom:
            authJSON: auth.identity.username
        resourceAttributes: # Comment for non-resource `SubjectAccessReview`
          namespace:
            valueFrom:
              authJSON: context.request.http.path.@extract:{"sep":"/","pos":2}
          resource:
            valueFrom:
              authJSON: context.request.http.path.@extract:{"sep":"/","pos":3}
          name:
            valueFrom:
              authJSON: context.request.http.path.@extract:{"sep":"/","pos":4}
          verb:
            valueFrom:
              authJSON: context.request.http.method.@case:lower|@replace:{"old":"post","new":"create"}|@replace:{"old":"put","new":"update"}
```

In the example, due to the use of `conditions`, `SubjectAccessReview`s will be created only when the client authenticates with a SA token, whereas for API key full access is granted. Remove the `conditions` option to see the policy enforced for all requests regardless of the authentication method.

`resourceAttributes` will cause the `SubjectAccessReview` to check for permissions of the user for accessing the given kind of resource in the specified namespace, with both extracted from the requested URL.

The following custom [gjson modifiers](https://github.com/tidwall/gjson/blob/master/SYNTAX.md#custom-modifiers) are being added to support string transformations:
- `extract:{sep,pos}`
- `replace:{old,new}`
- `case:case`

Check out the full example file included for more details about the exact `Role`s and `RoleBinding`s guaranteeing RBAC for every case, including for SA authentication, for users authenticating with API key, for resource request and for non-resource requests.

A `POST` request to `/v2/target-ns/configmaps` by a SA `system:serviceaccount:requestor-ns:api-consumer-1`, with `resourceAttributes` as part of the config, will result in the following `SubjectAccessReview` resource created by Authorino:

```yaml
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
spec:
  user: system:serviceaccount:requestor-ns:api-consumer-1
  resourceAttributes:
    namespace: target-ns
    resource: configmaps
    verb: create
```

A `GET` request to `/hello` by an API key owned by a user `john`, without `conditions` and without the `resourceAttributes` in the config, will result in the following `SubjectAccessReview` resource created by Authorino:

```yaml
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
spec:
  user: john
  nonResourceAttributes:
    path: "/hello"
    verb: "get"
```

Closes #153

----

**TODOs**

- [x] Extend AuthConfig API for Kubernetes authorization
- [x] Kubernetes authorization evaluator (creates a `SubjectAccessReview` for non-resource requests)
- [x] Extend for `SubjectAccessReview` with `resourceAttributes`
- [x] Add `conditions` option to `authorization.kubernetes`
- [x] Add gjson modifiers to extract and transform string values (`extract`, `replace`, `case`)
- [x] Update docs and examples


GrahamDumpleton:
Is this restricted to `SubjectAccessReview` using `nonResourceAttributes` only?

I was sort of thinking for what I wanted to do that I would need `resourceAttributes`. Ie., similar to `kube-rbac-proxy`.

https://github.com/brancz/kube-rbac-proxy/blob/master/examples/resource-attributes/deployment.yaml#L53

```
    authorization:
      resourceAttributes:
        namespace: default
        apiVersion: v1
        resource: services
        subresource: proxy
        name: kube-rbac-proxy
```

So for the service account for the request I will first want to be able to control in which namespace context the check is perform against, and then possibly should be done based on resources, and not a non resource URL.

Non resource URL seem to be global and not something that could be tied to a namespace, plus very easy to clash with other things using it since can't tie it to secondary context such as a resource. I could be wrong about all this as still theorising how could use it.

guicassolato:
> Is this restricted to `SubjectAccessReview` using `nonResourceAttributes` only?

It is. At least for this first round, `nonResourceAttributes` is what I could reason to make more sense for the API management use case. We can clearly map any requested endpoint of the managed API (HTTP verb + URL info, supplied contextually by Envoy) to the `SubjectAccessReview` resource to create, without requiring any further configuration (i.e. by inferring).

> I was sort of thinking for what I wanted to do that I would need `resourceAttributes`. Ie., similar to `kube-rbac-proxy`.

I can consider extending for supporting `resourceAttributes`, even in this first round. I think it's fair and definitely a valid addition. I'm just not sure how I feel about completely loosing this link between the requested API resource in one hand and the exact `SubjectAccessReview` resource attributes created on the other.

For example (something completely random here), does it make sense that Envoy receives a `PUT /imagestreams/my-is-1` request and Authorino creates a `SubjectAccessReview` checking the permissions for `{ "resourceAttributes": { "resource": "imagestreams", "verb": "get", ⃨ } }`? And then it's a request to `DELETE /whatever/123` and again the `SubjectAccessReview` is to `{ "resourceAttributes": { "resource": "imagestreams", "verb": "get", ⃨ } }`?

To provide more context... One of the things comprised in #110 is the decoupling of Authorino from Envoy's external auth protocol, effectively turning Authorino into a general-purpose authn/authz enforcer, capable of supporting other types of integration where, for instance, there's no API behind where to redirect the traffic to afterwards. In scenarios like these, queries need to be less implicit. We cannot infer things from attributes of an HTTP requests where there might not even be an HTTP request at all! (Think of Authorino protecting a Kafka broker for example.)

So I definitely see a case for `resourceAttributes`. We just need a path from what Authorino is now to that. I wonder... Should we allow the properties of `resourceAttributes` to be set from the `authJSON`?

> So for the service account for the request I will first want to be able to control in which namespace context the check is perform against, and then possibly should be done based on resources, and not a non resource URL.

This is another point I worry about. The safer thing to do would be tying the namespace of the `SubjectAccessReview` to the namespace of the `AuthConfig`, so Authorino can be running in any namespace (in `cluster-wide` deployment mode), but the a user of Authorino (i.e. one who defines an `AuthConfig`) should not be able to sneak into resources from another namespace, where the Kubernetes user who created the `AuthConfig` might not have any privilege.

We more or less enforced this same policy to ensure host names already taken by an `AuthConfig` in one namespace do not collide with other ones from other namespaces, as a way to prevent that users can mess up with API protections of each other in Authorino cluster-wide deployments.

On the other hand, I do understand that non-crossing namespaces restriction closes a few possibilities. What do you think? Should any `AuthConfig` in any namespace be able to sneak into the permissions of any resource of any other namespace across the cluster? Maybe it's okay for the `SubjectAccessReview` in a way it's not for the host name?

> Non resource URL seem to be global and not something that could be tied to a namespace, plus very easy to clash with other things using it since can't tie it to secondary context such as a resource. I could be wrong about all this as still theorising how could use it.

I think you are right about your assumption and I don't like that either in the current implementation. In a way, it kinda partially invalidates my previous point of preventing clashes between users from different namespaces. Though in the case of `nonResourceAttributes` this is given. Whoever uses this feature of Kubernetes should know it's a cluster-wide thing.

> Expanding on the use case. Imagine if have multiple image registry deployments in different namespaces, each fronted by envoy/authorino. I need to be able to control access to these independently. I can't see how that could be done with `nonResourceAttributes` unless you start encoding information about namespace, service type, registry instance etc as part of the special `path` attribute. eg., path of `/namespacename/applicationtype/applicationname/verb`. Is that how expect it would need to be done?

If you really have multiple envoy/authorino fronting each image registry in each namespace, then you can use `namespaced` Authorino deployments instead of `cluster-wide` mode, and my suggestion of tying the namespace in `resourceAttributes`  to the namespace of the `AuthConfig` would work.

Anyway, let know what you think about all this. Rather than one single use case, we need to think what makes sense in general.

----

**TL;DR**

1. Not closed to extending to support `resourceAttributes`.
2. Should we tie the values of `resourceAttributes` (or parts of it) somehow to the original request?
3. Should the user be able to define completely random stuff in the `resourceAttributes`?
4. Should we enforce same-namespace policy to prevent `SubjectAccessReview` based on attributes "owned" by other users in completely unrelated namespaces?

GrahamDumpleton:
> For example (something completely random here), does it make sense that Envoy receives a PUT /imagestreams/my-is-1 request and Authorino creates a SubjectAccessReview checking the permissions for { "resourceAttributes": { "resource": "imagestreams", "verb": "get", ⃨ } }? And then it's a request to DELETE /whatever/123 and again the SubjectAccessReview is to { "resourceAttributes": { "resource": "imagestreams", "verb": "get", ⃨ } }?

Not for image registry case trying to apply it to.

Examples for it are:

```
GET /v2/<targetnamespacename>/...
```

Means want a `SubjectAccessReview` of:

```
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
spec:
  user: system:serviceaccount:<requestornamespacename>:<requestorserviceaccountname>
  resourceAttributes:
      namespace: <targetnamespacename>
      group: registry.example.com
      resource: registry
      name: myregistry
      subresource: images
      verb: get
```

If it was:

```
POST /v2/<targetnamespacename>/...
```

Means want a `SubjectAccessReview` of:

```
apiVersion: authorization.k8s.io/v1
kind: SubjectAccessReview
spec:
  user: system:serviceaccount:<requestornamespacename>:<requestorserviceaccountname>
  resourceAttributes:
      namespace: <targetnamespacename>
      group: registry.example.com
      resource: registry
      name: myregistry
      subresource: images
      verb: create
```

The `<requestornamespacename>` need not be same namespace as `<targetnamespacename>`. As role grants would allow one to pull images from the image registry area related to another namespace (as OpenShift registry allows). So by default can only access own namespaces images, but access to others can be granted if required.

The `<targetnamespacename>` needs to be able to be extracted from the URL path somehow and you can't assume what part it should be as depends on use case. This may though be too complicated since don't have Roga language like OPA to do any manipulation and not sure there is another way of pulling just a segment out of the path. Hadn't though about difficulty of that part before. :-(

The `myregistry` is hardwired in config and relates to the host of that specific `AuthConfig`.

As much as it would have been nice to try and use authorino, had already been thinking whether better to write a custom xDS authz server which implements the logic required. Not knowing golang that would be a lot of work for me. Easier if can work out how to write an xDS authz server endpoint in Python as writing all the kube API stuff in Python is no issue for me.

So maybe I am hoping for too much from authorino and so you can put on hold working further on this until becomes more obvious what to do.

guicassolato:
I'll put some thought on it today, @GrahamDumpleton. The idea of Authorino as a possible replacement for kube-rbac-proxy for more complex use cases is interesting.

Maybe there's something we can do using https://github.com/google/cel-spec, which was in the radar already.

GrahamDumpleton:
Another nice embeddable configuration language is Starlark.

* https://github.com/bazelbuild/starlark/blob/master/spec.md

But then I am partial to it because it is a dialect of Python and so fits my brain. :-)

pkg/config/authorization/kubernetes_authz.go
```diff
@@ -0,0 +1,130 @@
...
+		} else if !match { // skip the policy if any of the conditions does not match
+			return true, nil
+		}
+	}
+
+	jsonValueToStr := func(value common.JSONValue) string {
+		return fmt.Sprintf("%s", value.ResolveFor(dataStr))
+	}
+
+	subjectAccessReview := kubeAuthz.SubjectAccessReview{
```
eguzki, line 82: It seems that the `SubjectAccessReview` does not need `metadata.name`, does it?

eguzki (approved):
Looking good

I cannot detect any issue :)

consider renaming `api/v1beta1/service_types.go` to `authconfig_types.go`

pkg/config/authorization/kubernetes_authz.go
guicassolato: It does not. `SubjectAccessReview`s are transient K8s resources. They pretty much work like queries whose `Status` subresource is provided synchronously in the response of the `create` action.

eguzki: good to know, thanks

PR #159, opened 2021-10-05
Support for custom denial status code/messages

guicassolato:
Adds `spec.denyWith` to the API for custom denial status codes and messages to replace `401 Unauthorized` and `403 Forbidden` responses.

**Example of `AuthConfig` customizing unauthenticated and unauthorized status responses:**

```yaml
spec:
  denyWith:
    unauthenticated:
      code: 302
      message: Login required
      headers:
        - name: Location
          value: http://echo-api.3scale.net/login
    unauthorized:
      code: 302
      headers:
        - name: Location
          value: http://echo-api.3scale.net/not-found
        - name: X-Requested-Path
          valueFrom:
            authJSON: context.request.http.path
```

Deploy the example:

```sh
kubectl -n authorino apply -f ./examples/deny-with.yaml
# authconfig.authorino.3scale.net/talker-api-protection created
# secret/friend-1-api-key-1 created
```

Send a request missing the API key to authenticate:

```sh
curl -H 'Host: talker-api' -H "Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx" http://localhost:8000/hello -i
# HTTP/1.1 302 Found
# location: http://echo-api.3scale.net/login
# x-ext-auth-reason: Login required
# date: Tue, 05 Oct 2021 22:45:45 GMT
# server: envoy
# content-length: 0
```

Without `spec.denyWith.unauthenticated`, the response status code would be `401 Unauthorized`, the `x-ext-auth-reason` header _"credential not found"_, and the `WWW-Authenticate` challenge headers would be present. Instead, status `302 Found` and the `Location` response header are returned.

Send a request forcing an authorization failure:

```sh
$ curl -H 'Host: talker-api' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' -H 'X-Mock-Unauthorized: 1' # http://localhost:8000/hello -i
# HTTP/1.1 302 Found
# location: http://echo-api.3scale.net/not-found
# x-requested-path: /hello
# x-ext-auth-reason: Unauthorized
# date: Tue, 05 Oct 2021 22:46:16 GMT
# server: envoy
# content-length: 0
```

Without `spec.denyWith.unauthorized`, the response status code would be `403 Forbidden`. Instead, status `302 Found` and the `Location` response header are returned.

Notice as well the presence of a custom HTTP header `x-requested-path`, with value dynamically set to the path of the original request.

PR #165, opened 2021-10-07
`valueFrom.authJSON` with mixed static string and variable placeholders

guicassolato:
Modifies `valueFrom.authJSON` so values can be
a) any [gjson](https://pkg.go.dev/github.com/tidwall/gjson) pattern (as today), with the pattern representing a path to be fetched from the authorization JSON (e.g. `context.request.http.host`)
b) a string template with variable placeholders that resolve to patterns (e.g. `Hello, {auth.identity.name}!`)

Patterns (whether or not inside of variable placeholders) keep supporting the following custom modifiers as introduced in #154:
- `@extract:{sep:" ",pos:0}`
- `@replace{old:"",new:""}`
- `@case:upper|lower`

Examples:

```yaml
- name: ts
  valueFrom:
    authJSON: context.request.time.seconds
- name: username
  valueFrom:
    authJSON: auth.identity.metadata.annotations.authorino\.3scale\.net/username|@case:lower
- name: geeting-message
   valueFrom:
     authJSON: Hello, {auth.identity.metadata.annotations.authorino\.3scale\.net/name}!
```

Inspired by https://github.com/Kuadrant/authorino/pull/159#issuecomment-936601468

### DIFF
--- a/pkg/common/json.go
+++ b/pkg/common/json.go
@@ -22,6 +22,11 @@ const (
 	unsupportedOperatorErrorMsg = "Unsupported operator for JSON authorization"
 )
 
+var (
+	allCurlyBracesRegex          = regexp.MustCompile("{")
+	curlyBracesForModifiersRegex = regexp.MustCompile(`[^@]+@\w+:{`)
+)
+
 // JSONProperty represents a name-value pair for a JSON property where the value can be a static value or
 // a pattern for a value fetched dynamically from the authorization JSON
 type JSONProperty struct {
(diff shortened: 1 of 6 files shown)


pkg/common/json_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func TestJSONValueResolveFor(t *testing.T) {
-	jsonData := `{
+	const jsonData = `{
```
eguzki, line 13: why? jsonData is always passed as value, hence copy is done.

pkg/common/json.go
```diff
@@ -123,14 +127,58 @@ func UnmashalJSONResponse(resp *http.Response, v interface{}, b *[]byte) error {
...
-	regex := regexp.MustCompile("{([^}]*)}")
-	matches := regex.FindAllStringSubmatch(source, -1)
-	for _, selector := range matches {
-		value := gjson.Get(jsonData, selector[1]).String()
-		replaced = strings.ReplaceAll(replaced, "{"+selector[1]+"}", value)
+	var replaced, buffer []byte
+	var escaping, insidePlaceholder bool
+	var nestedCurlyBraces int
+
+	for _, b := range []byte(source) {
```
eguzki, line 134: Too bad there is no library to use to parse closing matching curly braces and support escaping them. Using regexp would be impossible?

pkg/common/json_test.go
```diff
@@ -58,6 +58,70 @@ func TestJSONValueResolveFor(t *testing.T) {
 	var resolvedRoles []string
 	_ = json.Unmarshal(resolvedValueAsJSON, &resolvedRoles)
 	assert.DeepEqual(t, resolvedRoles, []string{"user", "admin"})
+
+	// Message with variable placeholder
+	value = JSONValue{Pattern: "Hello, {auth.identity.username}!"}
+	assert.Equal(t, value.ResolveFor(jsonData), "Hello, john!")
+}
+
+func TestReplaceJSONPlaceholders(t *testing.T) {
```
eguzki, line 67: Maybe add some tests for "invalid" placeholder specifications? 

Maybe check the CR for invalid fields and reject the creation/update?

eguzki (commented):
Looks good.

Left some comments

PR #186, opened 2021-11-10
Custom headers option for HTTP external metadata requests

### LINKED ISSUE #183: User-defined headers to HTTP external metadata request
Authorino does not allow users to freely customize the headers of the HTTP request sent by the [generic HTTP external metadata](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#external-http-authorization-metadata) feature. Only `Content-Type` and `Authorization` headers can be configured; even so, the former is limited to a couple values used to format the payload, and the latter can be set exclusively from a `sharedSecretRef` option.

An extension of the API changing that could possibly support:

```yaml
metadata:
  - name: ams-current-account
    http:
      endpoint: https://metadata-source.svc.cluster.local
      headers:
        - name: X-Custom-Header
          value: some-value
        - name: Authorization
          valueFrom:
            authJSON: context.request.http.headers.authorization # carries the same authentication provided in the original request to the protected service, to the request to the external metadata
      credentials:
        in: authorization_header
        keySelector: Bearer
      sharedSecretRef:
        name:  my-secret
        key: some-key
```

**Questions and thoughts based on the example above:**
1. Does `metadata.http.credentials.in='authorization_header'` override the custom header or add a second `Authorization` header?
    _A:_ My guess is that it would override the custom header.

2. If `metadata.http.credentials.in` is set to anything other than `authorization_header`, would it be virtually like providing 2 authentication values?
    _A:_ I guess so.

3. If `metadata.http.sharedSecretRef` had been omitted, then `metadata.http.credentials` would have no meaning. Would a user find this weird because a custom `Authorization` header has been set but `credentials` below changes nothing?
    _A:_ My guess is that this is not a problem, since it is how `credentials` and `sharedSecretRef` already work together today and custom headers would be a new thing, not necessarily meant only to customize the `Authorization` header.

4. Would users expect to be able to custom set the `Authorization` header from a secret but by adding it to `metadata.http.headers`?
    _A:_ Maybe yes, but there is already a way to set `Authorization` header and even custom headers from a secret for the purpose of authentication. Users should use that in this case.

5. Would users expect to set authentication `valueFrom.authJSON` into a credential vector other than the `Authorization` header (e.g. as a cookie).
    _A:_ Out of scope as this is about custom headers, not about enhancing authentication config.

guicassolato:
Extends the `AuthConfig` API with option to customize the headers of the HTTP request to external sources of metadata (Generic HTTP external metadata feature).

E.g.

```yaml
spec:
  metadata:
    - name: external-metadata
      http:
        endpoint: http://external-source.io/metadata
        headers:
          - name: X-Requested-By
            value: authorino
          - name: Authorization # overwritten if used in combination with `sharedSecretRef` and `credentials.in=authorization_header`
            valueFrom:
              authJSON: context.request.http.headers.authorization # reuses the same authn token to authenticate with the external metadata service
```

Closes #183

### DIFF
--- a/controllers/auth_config_controller.go
+++ b/controllers/auth_config_controller.go
@@ -234,10 +234,22 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
 				})
 			}
 
+			headers := make([]common.JSONProperty, 0, len(genericHttp.Headers))
+			for _, header := range genericHttp.Headers {
+				headers = append(headers, common.JSONProperty{
+					Name: header.Name,
+					Value: common.JSONValue{
+						Static:  header.Value,
+						Pattern: header.ValueFrom.AuthJSON,
+					},
+				})
+			}
+
 			translatedMetadata.GenericHTTP = &authorinoMetadata.GenericHttp{
 				Endpoint:        genericHttp.Endpoint,
 				Method:          string(genericHttp.Method),
 				Parameters:      params,
+				Headers:         headers,
 				ContentType:     string(genericHttp.ContentType),
 				SharedSecret:    sharedSecret,
 				AuthCredentials: auth_credentials.NewAuthCredential(creds.KeySelector, string(creds.In)),
--- a/install/crd/authorino.3scale.net_authconfigs.yaml
+++ b/install/crd/authorino.3scale.net_authconfigs.yaml
@@ -782,6 +782,37 @@ spec:
                             where "selector" is any pattern supported by https://pkg.go.dev/github.com/tidwall/gjson
                             and selects value from the authorization JSON. E.g. https://ext-auth-server.io/metadata?p={context.request.http.path}
                           type: string
+                        headers:
+                          description: Custom headers in the HTTP request.
+                          items:
+                            properties:
+                              name:
+                                description: The name of the claim
+                                type: string
+                              value:
+                                description: Static value of the claim
+                                x-kubernetes-preserve-unknown-fields: true
+                              valueFrom:
+                                description: Dynamic value of the claim
+                                properties:
+                                  authJSON:
+                                    description: 'Selector to fill the value from
+                                      the authorization JSON. Any patterns supported
+                                      by https://pkg.go.dev/github.com/tidwall/gjson
+                                      can be used. The value can be just the pattern
+                                      with the path to fetch from the authorization
+                                      JSON (e.g. ''context.request.http.host'') or
+                                      a string template with variable placeholders
+                                      that resolve to patterns (e.g. "Hello, {auth.identity.name}!")
+                                      The following string modifiers are available:
+                                      @extract:{sep:" ",pos:0}, @replace{old:"",new:""},
+                                      @case:upper|lower, and @base64:encode|decode.'
+                                    type: string
+                                type: object
+                            required:
+                            - name
+                            type: object
+                          type: array
                         method:
                           description: 'HTTP verb used in the request to the service.
                             Accepted values: GET (default), POST. When the request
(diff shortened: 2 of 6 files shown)


install/crd/authorino.3scale.net_authconfigs.yaml
```diff
@@ -782,6 +782,37 @@ spec:
                             where "selector" is any pattern supported by https://pkg.go.dev/github.com/tidwall/gjson
                             and selects value from the authorization JSON. E.g. https://ext-auth-server.io/metadata?p={context.request.http.path}
                           type: string
+                        headers:
+                          description: Custom headers of the HTTP request.
```
thomasmaas, line 786:
```suggestion
                          description: Custom headers in the HTTP request.
```

controllers/auth_config_controller.go
```diff
@@ -234,10 +234,22 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
 				})
 			}
 
+			headers := make([]common.JSONProperty, 0, len(genericHttp.Headers))
+			for _, header := range genericHttp.Headers {
+				headers = append(headers, common.JSONProperty{
```
eguzki, line 239: Having `JsonProperty` and `JSONProperty` is a little bit confusing. Small thing.. just for the record.

eguzki (dismissed):

jjaferson:
Code looks good to me 

Metadata added to the authconfig
```  
metadata:
  - http:
      contentType: application/x-www-form-urlencoded
      endpoint: http://talker-api.authorino.svc.cluster.local:3000/metadata
      headers:
      - name: X-Requested-By
        value: authorino
      - name: Authorization
        valueFrom:
          authJSON: context.request.http.headers.authorization
      method: GET
    name: external-metadata
```

Response from the request
```❯ curl -H 'Host: talker-api' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' http://localhost:8000/metadata
{
  "method": "GET",
  "path": "/metadata",
  "query_string": null,
  "body": "",
  "headers": {
    "Host": "talker-api",
    "User-Agent": "curl/7.64.1",
    "Accept": "*/*",
    "Authorization": "APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx",
    "X-Forwarded-For": "10.244.0.9",
    "X-Forwarded-Proto": "http",
    "X-Envoy-Internal": "true",
    "X-Request-Id": "bf4ce407-23c3-4fb9-8d51-129899657d31",
    "Echo-Api-Metadata": "{\"external-metadata\":{\"body\":\"\",\"headers\":{\"Accept-Encoding\":\"gzip\",\"Authorization\":\"APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx\",\"Content-Type\":\"text/plain\",\"Host\":\"talker-api.authorino.svc.cluster.local:3000\",\"User-Agent\":\"Go-http-client/1.1\",\"Version\":\"HTTP/1.1\",\"X-Requested-By\":\"{\\\"authorino\\\" \\u003cnil\\u003e}\"},\"method\":\"GET\",\"path\":\"/metadata\",\"query_string\":null,\"uuid\":\"dc89674e-9046-4d33-a461-b92ad6ec4d92\"},\"get-request\":{\"body\":\"\",\"headers\":{\"Accept-Encoding\":\"gzip\",\"Authorization\":\"Bearer 7f6fcf4961b979cd703e6c396d1378e282b109991344a97c2ad4d785eece3d54\",\"Content-Type\":\"text/plain\",\"Host\":\"talker-api.authorino.svc.cluster.local:3000\",\"User-Agent\":\"Go-http-client/1.1\",\"Version\":\"HTTP/1.1\"},\"method\":\"GET\",\"path\":\"/metadata\",\"query_string\":\"encoding=text/plain\\u0026original_path=/metadata\",\"uuid\":\"19f2c474-cdb2-4fe4-8835-a54ac875e812\"},\"post-request-form\":null,\"post-request-json\":null}",
    "X-Envoy-Expected-Rq-Timeout-Ms": "15000",
    "Version": "HTTP/1.1"
  },
  "uuid": "00a68c34-e230-4752-9f72-8abe4cf0d9b8"
}%```

PR #198, opened 2021-12-09
Fix DNS names of the cert-manager certificates

guicassolato:
Fix Authorino service names in the DNS names of the cert-manager certificates, according to how they are created by the Authorino Operator.


guicassolato:
I will merge this so the examples that rely on TLS enabled do not break and because soon we will be moving the logic for creating the certs to https://github.com/3scale/authorino-operator.

Meanwhile, I am not entirely happy with how the TLS certs are being defined, but at least they will work OTB with all the examples and dev workflow instructions we have linked in the docs.

PR #206, opened 2022-01-03
Conditions

### LINKED ISSUE #24: Per-request authN mode detection
Authorino currently does not provide support for per-request switching of the authentication mode, as requested by the client in the Authentication type portion of the `Authorization` request header. Right now, with only OIDC authentication mode implemented, Authorino expects the Authentication type to be always "Bearer". While implementing other authentication modes, it is expected that other types will eventually apply (e.g. Basic). This, however, does not mean necessarily that the client will get to chose the authentication mode. If only one authN mode is allowed per API config, it wouldn't be possible anyway.

This proposal is for the per-request AuthN mode switch to be done from interpreting the Authentication type provided by the client in each request (as long as more than one mode is enabled in the API config). Instead of always "Bearer", whenever an API was configured for multiple alternative authentication methods, a client would be able to pass in the Authorization request header Authentication types such as OIDC, APIKEY, MTLS or MAC, respectively activating each corresponding identity verification flow.

Any reasonable implementation here would try to stick with conventions adopted by the industry, and perhaps with optional configurable settings for the feature.

### LINKED ISSUE #205: Conditions
Authorino's [JSON pattern-matching authz](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#json-pattern-matching-authorization-rules-authorizationjson) and [K8s SubjectAccessReview-based authz](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#kubernetes-subjectaccessreview-authorizationkubernetes) both define a field called `conditions` that allows to specify rules for Authorino to decide when to apply or to skip that particular policy. The 3rd type of authz supported by Authorino, which is [OPA authz](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#open-policy-agent-opa-rego-policies-authorizationopa), does not define `conditions` because one could easily model the same logic in Rego.

Perhaps `conditions` could be brought to all evaluator configs of all phases (identity, metadata, authorization and dynamic response), and maybe even to top level of the `AuthConfig` as a whole.

After the lookup – which shall continue to be based on the host at first –, the config as a whole as well as parts of it can be decided to be either enforced or skipped based on those multiple `conditions` fields, specified at the top level and/or at the level of each evaluator.

With `conditions` defined at the top level of the `AuthConfig` (right after the lookup), maybe each condition could be given a unique name. In the other `conditions` fields defined at each specific evaluator (e.g. the one for a particular authn method to be evaluated, or for a particular metadata config), a named top-level condition could be referenced, thus avoiding repetition of rules whenever a same set can be reused in multiple places. There is also a potential for inheritance here.

The good thing of using `conditions` is that it counts with the entire Authorization JSON as input, meaning matching rules could be based on path, on any attribute of the HTTP request or even on attributes of objects appended to the Authorization JSON in other phases (e.g. only fetching some particular metadata when a certain authn method was used). Plus, the cost of implementation of `conditions` as described here is low since the functionality already exists.

Maybe in the future, eventually replacing the lookup based on host strictly, with the “top-level conditions” could be a no-brainer, assuming ways to solve the issues of the lookup complexity and preventing overlapping `AuthConfig` (aka "[avoiding host name collision](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#avoiding-host-name-collision)") can be figured out. This could be a way to address #171 (on steroids!)

----

**Example of `AuthConfig` using conditions:**

```yaml
kind: AuthConfig
metadata:
  name: my-api-protection
spec:
  hosts:
  - my-api.io

  conditions:
    operator: neq
    value: /status

  identity:
  - name: idp-users
    oidc:
      endpoint: https://my-idp/auth/realms/my-realm
  - name: admins
    apiKey:
      labelSelectors:
      credentials:
        in: authorization_header
        keySelector: APIKEY
    conditions:
      operator: matches
      value: /admin(/.*))$
    - selector: context.request.http.headers.authorization
      operator: matches
      value: APIKEY .+
    extendedProperties:
    - name: username
      valueFrom:
        authJSON: auth.identity.metadata.annotations.username
    - name: roles
      value: ["admin"]

  metadata:
  - name: resource-metadata
    http:
      endpoint: http://metadata-source/pets/{context.request.http.path.@extract:{"sep":"/","pos":2}}/meta
      method: GET
    conditions:
      operator: matches
      value: /pets/\d+(/.*))$  

  authorization:
  - name: owner-policy
    opa:
      inlineRego: |
        allow { input.auth.identity.username == input.metadata["resource-metadata"].owner }
    conditions:
      operator: matches
      value: /pets/\d+(/.*))$
  - name: admin-role-required
    json: # now reduced only to what's currently the 'rules' part of the spec, since 'conditions' moved to the upper level
      operator: incl
      value: "admin"
    conditions:
      operator: matches
      value: /admin(/.*))$
```

In the example above:
- Requests to `/status` will not be protected by the `AuthConfig`
- Users trying to access paths under `/admin` can _also_ authenticate using API key
- Conditions for `identity[name="admins"]` and `authorization[name="admin-role-required"]` share at least one rule in common – maybe this rule could be defined at the top-level and referenced.
- Conditions for `metadata[name="resource-metadata"]` and `authorization[name="owner-policy"]` are the exactly same – maybe it could be defined at the top-level and referenced.

guicassolato:
_Conditions_ (new feature), named `when` in the AuthConfig API, are sets of expressions (JSON patterns) that whenever included must evaluate to true against the Authorization JSON, so the scope where the expressions are defined is enforced; otherwise Authorino will skip that scope in the pipeline.

The scope for a set of conditions can be the entire `AuthConfig` (top level) or a particular evaluator of any phase of the auth pipeline.

Each expression is a tuple composed of a `selector`, an `operator` and a fixed comparable `value`, just like the pattern matching expressions used in the existing `conditions` field of [JSON pattern-matching authz](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#json-pattern-matching-authorization-rules-authorizationjson) and [K8s SubjectAccessReview-based authz](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#kubernetes-subjectaccessreview-authorizationkubernetes) – both now superseded by this new implementation.

Literal expressions and references to expression sets (`patterns`,  defined at the top level of the `AuthConfig` spec) can be listed, mixed and combined in the `when` conditions in the `AuthConfig`.

E.g., `when` conditions can be used:

i. to skip an entire `AuthConfig` based on the context:

```yaml
spec:
  when:
    operator: neq
    value: /status
```

ii. to skip parts of an `AuthConfig` (i.e. a specific evaluator):

```yaml
spec:
  metadata:
  - name: metadata-source
    http:
      endpoint: https://my-metadata-source.io
    when:
    - selector: context.request.http.method
      operator: neq
      value: OPTIONS
```

iii. to enforce a particular evaluator only in certain contexts (really the same as the above, though to a different use case):

```yaml
spec:
  identity:
  - name: authn-meth-1
    apiKey: {...} # this authn method only valid for POST requests to /foo[/*]
    when:
      operator: matches
      value: ^/foo(/.*)?$
    - selector: context.request.http.method
      operator: eq
      value: POST

  - name: authn-meth-2
    oidc: {...}
```

iv. to avoid repetition while defining patterns for conditions:

```yaml
spec:
  patterns:
    pets:
      operator: matches
      value: ^/pets/\d+(/.*)$

  metadata:
  - name: pets-info
    when:
    - patternRef: pets
    http:
      endpoint: https://pets-info.io?petId={context.request.http.path.@extract:{"sep":"/","pos":2}}

  authorization:
  - name: pets-owners-only
    when:
    - patternRef: pets
    opa:
      inlineRego: |
        allow { input.metadata["pets-info"].ownerid == input.auth.identity.userid }
```

v. to avoid repetition while defining JSON pattern-matching authorizaton rules:

```yaml
spec:
  patterns:
    admin:
      operator: incl
      value: admin

  authorization:
  - name: secured-endpoints
    when:
      operator: matches
      value: ^/secure(/.*)?$
    json:
      rules:
      - patternRef: admin

  - name: only-admins-can-delete
    when:
    - selector: context.request.http.method
      operator: eq
      value: DELETE
    json:
     rules:
      - patternRef: admin
```

vi. mixing and combining literal expressions and refs:

```yaml
spec:
  patterns:
    foo:
      operator: eq
      value: /foo

  when: # unauthenticated access to /foo always granted
  - patternRef: foo
  - selector: context.request.http.headers.authorization
    operator: eq
    value: ""

  authorization:
  - name: my-policy-1
    when: # authenticated access to /foo controlled by policy
    - patternRef: foo
    json: {...}
```

vii. to avoid evaluating unnecessary identity checks when the user can indicate the preferred authentication method (again the pattern of skipping based upon the context):

```yaml
spec:
  identity:
  - name: jwt
    when:
    - selector: selector: context.request.http.headers.authorization
      operator: matches
      value: JWT .+
    oidc: {...}

  - name: api-key
    when:
    - selector: context.request.http.headers.authorization
      operator: matches
      value: APIKEY .+
    apiKey: {...}
```

### Verification steps

Try this PR locally with an `AuthConfig` that combines most of the use case patterns above into one same spec:

```sh
make local-setup DEPLOY_KEYCLOAK=1

kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:

  patterns: # these are named sets of expressions that can be reused
    a-pet:
      operator: matches
      value: ^/pets/\d+(/.*)?$

  when: # top-level conditions for the entire AuthConfig to be enforced
    operator: neq
    value: /status

  identity:
  - name: idp-users
    oidc:
      endpoint: http://keycloak:8080/auth/realms/kuadrant
    when:
    - selector: context.request.http.headers.authorization
      operator: matches
      value: JWT .+
    credentials:
      in: authorization_header
      keySelector: JWT
    extendedProperties:
    - name: username
      valueFrom:
        authJSON: auth.identity.preferred_username
    - name: roles
      valueFrom:
        authJSON: auth.identity.realm_access.roles

  - name: admins
    apiKey:
      labelSelectors:
    when:
    - selector: context.request.http.headers.authorization
      operator: matches
      value: APIKEY .+
    credentials:
      in: authorization_header
      keySelector: APIKEY
    extendedProperties:
    - name: username
      valueFrom:
        authJSON: auth.identity.metadata.annotations.username
    - name: roles
      value: ["admin"]

  metadata:
  - name: resource-metadata
    http:
      endpoint: http://talker-api:3000/pets/{context.request.http.path.@extract:{"sep":"/","pos":2}}/meta
      method: GET
    when: # skip unnecessary out-of-context metadata fetching
    - patternRef: a-pet

  authorization:
  - name: owner-policy
    opa:
      inlineRego: |
        allow { input.auth.identity.username == input.metadata["resource-metadata"].owner }
    when:
    - patternRef: a-pet
      operator: excl
      value: admin

  - name: admin-role-required
    json:
      rules:
        operator: incl
        value: admin
    when:
      operator: matches
      value: ^/admin(/.*)?$
EOF

apiVersion: v1
kind: Secret
metadata:
  name: api-key-1
  labels:
  annotations:
    username: alice
stringData:
type: Opaque
EOF
```

Try the `/status` endpoint, skipped according to the top-level `when` conditional:

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/status -i
# HTTP/1.1 200 OK
```

Try with missing authentication on a case where the `AuthConfig` is enforced, according to the conditions:

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/pets -i
# HTTP/1.1 401 Unauthorized
```

Try protected endpoints as a user bound to the _member_ role in Keycloak:
 
```sh
ACCESS_TOKEN=$(kubectl -n authorino run token --attach --rm --restart=Never -q --image=curlimages/curl -- http://keycloak:8080/auth/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=john' -d 'password=p' | jq -r .access_token)

curl -H "Authorization: JWT $ACCESS_TOKEN" http://talker-api-authorino.127.0.0.1.nip.io:8000/pets -i
# HTTP/1.1 200 OK

curl -H "Authorization: JWT $ACCESS_TOKEN" http://talker-api-authorino.127.0.0.1.nip.io:8000/pets/123 -i
# HTTP/1.1 403 Forbidden

curl -H "Authorization: JWT $ACCESS_TOKEN" http://talker-api-authorino.127.0.0.1.nip.io:8000/admin -i
# HTTP/1.1 403 Forbidden
```

Try protected endpoints as a user bound to the _admin_ role in Keycloak:

```sh
ACCESS_TOKEN=$(kubectl -n authorino run token --attach --rm --restart=Never -q --image=curlimages/curl -- http://keycloak:8080/auth/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=jane' -d 'password=p' | jq -r .access_token)

curl -H "Authorization: JWT $ACCESS_TOKEN" http://talker-api-authorino.127.0.0.1.nip.io:8000/pets -i
# HTTP/1.1 200 OK

curl -H "Authorization: JWT $ACCESS_TOKEN" http://talker-api-authorino.127.0.0.1.nip.io:8000/pets/123 -i
# HTTP/1.1 200 OK

curl -H "Authorization: JWT $ACCESS_TOKEN" http://talker-api-authorino.127.0.0.1.nip.io:8000/admin -i
# HTTP/1.1 200 OK
```

Try protected endpoints as an API-key user (_admin_ by definition, according to the `AuthConfig`):

```sh
curl -H "Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx" http://talker-api-authorino.127.0.0.1.nip.io:8000/pets -i
# HTTP/1.1 200 OK

curl -H "Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx" http://talker-api-authorino.127.0.0.1.nip.io:8000/pets/123 -i
# HTTP/1.1 200 OK

curl -H "Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx" http://talker-api-authorino.127.0.0.1.nip.io:8000/admin -i
# HTTP/1.1 200 OK
```

Closes #205
Closes #24

----

- [x] Conditions
- [x] Tests
- [x] Docs


eguzki:
First thing in my mind after reading the descr of the PR: `s/conditions/when/` 

Minor thing, but wdyt?

eguzki:
`patterns` should not be `condition-patterns` or something similar? To be more meaning ful at top level

guicassolato:
> First thing in my mind after reading the descr of the PR: `s/conditions/when/`
> 
> Minor thing, but wdyt?

Totally valid suggestion, @eguzki.

I thought about `when` myself. Ended up going with `conditions` for 2 reasons:
1. because it's the same semantics and syntax of the already existing `spec.authorization.json.conditions` and `spec.authorization.kubernetes.conditions`
2. So very "creative" users don't end up thinking there's some cron thing involved here – "when" means "if" in this case

`if` would also make sense.

eguzki:
I cannot find it. What is the semantic meaning as a boolean operation of a list of conditions? AND, OR?

guicassolato:
> I cannot find it. What is the semantic meaning as a boolean operation of a list of conditions? AND, OR?

It's AND.

guicassolato:
> `patterns` should not be `condition-patterns` or something similar? To be more meaningful at top level

The patterns can also be used in `spec.authorization.json.rules`, not only in `conditions` fields.

eguzki:
implementation looks good to me

eguzki:
Influenced by OAS design (sorry I cannot avoid it), I miss some expression in the form of
```
(cond11 AND cond12 AND ... AND cond1N) OR (cond21 AND cond22 AND ... AND cond2N) OR ... OR (condM1 AND condM2 AND ... AND condMN)
```

I wonder if this is even a need. But I can  think two  example:

```
GET /path1 -> AUTH type: apikey OR http basic  (OAS allows this flexibility)
```
This use case is already covered having two different identity evaluators that are evaluated with OR semantics. So maybe does not apply but I think it is worth mentioning.

```
GET /path1 -> AUTH type: apikey
GET /path2 -> AUTH type: apikey 
```
This use case requires defining two different identity evaluators which are exactly the same (except for the conditions)

api/v1beta1/auth_config_types.go
```diff
@@ -175,6 +212,9 @@ type Metadata struct {
 	// +kubebuilder:default:=0
 	Priority int `json:"priority,omitempty"`
 
+	// Conditions that must match for Authorino to enforce this metadata config; otherwise, the config will be skipped.
+	Conditions []JSONPattern `json:"conditions,omitempty"`
```
maleck13, line 216: If there are no conditions set, is the default not to skip?

guicassolato:
> Influenced by OAS design (sorry I cannot avoid it), I miss some expression in the form of
> 
> ```
> (cond11 AND cond12 AND ... AND cond1N) OR (cond21 AND cond22 AND ... AND cond2N) OR ... OR (condM1 AND condM2 AND ... AND condMN)
> ```
> 
> I wonder if this is even a need. But I can think two example:
> 
> ```
> GET /path1 -> AUTH type: apikey OR http basic  (OAS allows this flexibility)
> ```
> 
> This use case is already covered having two different identity evaluators that are evaluated with OR semantics. So maybe does not apply but I think it is worth mentioning.
> 
> ```
> GET /path1 -> AUTH type: apikey
> GET /path2 -> AUTH type: apikey 
> ```
> 
> This use case requires defining two different identity evaluators which are exactly the same (except for the conditions)

No problem, @eguzki.

OR logics can be achieved by repeating the evaluator spec as many times as needed, varying only the `conditions` block.

For authorization based on JSON pattern-matching, because of https://github.com/Kuadrant/authorino/pull/206#issuecomment-1011033183, repeating the rules can get slightly cleaner.

E.g.:

```yaml
spec:
  patterns:
    my-authz-rules:
    - selector: some.json.path
      operator: eq
      value: some-expected-value
    - selector: some.other.json.path
      operation: incl
      value: some-value-that-must-be-present-in-the-selected-array
  authorization:
  - name: use-case-1
    conditions:
      operator: eq
      value: /path1
    json:
      rules:
      - patternRef: my-authz-rules
  - name: use-case-2
    conditions:
      operator: eq
      value: /path2
    json:
      rules:
      - patternRef: my-authz-rules
```

As for the examples you shared...

First one is simpler:

```yaml
spec:
  patterns:
    get-path1:
    - selector: context.request.http.method
      operation: eq
      value: GET
      operator: eq
      value: /path1
  identity:
  - name: api-key 
    conditions:
    - patternRef: get-path1
    apikey: {...}
    credentials:
      in: custom_header
      keySelector: API-KEY
  - name: http-basic-auth
    conditions:
    - patternRef: get-path1
    apikey: {...} # RFC 7235 happens to be mimicked in Authorino using API key authn as well, though the label selectors here would probably differ from the ones in the `apiKey` above
    credentials:
      in: authorization_header
      keySelector: Basic
```

Second one is a bit more annoying because it involves really duplicating the apiKey authn spec:

```yaml
spec:
  patterns:
    get:
    - selector: context.request.http.method
      operation: eq
      value: GET
  identity:
  - name: path1-authn 
    conditions:
    - patternRef: get
      operator: eq
      value: /path1
    apikey: {...}
  - name: path2-authn
    conditions:
    - patternRef: get
      operator: eq
      value: /path2
    apikey: {...} # identical to the `apiKey` above
```

api/v1beta1/auth_config_types.go
guicassolato: Correct. No conditions means always enforce it.

guicassolato: Ouch. Got it now. Description is wrong. On it!

guicassolato: Perhaps add to the description:

```go
// If omitted, the config will be enforced every time.
```

?

docs/user-guides/json-pattern-matching-authorization.md
```diff
@@ -124,19 +124,19 @@ spec:
       endpoint: http://keycloak.keycloak.svc.cluster.local:8080/auth/realms/kuadrant
   authorization:
   - name: email-verified-only
+    conditions:
```
maleck13, line 127: maybe not a good idea but will throw it out there: What if this were named: ```applyWhen``` do you think it would me more intuative? Perhaps not just a thought

maleck13 (commented):
I have given it a first review and it looks pretty good to me. As you say it is a breaking change. I am going to cast my eye over it once more as there is quite a bit to take in.

api/v1beta1/auth_config_types.go
maleck13: Yeah sounds good, it is a minor thing , it just caught my attention as I read it

docs/user-guides/json-pattern-matching-authorization.md
guicassolato: If you say it's an improvement and not subject to the possible misunderstanding I mention in my 2nd point in https://github.com/Kuadrant/authorino/pull/206#issuecomment-1011031494, sounds good to me too

To sum up a few options we've got so far:
- `conditions`
- `when`
- `applyWhen`
- `if`

I personally prefer single-word names whenever possible, but only when they are clear enough.

Let's start a quick poll?

maleck13: Yes as an improvement. Single word is fine also. ```When``` gets my vote

maleck13:
Other than the suggestion to change conditions to when and a broad conversation about whether to focus config in the CR and Policy in rego. This looks good to me 👍

docs/user-guides/json-pattern-matching-authorization.md
guicassolato: Two votes for `when`. Sounds like a crushing majority to me 🙂 I'll make the changes.

Thanks!

jjaferson (approved):
Code looks good and all the tests described in the verification steps passed.

PR #212, opened 2022-01-17
Creates Authorino CR in a tmp folder inside of the project dir

jjaferson:
This PR moves the creation of Authorino CR to a directory inside of the project oo avoid getting `Permission denied` when generating the file via makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,10 +122,13 @@ namespace:
 	kubectl create namespace $(NAMESPACE)
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+EDITOR ?= vim
 deploy: certs
 	@{ \
 	set -e ;\
-	TEMP_FILE=/tmp/authorino-deploy-$$(openssl rand -hex 4).yaml ;\
+	TEMP_DIR=$(PROJECT_DIR)/tmp ;\
+	mkdir -p $$TEMP_DIR ;\
+	TEMP_FILE=$$TEMP_DIR/authorino-deploy.yaml ;\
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -125,12 +125,13 @@ namespace:
 deploy: certs
 	@{ \
 	set -e ;\
-	TEMP_FILE=/tmp/authorino-deploy-$$(openssl rand -hex 4).yaml ;\
+	TEMP_DIR=$(PROJECT_DIR)/tmp ;\
+	mkdir $$TEMP_DIR ;\
```
guicassolato, line 129:
```suggestion
	mkdir -p $$TEMP_DIR ;\
```

otherwise it can fail with `File exists`.

Makefile
```diff
@@ -125,12 +125,13 @@ namespace:
...
+	TEMP_FILE=$$TEMP_DIR/authorino-deploy.yaml ;\
 	cp $(AUTHORINO_CR) $$TEMP_FILE ;\
 	sed -i "s/\$$(AUTHORINO_INSTANCE)/$(AUTHORINO_INSTANCE)/g;s/\$$(TLS_ENABLED)/$(TLS_ENABLED)/g" $$TEMP_FILE ;\
-	$(EDITOR) $$TEMP_FILE ;\
```
guicassolato, line 131: Why removing the editor step?

jjaferson: Added it back after setting a defaulting to vim in case EDITOR is empty

jjaferson:
It was giving permission denied for the factor that EDITOR variable was empty, therefore there is no need to create a `tmp` inside the project directory.

Issue #109, opened 2021-06-14
Fuzzy authorization


guicassolato:
Authorization policies in Authorino are **binary** in the sense that they evaluate to either GRANTED or DENIED, and **atomic** in the sense that either ALL policies shall grant access or no access at all is granted.

This issue is a placeholder to start working on the design of some "fuzzy authorization" in Authorino. The overall idea is that the policies can "partially" grant access to the requested resource, providing enough information to filter the response to the parts that are authorized, while censoring/avoiding the ones that are not.

The main use case we want to support is the "index page" use case, where the response to a request on a determined URL is often a set of resources, in which some of the resources may be authorized for the user while some others are not.

Particularly, I wouldn't expect Authorino to do the actual filtering of the response to the user, but simply a provider of authorization metadata info, so another agent knows what to filter.

PR #216, opened 2022-01-28
Fuzzy OPA

guicassolato:
Adds the option to return all values in the OPA virtual document (all rules), instead of just the "allow" rule that is used by Authorino to decide about the policy overall.

This allows to extract individual values from the Rego virtual documents, e.g. to be used in subsequent authorization policies (i.e. `> priority`) and custom responses and implement use cases such as of fuzzy authorization.

Partial OPA policies can be defined to evaluate complex authorization rules and "export" values via Authorization JSON to other evaluators. The overall authorization decision can be postponed by setting `allow = true` in the partial policies.

E.g.:

```yaml
spec:
  authorization:
  - name: pre
    priority: 0
    opa:
      inlineRego: |
        some_rule = x { x := ... } # some complex computation
        allow = true # postponing the authz decision
      allValues: true

  - name: policy-1
    priority: 1
    opa:
      inlineRego: |
        other_rule { ... }
        allow {
          other_rule
          input.auth.authorization.pre.some_rule != "foo"
        }

  - name: policy-2
    priority: 1
    json:
      rules:
      - selector: auth.authorization.pre.some_rule
        operator: neq
        value: baz

  response:
  - name: pre-authz
    json:
      properties:
      - name: some_rule
        valueFrom: { authJSON: auth.authorization.pre.some_rule }
```

### Verification steps

```sh
make local-setup

kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  identity:
  - name: anonymous
    anonymous: {}
  authorization:
  - name: fuzzy
    opa:
      inlineRego: |
        import input.context.request.http.method

        allowed_methods := ["GET", "POST"]

        is_allowed { allowed_methods[_] == method }

        allow = true
      allValues: true
  response:
  - name: authorization-data
    json:
      properties:
      - name: fuzzy
        valueFrom: { authJSON: auth.authorization.fuzzy }
EOF

kubectl -n authorino port-forward deployment/envoy 8000:8000
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000
# ...
# "Authorization-Data": "{\"fuzzy\":{\"allow\":true,\"allowed_methods\":[\"GET\",\"POST\"],\"is_allowed\":true}}",
# ...
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -X PUT
# ...
# "Authorization-Data": "{\"fuzzy\":{\"allow\":true,\"allowed_methods\":[\"GET\",\"POST\"],\"is_allowed\":null}}",
# ...
```

### Breaking changes

The resolved object returned by the OPA authorization evaluators is no longer a simple boolean value, but now an actual object `{ "allow": boolean, ...other rules }`, where `...other rules` is only available when `allValues: true`.

### Bug fixes

Fixed non-boolean values set for the ‘allow’ rule crashes Authorino (Thanks, @maleck13!)

---

Related to #109

### DIFF
--- a/pkg/config/authorization/opa.go
+++ b/pkg/config/authorization/opa.go
@@ -81,7 +81,7 @@ func (opa *OPA) Call(pipeline common.AuthPipeline, ctx context.Context) (interfa
 			return nil, err
 		} else if len(results) == 0 {
 			return nil, fmt.Errorf(invalidOPAResponseErrorMsg)
-		} else if allowed := results[0].Bindings[allowQuery].(bool); !allowed {
+		} else if allowed, ok := results[0].Bindings[allowQuery].(bool); !ok || !allowed {
 			return nil, fmt.Errorf(unauthorizedErrorMsg)
 		} else {
 			return results[0].Bindings, nil
--- a/pkg/config/authorization/opa.go
+++ b/pkg/config/authorization/opa.go
@@ -8,11 +8,13 @@ import (
 	"io/ioutil"
 	"net/http"
 	"regexp"
+	"strings"
 
 	"github.com/kuadrant/authorino/pkg/common"
 	"github.com/kuadrant/authorino/pkg/common/auth_credentials"
 	"github.com/kuadrant/authorino/pkg/common/log"
 
+	opaParser "github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/rego"
 )
 
@@ -20,15 +22,15 @@ const (
 	policyTemplate = `package %s
 default allow = false
 %s`
-
 	policyUIDHashSeparator = "|"
+	allowQuery             = "allow"
 
 	opaPolicyPrecompileErrorMsg = "Failed to precompile OPA policy"
 	regoDownloadErrorMsg        = "Failed to download Rego data"
 	invalidOPAResponseErrorMsg  = "Invalid response from OPA policy evaluation"
 )
 
-func NewOPAAuthorization(policyName string, rego string, externalSource OPAExternalSource, nonce int, ctx context.Context) (*OPA, error) {
+func NewOPAAuthorization(policyName string, rego string, externalSource OPAExternalSource, fuzzy bool, nonce int, ctx context.Context) (*OPA, error) {
 	logger := log.FromContext(ctx).WithName("opa")
 
 	if rego == "" && externalSource.Endpoint != "" {
@@ -45,6 +47,7 @@ func NewOPAAuthorization(policyName string, rego string, externalSource OPAExter
 	o := &OPA{
 		Rego:              rego,
 		OPAExternalSource: externalSource,
+		Fuzzy:             fuzzy,
 		policyUID:         generatePolicyUID(policyName, rego, nonce),
 		opaContext:        context.TODO(),
 	}
@@ -59,13 +62,14 @@ func NewOPAAuthorization(policyName string, rego string, externalSource OPAExter
 type OPA struct {
 	Rego              string `yaml:"rego"`
 	OPAExternalSource OPAExternalSource
+	Fuzzy             bool
 
 	opaContext context.Context
 	policy     *rego.PreparedEvalQuery
 	policyUID  string
 }
 
-func (opa *OPA) Call(pipeline common.AuthPipeline, ctx context.Context) (bool, error) {
+func (opa *OPA) Call(pipeline common.AuthPipeline, ctx context.Context) (interface{}, error) {
 	var authJSON interface{}
 	if err := json.Unmarshal([]byte(pipeline.GetAuthorizationJSON()), &authJSON); err != nil {
 		return false, err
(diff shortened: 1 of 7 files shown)


guicassolato:
I'm afraid this might affect performance with multiple Rego rules having to be registered as queries, both while pre-compiling the policy and later when the built-in OPA module evaluates the policy and handles control back to Authorino carrying a bigger, more complex response.

Maybe we should put this under a field option added to the API:

```yaml
spec:
  authorization:
    - opa:
        inlineRego: |
           r1 { ... }
           r2 { ... }
           allow { true }
        fuzzy: true # defaults to false
```

pkg/config/authorization/opa.go
```diff
@@ -74,24 +76,45 @@ func (opa *OPA) Call(pipeline common.AuthPipeline, ctx context.Context) (bool, e
...
 
 func (opa *OPA) precompilePolicy() error {
 	policyName := fmt.Sprintf(`authorino.authz["%s"]`, opa.policyUID)
 	policyContent := fmt.Sprintf(policyTemplate, policyName, opa.Rego)
-	regoQuery := rego.Query("allowed = data." + policyName + ".allow")
-	regoModule := rego.Module(opa.policyUID+".rego", policyContent)
+	policyFileName := opa.policyUID + ".rego"
+
+	var module *opaParser.Module
+	rules := make(map[string]interface{})
```
guicassolato, line 96: The "allow" query should always be injected or runtime errors may happen:

```
panic: interface conversion: interface {} is nil, not bool
```

guicassolato: I guess this error would not happen due to https://github.com/Kuadrant/authorino/blob/b07bb9cc6f3d37e2fa9fd17000040ecee32ac650/pkg/config/authorization/opa.go#L21

On the other hand, if https://github.com/Kuadrant/authorino/pull/216#issuecomment-1047097541 is done, then yes. In case of looping through the list of rules to register all the queries is behind a conditional (i.e. only when `fuzzy: true`), then we'd need to ensure the "allow" query is always injected.

guicassolato: Done.

jjaferson (dismissed):
Code looks good!

pkg/config/authorization/opa.go
```diff
@@ -74,24 +78,48 @@ func (opa *OPA) Call(pipeline common.AuthPipeline, ctx context.Context) (bool, e
 		results, err := opa.policy.Eval(opa.opaContext, options)
 
 		if err != nil {
-			return false, err
+			return nil, err
 		} else if len(results) == 0 {
-			return false, fmt.Errorf(invalidOPAResponseErrorMsg)
-		} else if allowed := results[0].Bindings["allowed"].(bool); !allowed {
```
maleck13, line 80: Should we check for the key before casting or will it always be there?

maleck13 (commented):
This looks ok to me, just small comment around the type casting

pkg/config/authorization/opa.go
guicassolato: > Should we check for the key before casting or will it always be there?

It will always be there:

https://github.com/Kuadrant/authorino/blob/65c2751473ddeaf1af6fda7e0e895c1d9c60e040/pkg/config/authorization/opa.go#L21

> what would happen if I wrote a rule called allowed that returned a set here?

I've tested and... It crashes. Nice catch! Thanks!

> Perhaps we have a test for this case?

Now we do 😜

Thanks again!

jjaferson:
Just tested the changes and it works 

```
curl http://talker-api-authorino.127.0.0.1.nip.io:8000
{
  "method": "GET",
  "path": "/",
  "query_string": null,
  "body": "",
  "headers": {
    "Host": "talker-api-authorino.127.0.0.1.nip.io:8000",
    "User-Agent": "curl/7.64.1",
    "Accept": "*/*",
    "X-Forwarded-For": "10.244.0.10",
    "X-Forwarded-Proto": "http",
    "X-Envoy-Internal": "true",
    "X-Request-Id": "e7767f72-f705-4de3-88a9-73b22c20e6ff",
    "Authorization-Data": "{\"fuzzy\":{\"allow\":true,\"allowed_methods\":[\"GET\",\"POST\"],\"is_allowed\":true}}",
    "X-Envoy-Expected-Rq-Timeout-Ms": "15000",
    "Version": "HTTP/1.1"
  },
  "uuid": "04164ef0-7bdc-4158-b73e-eb7fe2e0b350"
}%
```

```
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -X PUT

{
  "method": "PUT",
  "path": "/",
  "query_string": null,
  "body": "",
  "headers": {
    "Host": "talker-api-authorino.127.0.0.1.nip.io:8000",
    "User-Agent": "curl/7.64.1",
    "Accept": "*/*",
    "X-Forwarded-For": "10.244.0.10",
    "X-Forwarded-Proto": "http",
    "X-Envoy-Internal": "true",
    "X-Request-Id": "04818471-86f5-4693-8b86-9ff3b12dbbbc",
    "Authorization-Data": "{\"fuzzy\":{\"allow\":true,\"allowed_methods\":[\"GET\",\"POST\"],\"is_allowed\":null}}",
    "X-Envoy-Expected-Rq-Timeout-Ms": "15000",
    "Version": "HTTP/1.1"
  },
  "uuid": "2273e250-2635-49e8-bdcc-e60f11f2b3ad"
}%
```

jjaferson (approved):

PR #229, opened 2022-03-17
Support for "host" lookup key passed as context extension

### LINKED ISSUE #171: AuthConfig supports wildcard (sub)domains
**Is your feature request related to some problem you are facing? Please describe that problem here.**

I would like to have some auth config for an entire subdomain

```yaml
apiVersion: authorino.3scale.net/v1beta1
kind: AuthConfig
metadata:
  name: my-api-protection
spec:
  hosts:
    - "*.writers.newsletter.io"
```

**Describe the solution you'd like**

The solution should implement some "longest subdomain match" approach. When there is a request, the auth config associated to the longest (sub)domain that matches will apply.

For example, let's say we have the following configuration map
```
owner.writers.petstore.io -> authConfig 01
*.writers.petstore.io -> authConfig 02
*.petstore.io -> authConfig 03
```

The following table shows expected auth configuration resolution

| request authority | expected auth config ID |
| --- | --- |
|  owner.writers.petstore.io | authConfig 01 |
| bob.writers.petstore.io | authConfig 02 |
| alice.writers.petstore.io | authConfig 02 |
| others.readers.petstore.io | authConfig 03 |
| hacker.io | NOT FOUND |

**Additional context**

As a particular case, the *catch-all* or *any* domain could be supported (useful for dev/testing purposes)

```yaml
apiVersion: authorino.3scale.net/v1beta1
kind: AuthConfig
metadata:
  name: my-api-protection
spec:
  hosts:
    - "*"
```

### LINKED ISSUE #222: AuthConfig lookup based on host + path prefix
The AuthConfig reconciler caches and the authz service looks up for an AuthConfig in request-time, using the host names as key (with optional `:port` suffix). An AuthConfig can list multiple host names; each host name will be another key in the map pointing to that AuthConfig.

This lookup mechanism is simple and efficient, relying always in this one piece of info available in the Envoy payload, i.e. the host name. On the other hand, it requires different protected services to have different domain names (e.g. different subdomains), which is not always possible.

In some cases, using a combination of host name and path prefix is preferable to look up for a particular AuthConfig to enforce.

### Discussion

**Related issues**

Supporting host name+path prefix lookups is not the same as the requested in #171, though these requests can be related. Whereas that other issue requests and enhancement of the AuthConfig caching/lookup mechanism to add support for wildcard domains (e.g., AuthConfig lists `*.my.domain`, lookup matches both `foo.my.domain` and `bar.my.domain`), this one is for being able to use as key the host name plus a path prefix (e.g. AuthConfig lists `my.domain/foo`, lookup matches `my.domain/foo*`, but it does not match `my.domain/bar*`).

[`when` conditions](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#common-feature-conditions-when) defined at the upper level of an AuthConfig, introduced by #205, can also be relevant to this request. However, the sematincs is not the same. Conditions are post-lookup, i.e., after using the host name to lookup, Authorino evaluates the conditions to decide whether to effectively enforce the found AuthConfig or not. If the conditions do not match, Authorino returns right away with a `404 Not Found` deny access response and does NOT keep looking for other possible AuthConfig hits.

**Non-trivial implementation**

To maintain _O(1)_ lookup time complexity, unless some very strict rules are enforced regarding what qualifies for a path prefix and that becomes something valid for all AuthConfigs cached by an Authorino instance, supporting the path prefixes in the lookup keys for AuthConfigs is non-trivial.

In request-time, Authorino can only jump straight to a cache entry if it knowns beforehand what to extract from the Envoy payload as the lookup key. As an example, it would not be possible for an AuthConfig to list a lookup key, say, `my.domain/foo` (1-part path prefix `/foo`) and another AuthConfig to use `my.domain/bar/baz` (2-part path prefix `/bar/baz`), and that to a be a detail specified within each AuthConfig. It would have to be always, say, 1-part path prefixes, or always k-part path prefixes, to all AuthConfigs.

If such strict rule cannot be enforced via configuration option for the entire Authorino instance, Authorino would have to go through the entire list of keys to lookup for an AuthConfig, trying to match based on the specific rules specified within each AuthConfig (i.e., for each AuthConfig evaluating a more complex matcher against the Envoy payload), and therefore _O(1)_ lookup could no longer be guaranteed.

Another challenge comes from [preventing collisions](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#avoiding-host-name-collision) between AuthConfigs owned by different users (i.e. defined in different K8s namespaces). Straightforward host name-based caching/lookup makes it simple to prevent users from overriding someone else's AuthConfigs. Supporting more complex matchers can pose a serious challenge to that feature of Authorino.

**Next steps**
To move forward with this issue, first we need to decide whether we want to keep _O(1)_ lookup complexity or if that can be broken to favour more complex matchers.

In the latter case, top-level `when` conditions can provide the base for the implementation, either merging somehow inside of the cache lookup or even replacing it entirely (i.e. lookup fully based on top-level `when` conditions and no longer based on the host names). #171 would be achieved naturally.

On the other hand, efficient host name + path prefix lookup can still be achieved with single extraction rule to obtain the lookup key from the Envoy payload, valid for the entire Authorino instance. In this case, the semantics of top-level `when` conditions does not have to change and #171 will have to continue to be discussed separately.

In both cases, special care must be given to preventing AuthConfigs from being overridden by other specs (defined by other users in different namespaces), that could bend the lookup step in a way that deterministic AuthConfig enforcement can no longer be guaranteed – i.e., it becomes subject to chance due to order of overlapping lookup matchers in the cache. This would be unacceptable of course.

guicassolato:
Adds support for the "host" AuthConfig lookup key to be provided in the payload to the /Check request as Envoy ext-authz per-route "context extension".

Authorino will favor the "host" lookup key whenever passed in the context extensions before the actual "host" HTTP attribute.

This is an alternative approach for enhanced AuthConfig lookups, without changing the lookup structure too much, although relying more on Envoy to set the right payload.

This should be enough to cover use cases such as of path-based AuthConfig lookup and wildcard subdomains AuthConfig lookup.

Closes #222
Closes #171

----

#### Path-based AuthConfig lookup use case

In this use case, 2 different APIs are behind the same base domain and differentiated by the path prefix:
- `pets.com/dogs` →  Dogs API
- `pets.com/cats` →  Cats API

Build and deploy this branch:

```
make local-setup
```

Edit the sample Envoy config to include the required 'host' context extension that will make Authorino to favour this as lookup key before the actual host of the request:

```sh
kubectl -n authorino edit configmap/envoy
```

```yaml
[...]
routes:
- match:
    prefix: /dogs
  route:
    cluster: talker-api
  typed_per_filter_config:
    envoy.filters.http.ext_authz:
      \"@type\": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthzPerRoute
      check_settings:
        context_extensions:
          host: dogs.pets.com
- match:
    prefix: /cats
  route:
    cluster: talker-api
  typed_per_filter_config:
    envoy.filters.http.ext_authz:
      \"@type\": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthzPerRoute
      check_settings:
        context_extensions:
          host: cats.pets.com
[...]
```

In the above, we've used the same upstream API (i.e. `talker-api`) for both routes only for demo purposes. In a real-life scenario, "dogs" and "cats" would be 2 different upstream APIs.

Restart Envoy:

```sh
kubectl -n authorino rollout restart deployments/envoy
```

Forward the requests:

```sh
```

Apply the AuthConfigs:

```sh
kind: AuthConfig
metadata:
  name: dogs
spec:
  hosts:
  - dogs.pets.com
  identity:
  - name: friends
    apiKey:
      labelSelectors:
        api: dogs-api
    credentials:
      in: authorization_header
      keySelector: APIKEY
EOF
```

```sh
kind: AuthConfig
metadata:
  name: cats
spec:
  hosts:
  - cats.pets.com
  identity:
  - name: friends
    apiKey:
      labelSelectors:
        api: cats-api
    credentials:
      in: authorization_header
      keySelector: APIKEY
EOF
```

Create the API keys:

```sh
apiVersion: v1
kind: Secret
metadata:
  name: api-key-1
  labels:
    api: dogs-api
stringData:
type: Opaque
EOF
```

```sh
apiVersion: v1
kind: Secret
metadata:
  name: api-key-2
  labels:
    api: cats-api
stringData:
  api_key: WBGo87IR29ifJKUfxZvL8zLqBKYPRg2g
type: Opaque
EOF
```

Consume the APIs under same base domain:

```sh
curl -H 'Host: pets.com' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' http://localhost:8000/dogs -i
# 200

curl -H 'Host: pets.com' -H 'Authorization: APIKEY WBGo87IR29ifJKUfxZvL8zLqBKYPRg2g' http://localhost:8000/dogs -i
# 401
```

```sh
curl -H 'Host: pets.com' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' http://localhost:8000/cats -i
# 401

curl -H 'Host: pets.com' -H 'Authorization: APIKEY WBGo87IR29ifJKUfxZvL8zLqBKYPRg2g' http://localhost:8000/cats -i
# 200
```

#### Wildcard subdomain AuthConfig lookup use case

In this use case, a single Pets API serves requests for any subdomain that matches `*.pets.com`, e.g.:
- `dogs.pets.com` →  Pets API
- `cats.pets.com` →  Pets API

Build and deploy this branch:

```
make local-setup
```

Edit the sample Envoy config to include the required 'host' context extension that will make Authorino to favour this as lookup key before the actual host of the request:

```sh
kubectl -n authorino edit configmap/envoy
```

```yaml
[...]
virtual_hosts:
- name: local_service
  domains: ['*.pets.com']
  typed_per_filter_config:
    envoy.filters.http.ext_authz:
      \"@type\": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthzPerRoute
      check_settings:
        context_extensions:
          host: pets.com
[...]
```

The "host" context extension above is any key that matches one of the hosts listed in the targeted AuthConfig.

Restart Envoy:

```sh
kubectl -n authorino rollout restart deployments/envoy
```

Forward the requests:

```sh
```

Apply the AuthConfig:

```sh
kind: AuthConfig
metadata:
  name: pets
spec:
  hosts:
  - pets.com
  identity:
  - name: friends
    apiKey:
      labelSelectors:
        group: friends
    credentials:
      in: authorization_header
      keySelector: APIKEY
EOF
```

Create an API key:

```sh
apiVersion: v1
kind: Secret
metadata:
  name: api-key-1
  labels:
    group: friends
stringData:
type: Opaque
EOF
```

Consume the API under different subdomains:

```sh
curl -H 'Host: dogs.pets.com' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' http://localhost:8000 -i
# 200
```

```sh
curl -H 'Host: cats.pets.com' -H 'Authorization: APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx' http://localhost:8000 -i
# 200
```

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -182,6 +182,8 @@ Each phase is sequential to the other, from (i) to (iv), while the evaluators wi
 
 Authorino reads the request host from `Attributes.Http.Host` of Envoy's [`CheckRequest`](https://pkg.go.dev/github.com/envoyproxy/go-control-plane/envoy/service/auth/v3?utm_source=gopls#CheckRequest) type, and uses it as key to lookup in the [cache](#resource-reconciliation-and-status-update) of `AuthConfig`s.
 
+Alternatively, `host` can be supplied in `Attributes.ContextExtensions`, which takes precedence before the actual host attribute of the HTTP request. This is useful to support use cases such as of **path prefix-based lookup** and **wildcard subdomains lookup**.
+
 If more than one host name is specified in the `AuthConfig`, all of them can be used as the key, i.e. all of them can be requested in the authorization request and will be mapped to the same config.
 
 The host can include the port number (i.e. `hostname:port`) or it can be just the name of the host. Authorino will first try finding a config in the cache that is associated to `hostname:port` as supplied in the authorization request; if the cache misses an entry for `hostname:port`, Authorino will then remove the `:port` suffix and lookup again using just `hostname` as key. This allows to change port numbers for a same host, as long as the name of the host is the same, without having to list multiple combinations of `hostname:port` to the `AuthConfig` spec.
--- a/docs/user-guides.md
+++ b/docs/user-guides.md
@@ -60,6 +60,9 @@ Provide Envoy with dynamic metadata from the external authorization process to b
 - **[Redirecting to a login page](./user-guides/deny-with-redirect-to-login.md)**<br/>
 Customize response status code and headers on failed requests. E.g. redirect users of a web application protected with Authorino to a login page instead of a `401 Unauthorized`; mask resources on access denied behind a `404 Not Found` response instead of `403 Forbidden`.
 
+- **[Host override via context extension](./user-guides/host-override.md)**<br/>
+Induce the lookup of an AuthConfig by supplying extended host context, for use cases such as of path prefix-based lookup and wildcard subdomains lookup.
+
 - **[Reducing the operational space: sharding, noise and multi-tenancy](./user-guides/sharding.md)**<br/>
 Have multiple instances of Authorino running in the same space (Kubernetes namespace or cluster-scoped), yet watching particular sets of resources.
 
--- a/docs/user-guides/host-override.md
+++ b/docs/user-guides/host-override.md
@@ -1,0 +1,118 @@
+# Host override via context extension
+
+By default, Authorino uses the host information of the HTTP request ([`Attributes.Http.Host`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/attribute_context.proto#service-auth-v3-attributecontext-httprequest)) to lookup for a cached AuthConfig to be enforced. The host info be overridden by supplying a `host` entry as a (per-route) context entension ([`Attributes.ContextExtensions`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/attribute_context.proto#envoy-v3-api-field-service-auth-v3-attributecontext-context-extensions)), which takes precedence whenever present.
+
+Overriding the host attribute of the HTTP request can be useful to support use cases such as of **path prefix-based lookup** and **wildcard subdomains lookup**.
+
+- [Example of host override for path prefix-based lookup](#example-of-host-override-for-path-prefix-based-lookup)
+- [Example of host override for wildcard subdomain lookup](#example-of-host-override-for-wildcard-subdomain-lookup)
+
+For further details about Authorino lookup of AuthConfig, check out [Host lookup](./../architecture.md#host-lookup).
+
+## Example of host override for path prefix-based lookup
+
+In this use case, 2 different APIs (i.e. **Dogs API** and **Cats API**) are served under the same base domain, and differentiated by the path prefix:
+- `pets.com/dogs` →  Dogs API
+- `pets.com/cats` →  Cats API
+
+Edit the Envoy config to extend the external authorization settings at the level of the routes, with the `host` value that will be favored by Authorino before the actual host attribute of the HTTP request:
+
+```yaml
+virtual_hosts:
+- name: pets-api
+  domains: ['pets.com']
+  routes:
+  - match:
+      prefix: /dogs
+    route:
+      cluster: dogs-api
+    typed_per_filter_config:
+      envoy.filters.http.ext_authz:
+        \"@type\": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthzPerRoute
+        check_settings:
+          context_extensions:
+            host: dogs.pets.com
+  - match:
+      prefix: /cats
+    route:
+      cluster: cats-api
+    typed_per_filter_config:
+      envoy.filters.http.ext_authz:
+        \"@type\": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthzPerRoute
+        check_settings:
+          context_extensions:
+            host: cats.pets.com
+```
+
+Create the AuthConfig for the **Pets API**:
+
+```yaml
+apiVersion: authorino.kuadrant.io/v1beta1
+kind: AuthConfig
+metadata:
+  name: dogs-api-protection
+spec:
+  hosts:
+  - dogs.pets.com
+
+  identity: [...]
+```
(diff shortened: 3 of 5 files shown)


eguzki (approved):
nit: In the verification steps, regarding the "path based routing", how can you tell which configuration is being applied? Maybe two api keys would make it clearer.

guicassolato:
> nit: In the verification steps, regarding the "path based routing", how can you tell which configuration is being applied?

We can only tell from the Authorino logs in the case, I'm afraid.

> Maybe two api keys would make it clearer.

I'll change in the description of the PR (for whoever checks it in the future).

In the user guide for host overriding via context extensions, I made sure to abstract this detail. What matters is: 2 AuthConfigs, the right one will be used in each request based on path-prefix.

Thanks for pointing out, @eguzki!

guicassolato:
> I'll change in the description of the PR (for whoever checks it in the future).

Done.

PR #231, opened 2022-03-22
TTL for Rego policies pulled from external registry

### LINKED ISSUE #221: TTL for external OPA policies
Rego documents [pulled from an HTTP URL](https://github.com/Kuadrant/authorino/blob/50aa03d8e4d27523f37d6e2a333a4826dae23c4f/api/v1beta1/auth_config_types.go#L350) will never expire, requiring a reconciliation of the AuthConfig for a latest version of a policy to be fetched and updated in the cache.

It would be nice to have a "TTL" option to refresh OPA policies fetched from an external registry, similarly to existing OIDC's config's [TTL](https://github.com/Kuadrant/authorino/blob/50aa03d8e4d27523f37d6e2a333a4826dae23c4f/api/v1beta1/auth_config_types.go#L192) option.

guicassolato:
Adds support for `authconfigs.spec.authorization.opa.externalRegistry.ttl` (in seconds, default: `0`– i.e. auto-refresh disabled).

Closes #221

----

### Verification steps

1. Deploy an instance build based on this branch:

```sh
make local-setup DEPLOY_KEYCLOAK=1
```

2. Deploy an OPA server (it will work as external policy registry):

```sh
apiVersion: v1
kind: ConfigMap
metadata:
  name: opa-policies
data:
  allowed-methods.rego: |
    package talkerapi

    method = input.context.request.http.method

    allow { method == "GET" }
    allow { method == "POST" }
  allowed-paths.rego: |
    package talkerapi

    path = input.context.request.http.path

    allow { path == "/" }
EOF
```

```sh
apiVersion: apps/v1
kind: Deployment
metadata:
  name: opa
spec:
  selector:
    matchLabels:
      app: opa
  template:
    metadata:
      labels:
        app: opa
    spec:
      containers:
      - name: opa
        image: openpolicyagent/opa:0.38.1
        args:
        - "run"
        - "--server"
        - "--ignore=.*"
        - "--log-level=debug"
        - "--log-format=json-pretty"
        - "--set=decision_logs.console=true"
        - "/policies"
        ports:
        - name: http
          containerPort: 8181
        volumeMounts:
        - name: policies
          mountPath: /policies
          readOnly: true
      volumes:
      - name: policies
        configMap:
          name: opa-policies
EOF
```

```sh
apiVersion: v1
kind: Service
metadata:
  name: opa
  labels:
    app: opa
spec:
  selector:
    app: opa
  ports:
  - name: http
    port: 8181
    targetPort: 8181
    protocol: TCP
EOF
```

3. Deploy the AuthConfig:

```sh
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  identity:
  - name: anonymous
    anonymous: {}
  - name: keycloak
    oidc:
      endpoint: http://keycloak.authorino.svc.cluster.local:8080/auth/realms/kuadrant
      ttl: 300
    priority: 1 # to never actually use this identity as we just want to test the auto-refresh of the config
  authorization:
  - name: always-allow
    opa:
      inlineRego: |
        allow = true
  - name: allowed-paths
    opa:
      externalRegistry:
        endpoint: http://opa.authorino.svc.cluster.local:8181/v1/policies/policies/allowed-paths.rego
  - name: allowed-methods
    opa:
      externalRegistry:
        endpoint: http://opa.authorino.svc.cluster.local:8181/v1/policies/policies/allowed-methods.rego
        ttl: 60
EOF
```

- The OPA policy `always-allow` is based on inline Rego, therefore it will not change.
- The OPA policy `allowed-path` does not have a `ttl` set, therefore it will not be be refreshed.
- The OPA policy `allowed-methods` will be auto-refreshed from source every 60 seconds.

4. Watch the Authorino instance logs:

```sh
kubectl -n authorino logs -f $(kubectl -n authorino get pods -l authorino-resource=authorino -o name)
```

- You should see an _"openid connect configuration updated"_ at the beginning andevery 5 minutes.
- You should also see an _"auto-refresh of external policy disabled"_ message at the beginning, related to the `allowed-paths` policy.
- Every 60 seconds, you should see an _"external policy unchanged"_ message, related to the `allowed-methods` policy.

```
{"level":"debug","ts":1648029316.0687168,"logger":"authorino.controller-runtime.manager.controller.authconfig.identity.oidc","msg":"openid connect configuration updated","authconfig":"authorino/talker-api-protection","endpoint":"http://keycloak.authorino.svc.cluster.local:8080/auth/realms/kuadrant"}
{"level":"debug","ts":1648029316.3548667,"logger":"authorino.controller-runtime.manager.controller.authconfig.authorization.opa","msg":"auto-refresh of external policy disabled","authconfig":"authorino/talker-api-protection","policy":"authorino/talker-api-protection/allowed-paths","endpoint":"https://raw.githubusercontent.com/guicassolato/authorino-opa/main/allowed-paths.rego","reason":"interval must be greater than zero"}

[…]

{"level":"debug","ts":1648040290.1474848,"logger":"authorino.controller-runtime.manager.controller.authconfig.authorization.opa","msg":"external policy unchanged","authconfig":"authorino/talker-api-protection","policy":"authorino/talker-api-protection/allowed-methods","endpoint":"https://raw.githubusercontent.com/guicassolato/authorino-opa/main/allowed-methods.rego"}
{"level":"debug","ts":1648040350.0213318,"logger":"authorino.controller-runtime.manager.controller.authconfig.identity.oidc","msg":"openid connect configuration updated","authconfig":"authorino/talker-api-protection","endpoint":"http://keycloak.authorino.svc.cluster.local:8080/auth/realms/kuadrant"}
```

5. Send requests to the API:

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -i -X GET
# 200
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -i -X POST
# 200
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -i -X PUT
# 403
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/hello -i -X GET
# 403
```

6. Modify the source Rego for the `allowed-methods` policy, allowing `PUT`:

```sh
curl http://localhost:8181/v1/policies/policies/allowed-methods.rego -X PUT --data $'package talkerapi\n\nmethod = input.context.request.http.method\n\nallow { method == "GET" }\nallow { method == "POST" }\nallow { method == "PUT" }\n' -H 'Content-Type: text/plain'
# {}
```

You should see a _"policy updated from external registry"_ message in the logs after a few seconds:

```
{"level":"info","ts":1648050199.0521312,"logger":"authorino.controller-runtime.manager.controller.authconfig.authorization.opa","msg":"policy updated from external registry","authconfig":"authorino/talker-api-protection","policy":"authorino/talker-api-protection/allowed-methods","endpoint":"http://opa.authorino.svc.cluster.local:8181/v1/policies/policies/allowed-methods.rego"}
```

7. The `PUT` method should now be allowed:

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -i -X PUT
# 200
```

8. Now, try to update the Rego source for the `allowed-paths` policy, allowing `/hello`:

```sh
curl http://localhost:8181/v1/policies/policies/allowed-paths.rego -X PUT --data $'package talkerapi\n\npath = input.context.request.http.path\n\nallow { path == "/" }\nallow { path == "/hello" }' -H 'Content-Type: text/plain'
# {}
```

You will see that, even though the external source has changed, since Authorino is not refreshing this particular policy, no update shall be seen in the logs and the `/hello` path will remain forbidden:

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/hello -i -X GET
# 403
```

**(Optional) Try with a different policy registry** 

Alternatively to using an OPA server as external policy registry, you can use a Github repo, e.g.:

```yaml
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  identity:
  - name: anonymous
    anonymous: {}
  - name: keycloak
    oidc:
      endpoint: http://keycloak.authorino.svc.cluster.local:8080/auth/realms/kuadrant
      ttl: 300
    priority: 1 # to never actually use this identity as we just want to test the auto-refresh of the config
  authorization:
  - name: always-allow
    opa:
      inlineRego: |
        allow = true
  - name: allowed-paths
    opa:
      externalRegistry:
        endpoint: https://raw.githubusercontent.com/guicassolato/authorino-opa/main/allowed-paths.rego
  - name: allowed-methods
    opa:
      externalRegistry:
        endpoint: https://raw.githubusercontent.com/guicassolato/authorino-opa/main/allowed-methods.rego
        ttl: 60
```

Changes committed to the `allowed-methods.rego` file in Github will be reflected in Authorino. The update may take a little longer due to GitHub's cache policy `Cache-Control: max-age=300`.


pkg/config/authorization/opa.go
```diff
@@ -183,3 +232,36 @@ func (ext *OPAExternalSource) downloadRegoDataFromUrl() (string, error) {
...
 }
+
+func (ext *OPAExternalSource) setupRefresher(ctx context.Context, opa *OPA) {
+	logger := log.FromContext(ctx).WithValues("policy", opa.policyName, "endpoint", ext.Endpoint)
+
+	var startErr error
+
+	ext.refresher, startErr = cron.StartWorker(ctx, ext.TTL, func() {
+		if downloadedRego, err := ext.downloadRegoDataFromUrl(); err == nil {
+			if updated, err := opa.updateRego(downloadedRego, ctx, false); updated {
```
eguzki, line 243: shouldn't the error be checked first?

eguzki (commented):
As I understand, for each rego policy from external URL defined, there is a worker (goroutine in the background) started.

Did you considered a different approach instead like the following? 
* Fetch opa policy and store it in memory with the ttl.
* When a new request arrives and tries to evaluate that rego policy, check the TTL. 
* If expired. download again, replace the old one if changed, and evaluate
* if not expired, evaluate.

I think that it scales better when there are lots of external opa authorization defined.

guicassolato:
> Did you considered a different approach instead like the following?

I actually have a feeling that the approach proposed in the PR can scale better.

Imagine a number of requests per second sufficiently large and an instance of Authorino capable of serving multiple requests concurrently. Many concurrent goroutines may be checking the TTL of the policy "at the same time". The check itself is relatively cheap and doesn't require any concurrency control. However, those goroutines may also all decide to update the cache concurrently. So only one of the goroutines does it, we will need a lock.

Even if the other goroutines – i.e. the ones that fail to acquire the lock – give up and proceed right away using the existing version of the policy in the cache (rather than waiting for the update to happen by the one goroutine that succeeded to acquire the lock), acquiring the lock itself has a cost; the negotiation usually involves waiting and sometimes risk of deadlock (depending on implementation).

With the current approach, there is only one goroutine in charge of the updating the cache. I believe the number of concurrent requests in production in any reasonable deployment will beat the number of policies set to expire big time, making one goroutine a lot less expensive compared to the risk of throttling on every tick of cache expiry that coincides with many requests coming in. 

Moreover, this goroutine will be idle most of the time, as authorization policies are not supposed to change very often. I imagine TTLs set in production to the order of hours.

PR #289, opened 2022-05-26
Makefile enhancements

guicassolato:
- new targets: 'help', 'uninstall-operator', plus 'keycloak'and 'dex' (for installing the idps afterwards)
- target renaming: 'manager' -> 'build' and 'example-apps' -> 'user-apps'
- go-get-tool: controller-gen, kustomize, kind and setup-envtest all installed in $PROJECT_DIR/bin
- bump kustomize to v4.5.5

Screenshot 2022-05-26 at 16 28 24

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,101 +29,159 @@ AUTHORINO_CR = $(PROJECT_DIR)/deploy/authorino.yaml
 # Authorino Operator version
 OPERATOR_VERSION ?= latest
 
-.PHONY: vendor fmt vet generate manager manifests run test cover install  uninstall install-operator deploy cert-manager certs
+.PHONY: help
 
-# Download vendor dependencies
-vendor:
+help:
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf " \033[36m%-30s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+##@ Dependencies
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -29,124 +29,148 @@ AUTHORINO_CR = $(PROJECT_DIR)/deploy/authorino.yaml
...
+
+# go-get-tool will 'go get' any package $2 and install it to $1.
+define go-get-tool
+@[ -f $(1) ] || { \
+set -e ;\
+TMP_DIR=$$(mktemp -d) ;\
+cd $$TMP_DIR ;\
+go mod init tmp ;\
+echo "Downloading $(2)" ;\
+GOBIN=$(PROJECT_DIR)/bin go get $(2) ;\
```
eguzki, line 63: I think that go install is better. https://go.dev/doc/go-get-install-deprecation

eguzki (dismissed):

Makefile
guicassolato: Lazy copy & paste from https://github.com/Kuadrant/kuadrant-operator/blob/cd03aa9b374a21eda8bfa7e5d761ec311630bb75/Makefile#L225. Thanks for reminding me!

guicassolato: Ouch! `go install` fails with:

```
Downloading sigs.k8s.io/kustomize/kustomize/v3@v3.8.7
go install: sigs.k8s.io/kustomize/kustomize/v3@v3.8.7 (in sigs.k8s.io/kustomize/kustomize/v3@v3.8.7):
	The go.mod file for the module providing named packages contains one or
	more exclude directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
make[1]: *** [kustomize] Error 1
make: *** [e2e] Error 2
```

guicassolato: We'd need to upgrade kustomize to v4.5.2+.

PR #312, opened 2022-07-01
Support for host names with wildcards

### LINKED ISSUE #306: Wildcard support for `AuthConfig`'s Host field
[Figure this out](https://github.com/orgs/Kuadrant/discussions/3)

guicassolato:
When looking up for an AuthConfig by host name (authorization request), the longest host name match in the cache wins.

E.g., given the following `host -> authconfig` relation:

```
*.io                          -> auth-1
talker-api.nip.io, *.pets.com -> auth-2
api.acme.com                  -> auth-3
*.acme.com                    -> auth-4
```

Expect the following lookup results:

```
dogs.pets.com     -> auth-2 (matches *.pets.com)
api.acme.com      -> auth-3 (matches api.acme.com)
www.acme.com      -> auth-4 (matches *.acme.com)
talker-api.nip.io -> auth-2 (matches talker-api.nip.io)
foo.nip.io        -> auth-1 (matches *.io)
foo.org           -> 404
```

Without support for wildcards:

```
dogs.pets.com     -> 404
api.acme.com      -> auth-3 (matches api.acme.com)
www.acme.com      -> 404
talker-api.nip.io -> auth-2 (matches talker-api.nip.io)
foo.nip.io        -> 404
foo.org           -> 404
```

~Support for wildcards can be enabled/disabled at the level of the Authorino instance by setting `WILDCARDS_ENABLED=true|false` (default: `false`)~
  - [x] ~Open issue in [kuadrant/authorino-operator](https://github.com/kuadrant/authorino-operator) for exposing `WILDCARDS_ENABLED` in the `Authorino` CR~ - https://github.com/Kuadrant/authorino-operator/pull/78

Prevention of host name collisions between k8s namespaces is still subject to the order that the AuthConfigs are reconciled and host names ared listed within each AuthConfig. E.g. if `talker-api.nip.io` is accepted first, other user can still define `*.io` second; on the other hand, if one defines `*.io` first, `talker-api.nip.io` will be considered are taken host name and it will NOT be accepted second.
- #317 

Closes #306

### Verification steps

Build, deploy, setup:

```sh
make local-setup FF=1
```

Apply the AuthConfigs:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: auth-1
spec:
  hosts:
  - "*.io"
  response:
  - name: x-auth-data
    json:
      properties:
      - name: authconfig
        value: auth-1
---
kind: AuthConfig
metadata:
  name: auth-2
spec:
  hosts:
  - "*.pets.com"
  - talker-api.nip.io
  response:
  - name: x-auth-data
    json:
      properties:
      - name: authconfig
        value: auth-2
---
kind: AuthConfig
metadata:
  name: auth-3
spec:
  hosts:
  - api.acme.com
  response:
  - name: x-auth-data
    json:
      properties:
      - name: authconfig
        value: auth-3
---
kind: AuthConfig
metadata:
  name: auth-4
spec:
  hosts:
  - "*.acme.com"
  response:
  - name: x-auth-data
    json:
      properties:
      - name: authconfig
        value: auth-4
EOF
```

It will produce the following tree:

```
                   ┌───┐
         ┌─────────┤ . ├──────────┐
         │         └───┘          │
         │                        │
         │                        │
      ┌──┴─┐                   ┌──┴──┐
  ┌───┤ io ├───┐           ┌───┤ com ├───┐
  │   └────┘   │           │   └─────┘   │
  │            │           │             │
  │            │           │             │
  │            │           │             │
┌─┴─┐       ┌──┴──┐    ┌───┴──┐      ┌───┴──┐
│ * │       │ nip │    │ pets │    ┌─┤ acme ├─┐
└───┘       └──┬──┘    └───┬──┘    │ └──────┘ │
  ▲            │           │       │          │
  │            │           │       │          │
  │            │           │       │          │
  │      ┌─────┴──────┐  ┌─┴─┐  ┌──┴──┐     ┌─┴─┐
auth-1   │ talker-api │  │ * │  │ api │     │ * │
         └────────────┘  └───┘  └─────┘     └───┘
               ▲           ▲       ▲          ▲
               │           │       │          │
               │           │       │          │
               └───auth-2──┘     auth-3     auth-4
```

Send requests:

```sh
curl -H 'Host: dogs.pets.com' http://localhost:8000 -i     # x-auth-data.authconfig == auth-2
curl -H 'Host: api.acme.com' http://localhost:8000 -i      # x-auth-data.authconfig == auth-3
curl -H 'Host: www.acme.com' http://localhost:8000 -i      # x-auth-data.authconfig == auth-4
curl -H 'Host: talker-api.nip.io' http://localhost:8000 -i # x-auth-data.authconfig == auth-2
curl -H 'Host: foo.nip.io' http://localhost:8000 -i        # x-auth-data.authconfig == auth-1
curl -H 'Host: foo.org' http://localhost:8000 -i           # 404
```

_Extra:_ Make things weird by overriding a host name:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: auth-5
spec:
  hosts:
  - talker-api.nip.io
  response:
  - name: x-auth-data
    json:
      properties:
      - name: authconfig
        value: auth-5
EOF
```

```sh
curl -H 'Host: talker-api.nip.io' http://localhost:8000 -i # x-auth-data.authconfig == auth-5
curl -H 'Host: dogs.pets.com' http://localhost:8000 -i     # x-auth-data.authconfig == auth-2
```

```sh
kubectl delete authconfig/auth-5
```

```sh
curl -H 'Host: talker-api.nip.io' http://localhost:8000 -i # x-auth-data.authconfig == auth-1
curl -H 'Host: dogs.pets.com' http://localhost:8000 -i     # x-auth-data.authconfig == auth-2
```


guicassolato:
> I think it would be best if we also allowed this order of creation. I would say having a wildcard host as a fallback would be a quite common use case and having it depend on the order of creation doesn't seem like a great user experience.

We need some heuristic to decide on when a particular host is taken, otherwise users can freely override each other's AuthConfigs. If you somehow also control the routing or in the case of Authorino being used on direct app integration, or as validating webhook service, this can be a serious issue.

Nowadays, host name ownership is pretty straightforward with the lookup based on a simple hash table with no support for wildcards. With wildcards, on the other hand, overriding  some else's `AuthConfig` (entirely or parts of it for portions of the overall range of related host names) – either by sneaking in a specific host name under an existing bigger wildcard config or, the other way around, by adding an umbrella wildcard on top of specific ones – gets harder to handle and to track. (Think: sets and subsets between the host names, with and without wildcards, to better visualize it.)

In the end, what matters is: Who owns the domain? Who owns each subdomain? Making assumption about those question and therefore deciding about which AuthConfigs to accept and which to reject is not easy.

> How would this work with predefined AuthConfigs (one for `*.io`, one for `talker-api.nip.io`) and deploy new Authorino that should reconcile that (e.g. redeploy of Authorino), will it always reconcile the same way?

I'm afraid no particular order can be guaranteed in the reconciliation loop. This problem already exists today without support for wildcards, BTW.

pehala:
> I'm afraid no particular order can be guaranteed in the reconciliation loop. This problem already exists today without support for wildcards, BTW.

In this case, the use-case of having wildcard domain (`*.io`) and then the concrete domain (`talker-api.nip.io`) in its current state, will be inherently unstable and might work for a week and then out of nothing start randomly failing, because redeploying caused it to load in a different order.

> In the end, what matters is: Who owns the domain? Who owns each subdomain? Making assumption about those question and therefore deciding about which AuthConfigs to accept and which to reject is not easy.

I though that there aren't any permission levels in Authorino. If the user can create AuthConfig he can control Authorino and any of those users can manipulate other AuthConfigs, right? If you want more granularity you would have to setup ValidatingWebhook to limit who can create AuthConfig with specific limitations (e.g. domain locking).

I don't know the exact design and philosophy of Authorino, but from my point-of-view, the question of `Who owns the domain?` is not one that Authorino should resolve instead administrators should with their specific setup as I think Authorino doesn't have tools to do that.

didierofrivia:
> E.g., given the following host -> authconfig relation:
>```
>*.io              -> auth-1
>talker-api.nip.io -> auth-2
>api.acme.com      -> auth-3
>*.acme.com        -> auth-4
>```
>Expect the following lookup results:
>```
>dogs.pets.com     -> auth-2 (matches *.pets.com)
>api.acme.com      -> auth-3 (matches api.acme.com)
>www.acme.com      -> auth-4 (matches *.acme.com)
>talker-api.nip.io -> auth-2 (matches talker-api.nip.io)
>foo.nip.io        -> auth-1 (matches *.io)
>foo.org           -> 404
>```

I didn't understand this example until I read the `AuthConfig` auth-2, which enumerates 2 hosts, including the missing `*.pets.com` from the original example.

Regarding the _"longest host name match wins”_  in trivial examples make sense, but when we have many levels of subdomains, which equal _n_ subdivision... it gets tricky to get it right... Maybe we could introduce a _precedence_ param into this equation?

guicassolato:
> I didn't understand this example until I read the `AuthConfig` auth-2, which enumerates 2 hosts, including the missing `*.pets.com` from the original example.

Well spotted, @didierofrivia. I missed that when writing the description of the PR. Fixed.

> Regarding the _"longest host name match wins”_ in trivial examples make sense, but when we have many levels of subdomains, which equal _n_ subdivision... it gets tricky to get it right... Maybe we could introduce a _precedence_ param into this equation?

I would like to avoid one AuthConfig having direct links with another. How do you see the this `precedence` param working on exactly?

pehala:
> This points to removing the [host name collision prevention mechanism](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#avoiding-host-name-collision) from Authorino altogether (or IOW making Authorino to work in cluster-wide mode the same way it does in namespaced mode regarding host names). Not sure how this would be seen.

It could be preserved by allowing wildcard and concrete domain to be only in one namespace (e.g. creating `*.io` in a different namespace than the already existing `test.io` would be rejected and vice versa) could be a partial solution but I can definitely see a use-case where top-level is wildcard domain with defaults set up by the administrator and in the individual namespaces the users could only use only concrete rules, which would not be possible with that limitation. Maybe the host enforcement could be configurable?

guicassolato:
> > This points to removing the [host name collision prevention mechanism](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#avoiding-host-name-collision) from Authorino altogether (or IOW making Authorino to work in cluster-wide mode the same way it does in namespaced mode regarding host names). Not sure how this would be seen.
> 
> It could be preserved by allowing wildcard and concrete domain to be only in one namespace (e.g. creating `*.io` in a different namespace than the already existing `test.io` would be rejected and vice versa) could be a partial

That's how it works already. Avoiding host name collision is only across k8s namespaces. Within the same k8s namespace, Authorino considers the owner of the conflicting AuthConfigs to be the same and accepts both.

guicassolato:
I think what we trying to figure is whether or not there is a way to deliver all the following 3 requirements:
1. Authorino is safe against users defining AuthConfigs that virtually override one another (partially or totally)
2. Users can to define AuthConfigs for sets and subsets of domains with support for wildcards
3. AuthConfigs are simple, without explicit dependencies (references) between them, so Authorino (and users!) can read an AuthConfig alone and know exactly what to expect, without having to refer to who knows how many CRs to figure what applies and what doesn't

didierofrivia:
> I would like to avoid one AuthConfig having direct links with another. How do you see the this precedence param working on exactly?

I'd avoid linking to keep it simple, and what I imagine would be an integer representing the `AuthConfig` " _weight_ ". Say, the higher the number, the higher the precedence. Does it make sense? I know it'd be almost impossible to manage specific rules attached to hundreds of wildcard domains, but I reckon I'd love to have my very specific and most important `AuthConfig`weighted with a `9999`

guicassolato:
> [...] an integer representing the `AuthConfig` " _weight_ ". Say, the higher the number, the higher the precedence. Does it make sense? I know it'd be almost impossible to manage specific rules attached to hundreds of wildcard domains, but I reckon I'd love to have my very specific and most important `AuthConfig`weighted with a `9999`

Yes, it could be a solution for accepting all CRs regardless of overlapping host names, and yet to have a way to decide which AuthConfig to enforce in case multiple apply.

Implementation would need to change a little – we'd need to keep track of the weights associated with each AuthConfig (each matching host) while traversing the tree, to prioritize the one with higher value. Validating webhooks could be used to limit the values users can set for the weight, according to the level of privilege (e.g. higher values kept for admins).

The only remaining issue I see is visibility. A user can still get an AuthConfig accepted but that will never apply, and the user will never know why. E.g.: user _A_ owns `*.pets.com` with weight `100`, user _B_ owns `dogs.pets.com` with weight `50`; `*.pets.com` will always have precedence. As user _B_ who knows nothing about the AuthConfig for `*.pets.com`, I'd rather have my AuthConfig for `dogs.pets.com` rejected at the moment I try to define it, then having to figure out why it's never enforced.

I know the status of the CR could try to reflect that a config is superseded by another. That would mean tracking this in the status updater as well, which could be nightmare given the concurrency. 🤔

guicassolato:
Should wildcards be opt-in (and Authorino defaults to old and simple, more efficient, map-based lookup without wildcards), or should we go all in with host name wildcards, remove the old lookup system, and obviously drop https://github.com/Kuadrant/authorino-operator/pull/78? (Users could still disable wildcards in AuthConfigs by defining a control-plane AuthConfig triggered by a validating webhook request.) 🤔

guicassolato:
> I agree with this being a temporary solution until we properly discuss and decide on the issues, however, I would consider changing the behaviour of the current host name collision so that overlaps are prohibited in both directions e.g. if `*.io` exists we don't allow `test.io` and if `test.io` exists we do not allow `*.io`, due to the instabilities it would bring with each reload.

That would mean no support for special AuthConfigs on specific subdomains, falling back to more generic ones covering the complement. I.e. once you use a wildcard, you wouldn't be able to append anything below that.

Instead, I think I prefer enhancing feedback to the user with better status subresource, so for any AuthConfig we should be able to tell which requested hosts will be served and which will not.

guicassolato:
> I don't think we need to care about the relationship between different policies at different levels.

Exactly. That's why we understand it should be only in Kuadrant.

> Having a straightforward policy of applying all the AuthConfigs while traversing the radix tree seems simple for the user.

The way we implement it should be transparent to the user anyway. In fact, users (of Kuadrant) should know nothing about AuthConfigs, only about AuthPolicies.

> Without the ability to use multiple AuthConfigs per request means there is no value for AuthPolicy to target a Gateway resource unless the Kuadrant controller compiles different AuthConfigs into one (which I don't see how).

That is the general idea. Kuadrant controller compiles different ~AuthConfigs~ AuthPolicies into one single AuthConfig. That is because Kuadrant controller, with the knowledge it has about Gateway API and policy attachment is the only component capable of inferring all the rules from the different kinds of relationships between policies (default, override, merge) attached to the different levels of the tree (Gateway, HTTPRoutes).

> Even if we don't include Kuadrant into the picture, I still see value in authorino giving the ability to take one general AuthConfig and 'x' specific AuthConfigs and apply both for a single request.

Authorino standalone does not need that. It already has a way to represent conditional authorization rules (policies) in one single AuthConfig. In fact, that way is precisely how Kuadrant can compile an AuthConfig out of one or more AuthPolicies.

rahulanand16nov (approved):
for now, lgtm

guicassolato:
Thanks for sharing your thoughts, @rahulanand16nov. Let's keep the conversation going.

Leaving here a few of my own, so the line of thought don't get lost before the next iteration.

> Considering we have literally embedded the AuthConfig into the AuthPolicy, I don't know how users will know nothing about AuthConfig.

I mean what the user sees and owns is a Kuadrant AuthPolicy. The AuthConfig is owned by the Kuadrant controller. The user does not have to be granted permissions of any kind over Authorino AuthConfigs, not even read them.

> Each AuthPolicy creates an AuthConfig so essentially we are compiling multiple AuthConfgs (from Authpolicies) into one AuthConfig (when there is an overlap of hostname).

This is how it works today and would continue to for many use cases. Only for the uses cases where the intent is for policies A _and_ B (or A _or_ B) to be enforced in the same authorization request, where A and B are declared in separate AuthPolicies, we're suggesting translating the two policies into one AuthConfig instead of two.

> I am failing to see how Kuadrant will compile different AuthConfigs into one?

I will stop using "compiling" (because I don't think that's what it is), to favour "merging". The general procedure is: get the auth scheme of AuthPolicy A (it's composed of a few arrays – identity, metadata, authorization, response); get auth scheme of AuthPolicy B (same arrays); unite each pair of compatible arrays together; use the 'name' of each element in the arrays (required field in all the cases) to merge; use the directives 'default' and 'override' to decide which side of the pair takes precedence; use the rules of the AuthPolicies to inject 'when' conditions in the elements.

Again, the procedure above is when we want two policies evaluated in the same authorization request; only to give more granularity for users (possibly different users) to own and apply policies at a different level, yet targeting the same requests to the protected service.

> Policy attachment puts a guard rail around AuthConfigs to avoid overlapping policies and that's it.

I think it is (can be?) more than that. Once you add the knowledge it has about the policy 'rules' (i.e. the scope where the spec should apply) and the kind of relationship that dictates behaviour in case of overlap (default/override, either to an entire group of rules or partial), it opens up for more things IMO. It can go beyond deciding which nodes of the routing tree to just enable or disable the call to the external authorization/RL service. It can add more context, it can decide to throw parts of the spec away to favour others.

> Even in the case of RLP, compiling is done by Limitador and not Kuadrant.

I see it differently. Kuadrant controller builds one Limitador CR (or maybe one RL config file these days?) with all the rate limit definitions gathered from all RateLimitPolicy CRs. So it's already a merge. Then, how do we tell Limitador when a limit is to be treated as a default or as an override? I suppose we don't. Instead, Kuadrant controller has to decide which limit goes to the Limitador config, which doesn't.

Because Limitador does not work in terms of 'hosts' as groupings of rules – and neither Authorino BTW(*) – we might perceive an AuthConfig as too different from configuring Limitador. I believe that, in the end, they are not that different.

(*) only reason for hosts to be so important in Authorino is lack of access to context extensions via Istio AuthorizationPolicy

> If we solve this at the Authorino level, it becomes a tree traversal problem but if at the Kuadrant level, it becomes a compiler problem. I feel I don't have enough understanding of AuthConfig to combine multiple of them into one. So to make sure I don't push for something that is easy for me, can you please describe a way to compile AuthConfigs?

How to achieve multiple AuthConfigs per authz request was never a concern to me – wrap evaluation in a loop, trigger workloads in parallel, shove all GW API stuff about defaults, overrides, merges, etc into the AuthConfig (so Authorino knows the intended behaviour on the overlaps), some channels to control and sync, etc.

My concern is why doing all that in Authorino and in Limitador as well? Apart from neither of those components needing that today to support the use case behind, to me it all points to Kuadrant controller to be the right place to solve the issue. Kuadrant controller is the first common node in the tree of components that we own and where the info needed to solve occasional conflicts passes through, it knows about the _wheres_, _whens_ and _ifs_ (from the policy rules and different types resources targeted by the policies), and most importantly it is where our Kuadrant implementation of GW API is.

So I guess my point here is, IMO, even if solving the issue in Kuadrant controller is harder than handing it over to each component (Authorino, Limitador), something yet to be proven BTW, it is where it belongs nonetheless.

PR #318, opened 2022-07-11
AuthConfig status conditions

### LINKED ISSUE #314: Better AuthConfig/status sub resource for readiness check
`AuthConfig/Status` sub resource is very limited regarding the information is offers for readiness check. Effectively, only one field (`Ready: bool`) currently tries to reflect the readiness state of the AuthConfig CR. Even so, it's very reliable:
- It won't take into account the state for all hosts listed in the `AuthConfig` – as long as at least one of the hosts listed in the `hosts` field of the CR can be found associated in the cache, it will already consider the entire resource ready, instead of checking the state for all hosts;
- Only the state of the cache in the Authorino replica that happens to be leader (and therefore also the one taking care of status updates) is checked to update the status of the CR – i.e., if the CR has not been accepted by other replicas of Authorino – something to be treated as an anomaly after a significant amount of time passed, nevertheless possible and expected within small time windows, until all replicas finish updating their internal cache states –, that will not be reflected in the status readiness flag either;
- Unprevented host name collisions (AuthConfigs in the same k8s namespace) won't be reflected in the status either.

Apart from readiness state, the status sub resource offers just a summary of the AuthConfig spec, in the form of the simple counter indicators:
- `numIdentitySources` (int64)
- `numMetadataSources` (int64)
- `numAuthorizationPolicies` (int64)
- `numResponseItems` (int64)
- `festivalWristbandEnabled` (bool)

Perhaps not all of the above needs to be enhanced right away, but at least proper report of the state for the different hosts, separation of readiness conditions from AuthConfig summary would both be nice to have.

guicassolato:
Adds conditions and proper summary of the status of the AuthConfig, taking into account the state for each host name listed in `spec.hosts`.

Two conditions are defined:
- `Available` means at least one host from `spec.hosts` linked to AuthConfig in the cache
- `Ready` means all host names from `spec.hosts` linked to the AuthConfig in the cache

When it is not possible to link all hosts to the AuthConfig, due to a host already taken by another AuthConfig in a different k8s namespace, the AuthConfig will never be `Ready` (condition reason `HostNotLinked`). It can still be `Available` though (activated via other hosts).

Users can read the list of hosts that succeeded being linked to the AuthConfig in `status.summary.hostsReady: []string`.

**Summary status conditions and possible states:**

| Type        | Status  | Reason           | Message                                      | Meaning |
|-------------|---------|------------------|----------------------------------------------|---------|
| `Available` | `True`  | `HostsLinked`    | N/A                                          | At least one host from `spec.hosts` was linked to the AuthConfig in the cache. Requests sent to the linked host can be expected to be protected by this AuthConfig. User can check `status.summary.hostsReady` for the list of host names that did succeed being linked to the AuthConfig in the cache. |
| `Available` | `False` | `HostsNotLinked` | No hosts linked to the resource.             | No host from `spec.hosts` was linked to the AuthConfig in the cache. No request will cause this AuthConfig to be evaluated by Authorino. Only possible status for `Ready` is `False` in this case. |
| `Ready`     | `True`  | `Reconciled`     | N/A                                          | The AuthConfig is cached for all host names from `spec.hosts`. |
| `Ready`     | `False` | `HostsNotLinked` | One or more hosts not linked to the resource | Failed to cache the AuthConfig for one or more host names from `spec.hosts`. User should check `status.summary.hostsReady` for the list of host names that did succeed being linked to the AuthConfig in the cache; occasionally none. |

The counters and indicators in the status sub resource, used as well to define the print columns when listing the authconfig resources, were moved into Status.Summary.

The following indicators have been added to the summary:
- `hostsReady []string` - list of the host names from `spec.hosts` linked to the AuthConfig in the cache
- `numHostsReady string` - indicator of the number of hosts linked to the AuthConfig in the cache, in the format `n/N`, where `n` is number of hosts linked and `N` is the total number of hosts in `spec.hosts`.
 
Column names were also modified for simplicity.

Closes #314.

### Verification steps

```sh
make local-setup # clusterWide: true

kubectl create namespace user-1
kubectl -n user-1 apply -f -<<EOF
kind: AuthConfig
metadata:
  name: auth-1
  namespace: user-1
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  response:
  - name: x-auth-data
    json:
      properties:
      - name: authconfig
        value: auth-1
EOF

kubectl create namespace user-2
kubectl -n user-2 apply -f -<<EOF
kind: AuthConfig
metadata:
  name: auth-2
  namespace: user-2
spec:
  hosts:
  - talker-api.io
  - talker-api.127.0.0.1.nip.io
  response:
  - name: x-auth-data
    json:
      properties:
      - name: authconfig
        value: auth-2
EOF
```

```sh
# NAMESPACE   NAME     READY   HOSTS   AUTHENTICATION   METADATA   AUTHORIZATION   RESPONSE   WRISTBAND
# user-1      auth-1   true    1/1     0                0          0               1          false
# user-2      auth-2   false   1/2     0                0          0               1          false
```

```sh
kubectl -n user-1 get authconfig/auth-1 -o jsonpath={.status}
# {"conditions":[{"lastTransitionTime":"2022-07-12T17:30:16Z","reason":"HostsLinked","status":"True","type":"Available"},{"lastTransitionTime":"2022-07-12T17:30:16Z","reason":"Reconciled","status":"True","type":"Ready"}],"summary":{"festivalWristbandEnabled":false,"hostsReady":["talker-api.127.0.0.1.nip.io"],"numAuthorizationPolicies":0,"numHostsReady":"1/1","numIdentitySources":0,"numMetadataSources":0,"numResponseItems":1,"ready":true}}
```

```sh
kubectl -n user-2 get authconfig/auth-2 -o jsonpath={.status}
# {"conditions":[{"lastTransitionTime":"2022-07-12T17:30:16Z","reason":"HostsLinked","status":"True","type":"Available"},{"lastTransitionTime":"2022-07-12T17:30:16Z","message":"One or more hosts not linked to the resource","reason":"HostsNotLinked","status":"False","type":"Ready"}],"summary":{"festivalWristbandEnabled":false,"hostsReady":["talker-api.io"],"numAuthorizationPolicies":0,"numHostsReady":"1/2","numIdentitySources":0,"numMetadataSources":0,"numResponseItems":1,"ready":false}}
```


controllers/auth_config_status_updater.go
```diff
@@ -51,12 +51,14 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 }
 
-func (u *AuthConfigStatusUpdater) updateAuthConfigStatus(ctx context.Context, cacheId string, authConfig *api.AuthConfig, ready bool) error {
-	if len(u.Cache.FindKeys(cacheId)) == 0 {
+func (u *AuthConfigStatusUpdater) updateAuthConfigStatus(ctx context.Context, cacheId string, authConfig *api.AuthConfig) error {
+	if len(u.Cache.FindKeys(cacheId)) != len(authConfig.Spec.Hosts) {
```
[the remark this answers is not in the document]
guicassolato: This changes significantly in the next commit.

controllers/auth_config_status_updater.go
```diff
@@ -41,41 +46,152 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
...
 	}
 }
 
-func (u *AuthConfigStatusUpdater) updateAuthConfigStatus(ctx context.Context, cacheId string, authConfig *api.AuthConfig, ready bool) error {
-	if len(u.Cache.FindKeys(cacheId)) == 0 {
-		return fmt.Errorf("resource not ready")
+func (u *AuthConfigStatusUpdater) updateAuthConfigStatus(ctx context.Context, cacheId string, authConfig *api.AuthConfig) (err error) {
+	logger := log.FromContext(ctx)
+
+	var condition api.Condition
```
maleck13, line 60: nit: would it be reasonable to remove the else and just have var condition api.Condition = statusReady()

controllers/auth_config_status_updater.go
```diff
@@ -41,41 +46,152 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
...
+	authConfig.Status.Conditions, conditionsChanged = buildStatusConditions(authConfig.Status.Conditions, condition)
+	authConfig.Status.Summary, summaryChanged = buildStatusSummary(authConfig, linked)
+
+	if !authConfig.Status.Ready() {
+		err = fmt.Errorf("resource not ready")
 	}
-	authConfig.Status.FestivalWristbandEnabled = issuingWristbands
 
-	return u.Status().Update(ctx, authConfig)
+	if changed := conditionsChanged || summaryChanged; !changed {
```
maleck13, line 77: this is a little difficult to read is it equiv to ``` ! conditionsChanged && ! summaryChanged ```

controllers/auth_config_status_updater.go
```diff
@@ -41,41 +46,152 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
...
+		logger.V(1).Info("resource status did not change")
+		return // to save an update request
+	}
+
+	logger.V(1).Info("resource status changed", "authconfig/status", authConfig.Status)
+
+	if updateErr := u.Status().Update(ctx, authConfig); updateErr != nil {
+		logger.Error(updateErr, "failed to update the resource")
```
maleck13, line 85: when an error is returned does something at the top level also log the error? Wondering will we end up with double log lines ?

controllers/auth_config_status_updater.go
```diff
@@ -41,41 +46,152 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
...
+		err = updateErr
```
maleck13, line 86: should there be a return here?

controllers/auth_config_status_updater.go
```diff
@@ -41,43 +46,152 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
...
-		return fmt.Errorf("resource not ready")
+func (u *AuthConfigStatusUpdater) updateAuthConfigStatus(ctx context.Context, cacheId string, authConfig *api.AuthConfig) (err error) {
+	logger := log.FromContext(ctx)
+
+	var condition api.Condition
+	linked, loose := u.partitionHostsByStatus(cacheId, authConfig)
+	if len(loose) > 0 {
+		condition = statusNotReady(api.StatusReasonHostNotLinked, "One or more hosts not linked to the resource")
+	} else {
+		condition = statusReady()
```
didierofrivia, line 65: As discussed before, maybe we could marked as `Ready` even when it's "partially" applied, some hosts were already taken.
And find a different status for the "hostsNotLinked"

controllers/auth_config_status_updater.go
```diff
@@ -41,41 +46,152 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
...
+	return api.Condition{
+		Type:    api.StatusConditionReady,
+		Status:  k8score.ConditionFalse,
+		Reason:  reason,
+		Message: message,
+	}
+}
+
+func buildStatusConditions(current []api.Condition, new api.Condition) ([]api.Condition, bool) {
+	new.LastTransitionTime = metav1.Now()
```
maleck13, line 136: nit: might be worth naming this something like ```next``` rather than new that is also a keyword in go?

maleck13 (commented):
changes look reasonable. Couple of small comments

controllers/auth_config_status_updater.go
```diff
@@ -41,41 +46,152 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
...
+}
+
+func (u *AuthConfigStatusUpdater) partitionHostsByStatus(cacheId string, authConfig *api.AuthConfig) ([]string, []string) {
+	type obj struct{}
+	hosts := make(map[string]obj)
+	for _, host := range u.Cache.FindKeys(cacheId) {
+		hosts[host] = obj{}
+	}
+	linked := []string{}
+	loose := []string{}
```
didierofrivia, line 101: `var linked, loose []string`

controllers/auth_config_status_updater.go
guicassolato: No double log lines.

This is an exceptional case. It should only happen if the AuthConfig becomes stale in the middle of the status update (due to concurrent updates going on, directly by the user) or some exception in the Authorino <-> Kube API server comm. Nonetheless, the important part is retrying the status update is that happens, and that is being done.

controllers/auth_config_status_updater.go
guicassolato: Yep.  Well... a `return` or an `else`, to avoid printing "resource status updated" when actually it wasn't.
Thanks!

controllers/auth_config_status_updater.go
guicassolato: Yes, @didierofrivia. The option I made before for `Ready=false` when there are actually hosts linked to the AuthConfig was also bugging me. Maybe we can do better with more conditions instead of one boolean condition + reasons. Let us think about that for a while.

controllers/auth_config_status_updater.go
guicassolato: It raises an error in cases where no hosts are linked, therefore nothing is ever appended to `linked`, and later `linked` is used to populate `summary.HostsReady`. Since the field does not accept null values, of fails. That's why I default to initialising the slices always.

controllers/auth_config_status_updater.go
```diff
@@ -41,41 +46,152 @@ func (u *AuthConfigStatusUpdater) Reconcile(ctx context.Context, req ctrl.Reques
...
+					return current, false
+				}
+
+				new.LastTransitionTime = curr.LastTransitionTime
+			}
+
+			res := make([]api.Condition, len(current))
+			copy(res, current)
+			res[i] = new
+			return res, true
```
didierofrivia, line 155: Could it be the case that a new condition that has same type, different status would force us return an "early" array of conditions leaving behind other conditions to check?

guicassolato: Not really, because `buildStatusConditions` focuses on one "new" condition at a time. The goal is to get that one condition either added or updated in the existing array of conditions. Once that one is found, there's no need to keep checking for others.

We could be invoking `buildStatusConditions` multiple times, accumulating changes, one by one, into a variable `conditions` and then set `authConfig.Status.Conditions = conditions`. Right now, the whole status update only adds/updates one condition; because of that, we set `authConfig.Status.Conditions, ... = buildStatusConditions(...)` directly.

controllers/auth_config_status_updater.go
guicassolato: IDK if it's really "next". The semantics is: "given an existing (current) list of conditions, ensure this one is in the list".

It's more like an "add or update", copied from [here](https://github.com/Kuadrant/authorino-operator/blob/4bc68ef986c4d0ab463ff2ccabcb71d7d54fe0d4/pkg/condition/condition.go#L11), where the function is called – guess what? – "AddOrUpdateStatusConditions". Should I have stuck with that instead of "buildStatusConditions"?

guicassolato: Renamed to `updateStatusConditions(currentConditions []api.Condition, newCondition api.Condition)`.

controllers/auth_config_status_updater.go
guicassolato: Fixed

maleck13 (approved):

didierofrivia (approved):
The addition of the status "Available" when there are hosts linked makes total sense, still not sure about "Ready" when there are any loose ones but most likely using it and time will prove us we were on the right path. 🥇

guicassolato:
> not sure about "Ready" when there are any loose ones

Ready should be true only when there are no loose ones, @didierofrivia.

https://github.com/Kuadrant/authorino/blob/6bf5315a63e2a0cdc580623cf088534506476cd4/controllers/auth_config_status_updater.go#L66

PR #321, opened 2022-07-11
[makefile] Default EDITOR set to vi if undefined

guicassolato:
To prevent `make deploy` of raising a permission error when the `EDITOR` env var is not set.


didierofrivia (approved):
👍🏼

PR #330, opened 2022-07-21
Bootstrap cache index on startup

### LINKED ISSUE #317: Order of reconcilliation of the AuthConfigs should not matter to rebuild the cache
The outcome state of the reconcilliation of AuthConfigs containing host overlaps is subject to the order that the CRs are processed by Authorino and added to the cache.

The behaviour depends on the reconcilliation mode set for the Authorino instance, either namespaced or cluster-wide.

<details>
  <summary>What's the difference?</summary>

  <ul>
    <li><b>Cluster-wide:</b> Authorino refuses to override the currently cached auth rules when the host name is already taken and the two AuthConfigs involved (i.e. the one cached and the new incoming one) are from different k8s namespaces.</li>
    <li><b>Namespaced:</b> the AuthConfig that comes after always overrides what's in the cache.</li>
  </ul>
</details>

Nonetheless, effectively in both cases the order that the AuthConfigs are reconciled matters, potentially leading to inconsistent states of the cache for overlapping host names after a pod occasionally restarts.

### Example

Given:
- AuthConfig _A_ specifies a set of auth rules _a_ for a host name _h_;
- AuthConfig _B_ specifies a set of auth rules _b_ for also host name _h_.

If _A_ is reconciled and then comes _B_,
- In cluster-wide mode:
  - Authorino rejects _B_ because _h_ is 'already taken' (linked to _A_) and therefore _a_ will be enforced.
- In namespaced mode:
  - Authorino accepts _B_ overriding _a_ with _b_ which thereafter will be enforced to protect _h_.

(And vice-versa if the order is instead first _B_, then _A_.)

If the Authorino pod restarts, the order that the CRs will be reconciled cannot be guaranteed and therefore the state of the cache can end up different than the one before restarting.

The table below summarises for all the cases:

<table>
  <thead>
    <tr>
      <th rowspan="2">Reconcilliation mode</th>
      <th colspan="2">Pod running at <i>t</i></th>
      <th colspan="2">Pod running at <i>t+1</i> after restart</th>
      <th rowspan="2">#</th>
    </tr>
    <tr>
      <th>Order</th>
      <th>State of the cache for <i>h</i> (<i>t</i>)</th>
      <th>Order</th>
      <th>State of the cache for <i>h</i> (<i>t+1</i>)</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td rowspan="4">Cluster-wide</td>
      <td><i>A</i> → <i>B</i></td>
      <td><i>a</i></td>
      <td><i>A</i> → <i>B</i></td>
      <td><i>a</i></td>
      <td>1</td>
    </tr>
    <tr>
      <td><i>A</i> → <i>B</i></td>
      <td><i>a</i></td>
      <td><i>B</i> → <i>A</i></td>
      <td><i>b</i></td>
      <td>2</td>
    </tr>
    <tr>
      <td><i>B</i> → <i>A</i></td>
      <td><i>b</i></td>
      <td><i>B</i> → <i>A</i></td>
      <td><i>b</i></td>
      <td>3</td>
    </tr>
    <tr>
      <td><i>B</i> → <i>A</i></td>
      <td><i>b</i></td>
      <td><i>A</i> → <i>B</i></td>
      <td><i>a</i></td>
      <td>4</td>
    </tr>
    <tr>
      <td rowspan="4">Namespaced</td>
      <td><i>A</i> → <i>B</i></td>
      <td><i>b</i></td>
      <td><i>A</i> → <i>B</i></td>
      <td><i>b</i></td>
      <td>5</td>
    </tr>
    <tr>
      <td><i>A</i> → <i>B</i></td>
      <td><i>b</i></td>
      <td><i>B</i> → <i>A</i></td>
      <td><i>a</i></td>
      <td>6</td>
    </tr>
    <tr>
      <td><i>B</i> → <i>A</i></td>
      <td><i>a</i></td>
      <td><i>B</i> → <i>A</i></td>
      <td><i>a</i></td>
      <td>7</td>
    </tr>
    <tr>
      <td><i>B</i> → <i>A</i></td>
      <td><i>a</i></td>
      <td><i>A</i> → <i>B</i></td>
      <td><i>b</i></td>
      <td>8</td>
    </tr>
  </tbody>
</table>

Obviously cases 2, 4, 6 and 8, where state of the cache at _t+1_ is not the same as at _t_, are not desirable.

Ideally, Authorino would repopulate the cache after a pod restarts in a way such that the same state is guaranteed before and after the restart, regardless of the order that the AuthConfigs are reconciled – possibly by concurrent threads.

guicassolato:
This fix is needed for when a new Authorino pod starts and finds pre-existing AuthConfigs marked as available (i.e. at least one host of the AuthConfig linked in the cache), therefore will populate the cache by reconciling all those resources.

We want for the new pod to respect the order of creation/acceptance of the resources when populating the cache, so previous decisions regarding host collisions are repeated exactly as before. If that is not the case, we risk incurring on issues such as the one described in #317.

A few aspects of the proposed solution:
- Whenever the reconciler (triggered by whatever reconciliation event) spots the cache empty, it will try to bootstrap the cache index before continuing with the actual reconciliation workload.
- A lock prevents multiple goroutines from trying to rebuild the cache index at the same time.
- To rebuild the cache, it first sorts the AuthConfigs by `creationTimestamp`.
  - [ ] Use the timestamp of when the resource was last updated instead.
- When bootstrapping, it only rebuilds the cache index of keys (hosts), associating each host with a trivial deny-all policy (returns `503 Busy` status code); the bootstrapping function then gives control back to the reconcilers, which will fill the cache with the proper parsed AuthConfig specs, performed as cache updates, and thus replacing the temporary deny-all policy.

Closes #317.

### Verification steps

Build, deploy (choose cluster-wide mode) and setup the env with all example apps (Talker API, Envoy, etc):

```sh
make local-setup # clusterWide: true
```

Create a couple AuthConfigs in 2 different namespaces:

```sh
kubectl create namespace admin
kubectl create namespace userapps

kubectl -n admin apply -f -<<EOF
kind: AuthConfig
metadata:
  name: default-deny-all-policy
  namespace: admin
spec:
  hosts:
  - "*.127.0.0.1.nip.io"
  authorization:
  - name: deny-all
    opa:
      inlineRego: |
        allow = false
EOF

kubectl -n userapps apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
  namespace: userapps
spec:
  hosts:
  - talker-api.userapps.cluster.local
  - talker-api.127.0.0.1.nip.io # will be rejected to avoid partially overriding admin/default-deny-all-policy
EOF
```

Check the state we want Authorino to reconcile to regardless of the order that the AuthConfigs are picked by the reconciler in case of populating the cache from scratch (pod restart, new replica, etc):

```sh
# NAMESPACE   NAME                      READY   HOSTS   AUTHENTICATION   METADATA   AUTHORIZATION   RESPONSE   WRISTBAND
# admin       default-deny-all-policy   true    1/1     0                0          1               0          false
# userapps    talker-api-protection     false   1/2     0                0          0               0          false
```

Restart the Authorino pod, so it bootstraps the cache again.

```sh
kubectl rollout restart deployment/authorino
```

You may need to repeat the previous step several times, until eventually you spot something like the following in the Authorino logs:

```
kubectl logs $(kubectl get pods -l authorino-resource=authorino -o name)
# 1.6583995609075952e+09	INFO	authorino.controller-runtime.manager.controller.authconfig.bootstrap	building cache index
# 1.6583995609088998e+09	DEBUG	authorino.controller-runtime.manager.controller.authconfig.bootstrap	building cache index	{"authconfig": "admin/default-deny-all-policy"}
# 1.6583995609089446e+09	DEBUG	authorino.controller-runtime.manager.controller.authconfig.bootstrap	building cache index	{"authconfig": "userapps/talker-api-protection"}
# 1.6583995609089522e+09	INFO	authorino.controller-runtime.manager.controller.authconfig.bootstrap	host already taken in another namespace	{"authconfig": "userapps/talker-api-protection", "host": "talker-api.127.0.0.1.nip.io"}
# 1.6583995609091828e+09	INFO	authorino.controller-runtime.manager.controller.authconfig	host already taken in another namespace	{"authconfig": "userapps/talker-api-protection", "host": "talker-api.127.0.0.1.nip.io"}
# 1.6583995609091964e+09	INFO	authorino.controller-runtime.manager.controller.authconfig	resource reconciled	{"authconfig": "userapps/talker-api-protection"}
# 1.6583995609098206e+09	INFO	authorino.controller-runtime.manager.controller.authconfig	resource reconciled	{"authconfig": "admin/default-deny-all-policy"}
```

Notice that the AuthConfigs were picked to rebuild the cache index in the same order they were created. However, later the AuthConfigs were picked by the reconciler, to effectively fill the cache, in a different order. Nevertheless, the resulting state regarding host overrides was the same as when they AuthConfigs were orignally created, i.e., with `talker-api.127.0.0.1.nip.io` taken by `admin/default-deny-all-policy` and rejected to be linked to `userapps/talker-api-protection`.

```sh
# NAMESPACE   NAME                      READY   HOSTS   AUTHENTICATION   METADATA   AUTHORIZATION   RESPONSE   WRISTBAND
# admin       default-deny-all-policy   true    1/1     0                0          1               0          false
# userapps    talker-api-protection     false   1/2     0                0          0               0          false
```


api/v1beta1/auth_config_types.go
```diff
@@ -698,7 +698,21 @@ type AuthConfig struct {
...
+	Items           AuthConfigSlice `json:"items"`
+}
+
+type AuthConfigSlice []AuthConfig
+
+func (s AuthConfigSlice) Len() int {
+	return len(s)
+}
+
+func (s AuthConfigSlice) Less(i, j int) bool {
```
didierofrivia, line 710: Ok, so `sort.Sort` will sort the `Items` slice based on `CreationTimestamp` comparison :+1:

controllers/auth_config_controller.go
```diff
@@ -109,23 +117,16 @@ func (r *AuthConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-				}
-			}
+		linkedHosts, looseHosts, err = r.addToCache(log.IntoContext(ctx, logger), req.Namespace, resourceId, evaluatorConfigByHost)
 
-			if err := r.Cache.Set(resourceId, host, evaluatorConfig, true); err != nil {
-				r.StatusReport.Set(resourceId, api.StatusReasonCachingError, err.Error(), linkedHosts)
-				return ctrl.Result{}, err
-			}
+		if len(looseHosts) > 0 {
+			r.StatusReport.Set(resourceId, api.StatusReasonHostsNotLinked, "one or more hosts not linked to the resource", linkedHosts)
```
didierofrivia, line 123:
```suggestion
			r.StatusReport.Set(resourceId, api.StatusReasonHostsNotLinked, "one or more hosts are not linked to the resource", linkedHosts)
```

controllers/auth_config_controller.go
```diff
@@ -589,6 +590,84 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
...
+		}
+
+		authConfigName := types.NamespacedName{Namespace: authConfig.Namespace, Name: authConfig.Name}
+		logger.V(1).Info("building cache index", "authconfig", authConfigName)
+
+		_, _, err := r.addToCache(
+			log.IntoContext(ctx, logger.WithValues("authconfig", authConfigName)),
+			authConfig.Namespace,
+			authConfigName.String(),
+			evaluatorConfigByHost,
```
didierofrivia, line 660: I think I need of context to understand the crafted before "deny all" policy added to the cache here

didierofrivia: Ok, just read it... then it willl update with the true values from the `AuthConfig`s replacing the bootstrapped deny all one :+1:

didierofrivia (dismissed):
It works like a charm! :+1:

api/v1beta1/auth_config_types.go
guicassolato: Correct

alexsnaps (approved):

guicassolato:
Thanks, @didierofrivia and @alexsnaps for the excellent inputs!

Leaving here for a future iteration:
- Improvements in the sorting of the resources, maybe based on the time when last updated (last marked as available?), instead of the creation timestamp.
- Better ways to avoid queuing an excessively large number of goroutines, all trying to process reconciliation events in concurrency, while one rebuilds the index.

PR #340, opened 2022-08-02
Option to disable the prevention of host name collision

### LINKED ISSUE #329: Possibility to disable host collision prevention
To support use cases of AuthConfigs defined across different namespaces, involving hosts already taken and where collision/override policies are handled upfront by other means (validating webhook, custom controller, etc), we want to be able to switch Authorino's [host name collision prevention](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#avoiding-host-name-collision) off.

**_Attention!_**<br/>Without prevention of host name collisions, users working in different K8s namespaces will be able to take control of each other's AuthConfigs – by simply defining an AuthConfig _X_ that includes a host name that is either more specific than (for partial API auth override) or identical to (full override) a host name listed by another AuthConfig _Y_ which was accepted by the Authorino instance prior. This is already the case for AuthConfigs created in the same K8s namespace, where some coordination between users is expected regarding host names listed in their AuthConfigs. Therefore, a switch that turns off the host name collision prevention is to be used with care. Either such collisions do not matter for the target use case or it is imperative that their are handled by other means before Authorino tries to reconcile the AuthConfigs.

guicassolato:
Adds a setting to disable the prevention of host name collision also in cluster-wide mode, at the level of the Authorino deployment.

The PR introduces a new environment variable `PREVENT_COLLISIONS` (default: `true`) that when set to `false` will cause Authorino to behave in cluster-wide reconciliation mode just like it does in namespaced mode, regarding preventing collisions between AuthConfigs that claim association to identical host names.

**_Breaking change alert!_**

Because now this configuration controls if users can or cannot replace each others' AuthConfigs (fully, by choosing identical host names; or partially, by choosing more specific host names that match more general host wildcards), we are unifying the two traditional behaviours for namespaced mode and for cluster-wide modes into one single behaviour regarding how a host is considered to be taken or not. This means that it is no longer about the namespace that the AuthConfigs belong to, but about whether the prevention of collisions is either enabled or disabled, according to the value of `PREVENT_COLLISIONS`. 

In other words, if `PREVENT_COLLISIONS=true` (default), Authorino will now reject linking AuthConfigs to hosts already taken _including when both resources are defined in the same namespace_.

To disable the prevention of host name collisions and let users freely replace each others' auth definitions by picking the same host names, whoever controls the Authorino instance can set `PREVENT_COLLISIONS` to `false`, and this behaviour will apply both in cluster-wide mode or namespaced mode.

### Verification steps

Create the cluster:

```sh
kind create cluster --name authorino
```

Deploy a custom image of the Authorino Operator based on [authorino-operator:disable-host-collision-prevention](https://github.com/Kuadrant/authorino-operator/tree/disable-host-collision-prevention):

```sh
cd $GOPATH/src/github.com/kuadrant/authorino-operator && git fetch && git checkout disable-host-collision-prevention
docker build -t authorino-operator:local .
kind load docker-image authorino-operator:local --name authorino
sed -e 's/quay.io\/kuadrant\/authorino-operator:latest/authorino-operator:local/' config/deploy/manifests.yaml | kubectl apply -f -
```

Build and push an image of Authorino based on this branch:

```sh
cd $GOPATH/src/github.com/kuadrant//authorino && git fetch && git checkout disable-host-collision-prevention
docker build -t authorino:local .
kind load docker-image authorino:local --name authorino
```

Prepare the environment and request an instance of Authorino:

```sh
make cert-manager namespace certs user-apps
kubectl wait --timeout=300s --for=condition=Available deployments --all
kubectl port-forward deployment/envoy 8000:8000 &

kubectl apply -f -<<EOF
apiVersion: operator.authorino.kuadrant.io/v1beta1
kind: Authorino
metadata:
  name: authorino
spec:
  clusterWide: true # we will be trying in cluster-wide mode, but you can change this to 'false' and try it for namespaced as well
  preventCollisions: false # we will let users freely replace each others' AuthConfigs
  image: authorino:local
  imagePullPolicy: IfNotPresent
  listener:
    tls:
      certSecretRef:
        name: authorino-server-cert
  oidcServer:
    tls:
      certSecretRef:
        name: authorino-oidc-server-cert
EOF
```

Create a few namespaces (if indeed verifying in cluster-wide mode):

```sh
kubectl create namespace user-1
kubectl create namespace user-2
kubectl create namespace user-3
kubectl create namespace user-4
```

Start creating AuthConfigs with conflicting host names:

```sh
kubectl -n user-1 apply -f -<<EOF
kind: AuthConfig
metadata:
  name: auth-1
spec:
  hosts:
  response:
  - name: auth-data
    wrapperKey: X-Auth-Data
    json:
      properties:
      - name: config
        value: auth-1
EOF
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -i
# "X-Auth-Data": "{\"config\":\"auth-1\"}",
```

```sh
kubectl -n user-2 apply -f -<<EOF
kind: AuthConfig
metadata:
  name: auth-2
spec:
  hosts:
  response:
  - name: auth-data
    wrapperKey: X-Auth-Data
    json:
      properties:
      - name: config
        value: auth-2
EOF
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -i
# "X-Auth-Data": "{\"config\":\"auth-2\"}",
```

```sh
kubectl -n user-3 apply -f -<<EOF
kind: AuthConfig
metadata:
  name: auth-3
spec:
  hosts:
  - "*.127.0.0.1.nip.io"
  response:
  - name: auth-data
    wrapperKey: X-Auth-Data
    json:
      properties:
      - name: config
        value: auth-3
EOF
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000 -i
# "X-Auth-Data": "{\"config\":\"auth-2\"}",
```

```sh
curl http://other.127.0.0.1.nip.io:8000 -i
# "X-Auth-Data": "{\"config\":\"auth-3\"}",
```

```sh
kubectl -n user-4 apply -f -<<EOF
kind: AuthConfig
metadata:
  name: auth-4
spec:
  hosts:
  - "*.127.0.0.1.nip.io"
  response:
  - name: auth-data
    wrapperKey: X-Auth-Data
    json:
      properties:
      - name: config
        value: auth-4
EOF
```

```sh
curl http://other.127.0.0.1.nip.io:8000 -i
# "X-Auth-Data": "{\"config\":\"auth-4\"}",
```

### Upcoming issues to solve

**Problem 1: Outdated statuses**

By disabling the prevention of host name collision, users will have the power to replace each others' AuthConfigs for specific host names, including across k8s namespaces. Since Authorino does not update the statuses of preexisting AuthConfigs as part of reconciling a new one just applied, but just the status of the new AuthConfig itself, we will end up with a situation like the following after applying the 4 AuthConfigs from the verification steps above:

```sh
# NAMESPACE   NAME     READY   HOSTS   AUTHENTICATION   METADATA   AUTHORIZATION   RESPONSE   WRISTBAND
# user-1      auth-1   true    1/1     0                0          0               1          false
# user-2      auth-2   true    1/1     0                0          0               1          false
# user-3      auth-3   true    1/1     0                0          0               1          false
# user-4      auth-4   true    1/1     0                0          0               1          false
```

Notice that all resources are reported ready and fully available, when actually only `user-2/auth-2` and `user-4/auth-4` are active. Instead, Authorino should reflect the latest changes that allowed 2 AuthConfigs to claim the host names previously linked to `user-1/auth-1` and `user-3/auth-3` in the status subresource of these CRs.

This is a problem that currently already exists in namespaced mode, but it actually derives from when/how we check whether a host name is taken and which status sub-resources we update when that happens. Therefore, it is a problem that will from now on happen always when `PREVENT_COLLISIONS=false`.

<br/>

**Problem 2: Once again, the order of reconciliation strikes us**

The fix introduced in #330, for when an Authorino pod starts and finds preexisting AuthConfigs, and therefore it needs to populate the index, will not work when `PREVENT_COLLISIONS=false`.

Authorino ensures AuthConfigs and host names are added and linked together in the index following the chronological order that the resources were originally applied. It does that in a quick single-pass step, just to rebuild the structure of the index tree. Then, it gives control back to the usual reconciliation loop, which effectively parses each AuthConfig and finishes populating the index, this time leveraging concurrency.

In this second step, the order of the resources can no longer be guaranteed. Without the check on whether the reconciliation is an update for the same AuthConfig or an attempt by someone to replace some else’s AuthConfig by choosing the same host name (i.e. the prevention of host name collisions is _off_), Authorino can rebuild the index in the right order but it will end up finishing linking the host names to the AuthConfigs according to chance.

This is not a problem when `PREVENT_COLLISIONS=true`.

---

Closes #329


guicassolato:
It turns out disabling the host name collision mechanism can lead to more problems than the solutions it enables, such as the handling of the status updates of superseded AuthConfigs. So, until we have a valid use case that justifies reconsidering the amount the work necessary to address those problems, we're closing this PR, together with the issue that originated it.

PR #354, opened 2022-10-07
Upgrade to Golang v1.18.x

### LINKED ISSUE #353: Upgrade to Go v1.18+
Although Golang v1.19 has been out for about 2 months now, v1.18 is the latest stable version officially supported by Red Hat for productised images.

We can choose to bump straight to v1.19.x and leave v1.18.x only for internal (midstream) builds, as they should be compatible (see Golang's [promise of compatibility](https://tip.golang.org/doc/go1compat)). However, keeping upstream and midstream closer in sync tends to be a better, safer practice.

guicassolato:
Closes #353.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,23 @@
 # Build the authorino binary
-FROM registry.access.redhat.com/ubi8/go-toolset:1.17.10 as builder
+FROM registry.access.redhat.com/ubi9-minimal:latest AS builder
 USER root
+RUN microdnf install -y tar gzip
+RUN arch=""; \
+    case $(uname -m) in \
+      x86_64) arch="amd64";; \
+      aarch64) arch="arm64";; \
+    esac; \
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -1,14 +1,19 @@
 # Build the authorino binary
-FROM registry.access.redhat.com/ubi8/go-toolset:1.17.10 as builder
+FROM registry.access.redhat.com/ubi9-minimal:9.0.0 AS builder
 USER root
+ARG TARGETARCH
+RUN microdnf install -y tar gzip
+RUN curl -O -J "https://dl.google.com/go/go1.18.7.linux-${TARGETARCH}.tar.gz"; \
```
guicassolato, line 6: I'm afraid no tag has been published yet for v1.18 at `registry.access.redhat.com/ubi{8|9}/go-toolset`, so we need to install Golang manually.

eguzki (approved):

PR #365, opened 2022-11-18
Readiness probe

### LINKED ISSUE #355: Readiness Probe
### Challenge
When restarting authorino deployments (with 3 replicas and a pdb), for a few brief seconds, requests are denied with a 403. 

### Solution
Have a healtz endpoint that can be used as a readiness probe that is true when all of the auth configs have been loaded.

guicassolato:
Implements health and readiness probe endpoints for the controllers. ATM, the endpoints are trivial readiness and health check endpoints the by default always return "ok".

The aggregated status of all AuthConfigs watched by the controller is an opt-in check.

New endpoints introduced:
- `/healthy`: Health probe (ping)
- `/readyz`: Aggregated readiness probe (by default, none)
- `/readyz/authconfigs`: Aggregated status of all AuthConfigs watched by the controller

In general, all endpoints return either `200` ("ok") or `500` (when 1+ probes fail).

The default binding network address is `:8081`. It can be changed using a newly introduced flag (command-line arg) `--health-probe-addr`.

The following query string parameters are supported:
- `verbose=any`: more verbose response messages;
- `include=authconfigs`: to include the aggregated status of AuthConfigs in the response;
- `exclude=(check name)`: to exclude a particular probe - has no effect when passed on requests to `/readyz/authconfigs`.

Closes #355

### Verification steps

```sh
$ make local-setup FF=1
$ kubectl port-forward deployment/authorino 8081:8081 &
```

```sh
$ curl http://localhost:8081/healthz
ok
```

```sh
$ curl http://localhost:8081/readyz
ok
```

```sh
$ kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
EOF
authconfig.authorino.kuadrant.io/talker-api-protection created
```

```sh
$ curl http://localhost:8081/readyz
ok
```

```sh
$ curl "http://localhost:8081/readyz?verbose=true"
[+]authconfigs ok
healthz check passed
```

```sh
$ kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection-2
spec:
  hosts:
EOF
authconfig.authorino.kuadrant.io/talker-api-protection-2 created
```

```sh
$ curl http://localhost:8081/readyz
ok
```

```sh
$ curl "http://localhost:8081/readyz?include=authconfigs"
[-]authconfigs failed: reason withheld
healthz check failed
```

```sh
$ curl http://localhost:8081/readyz/authconfigs
internal server error: authconfig is not ready: default/talker-api-protection-2 (reason: HostsNotLinked)
```

### DIFF
--- a/controllers/status_report.go
+++ b/controllers/status_report.go
@@ -3,6 +3,8 @@ package controllers
 import (
 	"sync"
 	"time"
+
+	"github.com/kuadrant/authorino/pkg/utils"
 )
 
 func NewStatusReportMap() *StatusReportMap {
@@ -40,7 +42,7 @@ func (m *StatusReportMap) ReadAll() map[string]StatusReport {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 
-	return m.statuses
+	return utils.CopyMap(m.statuses)
 }
 
 func (m *StatusReportMap) Clear(id string) {
--- a/controllers/status_report.go
+++ b/controllers/status_report.go
@@ -36,6 +36,13 @@ func (m *StatusReportMap) Set(id, reason, message string, hosts []string) {
 	}
 }
 
+func (m *StatusReportMap) ReadAll() map[string]StatusReport {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return m.statuses
+}
+
 func (m *StatusReportMap) Clear(id string) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -1,0 +1,53 @@
+package health
+
+import (
+	"net/http"
+	"sync"
+)
+
+type Observable interface {
+	Ready() error
+}
+
+type HealthzHandler interface {
+	Observe(...Observable)
+	HandleReadyzCheck(*http.Request) error
+}
+
+type HandlerOption func(HealthzHandler)
+
+func Observe(observables ...Observable) HandlerOption {
+	return func(h HealthzHandler) {
+		h.Observe(observables...)
+	}
+}
+
+func NewHandler(options ...HandlerOption) HealthzHandler {
+	h := &handler{}
+	for _, o := range options {
+		o(h)
+	}
+	return h
+}
+
+type handler struct {
+	observables []Observable
+	mu          sync.RWMutex
+}
+
+func (h *handler) Observe(observables ...Observable) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.observables = append(h.observables, observables...)
+}
+
+func (h *handler) HandleReadyzCheck(_ *http.Request) error {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+	for _, reconciler := range h.observables {
+		if err := reconciler.Ready(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -1,12 +1,14 @@
 package health
 
 import (
+	"fmt"
 	"net/http"
+	"strings"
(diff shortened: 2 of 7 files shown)


guicassolato:
It turns out this approach can be improved to solve the issue for the readiness probe.

An AuthConfig being not ready (due to a host collision, for example) has nothing to do with the AuthConfig Reconciler (or even less with the Authorino instance!) not being ready. Exposing this endpoint through controller-runtime's readiness check mixes those two things together and introduces a risk of occasionally shutting the entire Authorino instance off because of a unexceptional situation of having one or more AuthConfigs not ready.

This is because one could think that the `/readyz` endpoint is suitable for configuring the [`readinessProbe`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes) of the Authorino Deployment, when in fact, as-is, it is not. Currently, this is not done by the Authorino Operator, but when it starts doing so, this endpoint cannot be used indiscriminately. To use this endpoint, the `authconfigs` check would have to be excluded by supplying in the request the `exclude=authconfigs` parameter.

guicassolato:
> This is because one could think that the /readyz endpoint is suitable for configuring the [readinessProbe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes) of the Authorino Deployment, when in fact, as-is, it is not.

No longer true.

I've changed so the `authconfigs` check is now opt-in. Therefore people can use the upper level (aggregated) `/readyz` endpoint to set readiness probe for the Authorino Deployment without risking the status of AuthConfig CRs to affect the state of the service itself.

One who wants to include the readiness state of the AuthConfigs in the overall aggregated readiness probe check can call `/readyz?include=authconfigs` (not suitable for setting readiness check for the Authorino Deployment).

Alternatively, only the aggregated status of the AuthConfig CRs watched by that particular Authorino instance can be checked by calling `/readyz/authconfigs`.

controllers/status_report.go
```diff
@@ -36,6 +36,13 @@ func (m *StatusReportMap) Set(id, reason, message string, hosts []string) {
 	}
 }
 
+func (m *StatusReportMap) ReadAll() map[string]StatusReport {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return m.statuses
```
eguzki, line 43: Shouldn't be a copy? returning a reference, you are exposing the shared data and can be updated externally potentially running into race conditions errors.

pkg/health/health.go
```diff
@@ -0,0 +1,69 @@
...
+		o(h)
+	}
+	return h
+}
+
+type handler struct {
+	name string
+
+	observables []Observable
+	mu          sync.RWMutex
```
eguzki, line 39: what shared (access) data is this MUTEX protecting?

pkg/utils/utils.go
```diff
@@ -30,3 +30,12 @@ func SubtractSlice(sl1, sl2 []string) []string {
 	}
 	return diff
 }
+
+func SliceContains[T comparable](s []T, val T) bool {
```
eguzki, line 34: now I have to learn Golang generics :partying_face:

pkg/health/health.go
guicassolato: The `observables` slice.

I usually try to group mutex and corresponding data structures it's intended to protect in one block of declarations, separated from the other (non related) data structures.

didierofrivia: I think that would be the `observables`

didierofrivia (commented):
Looks good to me! The only issue I see is the one reported by Eguz regarding returning the map of statuses

eguzki (approved):

Issue #163, opened 2021-10-06
Tracing


guicassolato:
As part of a chain of requests that by definition includes several other systems and services, the authorization service should have support for request tracing. That includes possibility to track and trace request tokens/IDs throughout the end-to-end steps and hops of the auth pipeline, either a token verification request that relies on an external auth server, ad-hoc auth metadata fetched, etc.

Rohith-Raju:
Hi, I'm new here 😄. Is this issue still relavent?

guicassolato:
Hi @Rohith-Raju. 

> Is this issue still relavent?

It is. Since it was opened, over a year ago, a few things changed; in a way that the issue deserves at least a better scope definition nevertheless.

Starting by what we still miss in Authorino, I'd say: proper embedded telemetry reporting, like by integrating OpenTracing. This is a relatively easy task to implement that could re-shape the description above.

What Authorino can do already, on the other hand, – and I should be more clear about it in the description as well – is:
1. it propagates a `request id` to the logs - this ID is typically generated by the API gateway and retrieved by Authorino (although there is at least one [exception](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#raw-http-authorization-interface) where the ID is generated by Authorino itself) and, provided you are exporting the logs to something like CloudWatch, it allows you to trace authorization workloads to the level of linking requests to corresponding responses and original HTTP contexts from the gateway or proxy - doc available [here](https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/observability.md#tracing);
2. with the newly introduced [`notify`](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#notifications-notify) phase of the AuthPipeline (#368, soon to be renamed as `callbacks` by #371), one can "trace" authorization requests by specifying one or more callback configs in the AuthConfig to HTTP endpoints of a tracing system - authentication options between Authorino and this external HTTP server are still limited, but if that's not an issue the requests themselves can contain fairly detailed information, such results of the evaluation of each authorization rule, combined with some user info, etc.

I hope this answers your question.

Rohith-Raju:
Thanks for the info. I'd like to work on this issue.
I'm aware of the working of authorino (by deploying a hello world scenario) also done setting up the project locally.
How do I start working on this issue?

guicassolato:
That's brilliant, @Rohith-Raju! We really appreciate your interest in contributing to Authorino.

[...]

PR #384, opened 2023-03-01
Tracing enhancements

guicassolato:
- [x] `observability-*` cmd-line flags renamed as `tracing-*` – so it's clear this is the only kind of signal being pushed, other types would require their own separate endpoints anyway
- [x] Tracing `seed` flag turned into a string array flag that accepts key=value pairs, renamed `tag`
- [x] Traceparent and Baggage propagators always enabled
- [x] Docs

---

Related to #163


docs/user-guides/observability.md
```diff
@@ -1407,6 +1407,24 @@ The examples below are all with `--log-level=debug` and `--log-mode=production`.
...
+
+Authorino propagates trace identifiers compatible with the W3C Trace Context format (https://www.w3.org/TR/trace-context/) and user-defined baggage data in the W3C Baggage format (https://www.w3.org/TR/baggage).
+
+### Log tracing
+
+Most log messages associated with an authorization request include the [`request id`](#request-id) value. This value can be used to match incomming request and corresponding outgoing response log messages, including at deep level when more fine-grained log details are enabled ([`debug` level level](#log-levels-and-log-modes)).
+
+### OpenTelemetry integration
+
+Integration with an OpenTelemetry collector can be enabled by supplying the `--tracing-service-endpoint` command-line flag (e.g. `authorino server --observability-service-endpoint=http://jaeger:14268/api/traces`).
```
eguzki, line 1426: the example uses old flag

docs/user-guides/observability.md
```diff
@@ -1407,6 +1407,24 @@ The examples below are all with `--log-level=debug` and `--log-mode=production`.
...
+
+The additional `--tracing-service-tags` command-line flag allow to specify fixed agent-level key-value tags for the trace signals emitted by Authorino (`key1=value1[,keyN=valueN]*`).
```
eguzki, line 1428: have you considered supporting multiple flags? That would leave the parsing logic to the CLI lib.

```
authorino server --tracing-service-tags a=1 --tracing-service-tags b=2
```

eguzki (changes requested):

docs/user-guides/observability.md
guicassolato: I have not, but it's a good idea. It feels more Golang flags.

Done.

PR #386, opened 2023-03-09
WIP: Add propagation to external services

### LINKED ISSUE #385: Propagate traces to external services
Make HTTP requests sent by Authorino to external services traceable, by injecting the headers (`traceparent` and `tracestate`, and `baggage`), from the context:
- External sources of metadata
- HTTP callbacks
- OPA policy registries
- etc

Related to #163.

---

Implementation:

```go
otel.GetTextMapPropagator().Inject(ctx, otel_propagation.HeaderCarrier(req.Header()))
```

Rohith-Raju:
Fixes #385 
cc @guicassolato 

Verification steps courtesy @guicassolato :

Step 1: Initialize cluster
```
make cluster install build
```

Step 2: Bring up authorino
```
bin/authorino server --tracing-service-endpoint http://localhost:14268/api/traces
```
Step 3: Apply auth-config
```
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  metadata:
  - name: echo
    http:
      endpoint: https://echo-api.3scale.net/
  response:
  - name: echo
    json:
      properties:
      - name: response
        valueFrom:
          authJSON: auth.metadata.echo
```
Step 4: Sending an external authorization request to a gRPC server using `grpcurl`
```
for repo in \
envoyproxy/data-plane-api \
googleapis/googleapis \
envoyproxy/protoc-gen-validate \
cncf/udpa
do
git clone git@github.com:$repo /tmp/authorino-grpc-client/$repo
done

brew install grpcurl

grpcurl -import-path /tmp/authorino-grpc-client/envoyproxy/data-plane-api  \
        -import-path /tmp/authorino-grpc-client/googleapis/googleapis \
        -import-path /tmp/authorino-grpc-client/envoyproxy/protoc-gen-validate \
        -import-path /tmp/authorino-grpc-client/cncf/udpa \
        -proto /tmp/authorino-grpc-client/envoyproxy/data-plane-api/envoy/service/auth/v3/external_auth.proto \
        -plaintext -d '{"attributes":{"request":{"http":{"host":"talker-api-authorino.127.0.0.1.nip.io"}}}}' \
        localhost:50051 \
        envoy.service.auth.v3.Authorization.Check
```
The response should look something like this
```
"status": {
    
  },
  "okResponse": {
    "headers": [
      {
        "header": {
          "key": "echo",
          "value": "{\"response\":{\"args\":\"\",\"body\":\"\",\"headers\":{\"CONTENT_TYPE\":\"text/plain\",\"HTTP_ACCEPT_ENCODING\":\"gzip\",\"HTTP_HOST\":\"echo-api.3scale.net\",\"HTTP_TRACEPARENT\":\"00-93b16549c5f367af4478a61dcdfc54e0-2a038657302c4690-01\",\"HTTP_USER_AGENT\":\"Go-http-client/1.1\",\"HTTP_VERSION\":\"HTTP/1.1\",\"HTTP_X_ENVOY_EXPECTED_RQ_TIMEOUT_MS\":\"15000\",\"HTTP_X_ENVOY_EXTERNAL_ADDRESS\":\"103.92.103.133\",\"HTTP_X_FORWARDED_FOR\":\"103.92.103.133\",\"HTTP_X_FORWARDED_PROTO\":\"https\",\"HTTP_X_REQUEST_ID\":\"be1eac9d-d925-4159-97c2-5aa235ee46bc\"},\"method\":\"GET\",\"path\":\"/\",\"uuid\":\"5e60b641-d7cd-496c-a594-ef30d5437467\"}}"
        }
      }
    ]
  },
  "dynamicMetadata": {
    }
} 
```

### DIFF
--- a/pkg/evaluators/authorization/opa.go
+++ b/pkg/evaluators/authorization/opa.go
@@ -17,6 +17,9 @@ import (
 
 	opaParser "github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/rego"
+
+	"go.opentelemetry.io/otel"
+	otel_propagation "go.opentelemetry.io/otel/propagation"
 )
 
 const (
@@ -207,6 +210,8 @@ func (ext *OPAExternalSource) downloadRegoDataFromUrl() (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	otel.GetTextMapPropagator().Inject(req.Context(), otel_propagation.HeaderCarrier(req.Header))
 
 	if resp, err := http.DefaultClient.Do(req); err != nil {
 		return "", fmt.Errorf("failed to fetch Rego config: %v", err)
--- a/pkg/evaluators/identity/oauth2.go
+++ b/pkg/evaluators/identity/oauth2.go
@@ -11,6 +11,9 @@ import (
 	"github.com/kuadrant/authorino/pkg/auth"
 	"github.com/kuadrant/authorino/pkg/context"
 	"github.com/kuadrant/authorino/pkg/log"
+
+	"go.opentelemetry.io/otel"
+	otel_propagation "go.opentelemetry.io/otel/propagation"
 )
 
 type OAuth2 struct {
@@ -67,6 +70,8 @@ func (oauth *OAuth2) Call(pipeline auth.AuthPipeline, ctx gocontext.Context) (in
 	}
 
 	log.FromContext(ctx).WithName("oauth2").V(1).Info("sending token introspection request", "url", tokenIntrospectionURL.String(), "data", encodedFormData)
+
+	otel.GetTextMapPropagator().Inject(ctx, otel_propagation.HeaderCarrier(req.Header))
 
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
--- a/pkg/evaluators/metadata/generic_http.go
+++ b/pkg/evaluators/metadata/generic_http.go
@@ -15,6 +15,9 @@ import (
 	"github.com/kuadrant/authorino/pkg/json"
 	"github.com/kuadrant/authorino/pkg/log"
 	"github.com/kuadrant/authorino/pkg/oauth2"
+
+	"go.opentelemetry.io/otel"
+	otel_propagation "go.opentelemetry.io/otel/propagation"
 )
 
 type GenericHttp struct {
@@ -128,6 +131,7 @@ func (h *GenericHttp) buildRequest(ctx gocontext.Context, endpoint, authJSON str
 	}
 
 	req.Header.Set("Content-Type", contentType)
+	otel.GetTextMapPropagator().Inject(ctx, otel_propagation.HeaderCarrier(req.Header))
 
 	if logger := log.FromContext(ctx).WithName("http").V(1); logger.Enabled() {
 		logData := []interface{}{
--- a/pkg/evaluators/metadata/uma.go
+++ b/pkg/evaluators/metadata/uma.go
@@ -13,6 +13,9 @@ import (
 	"github.com/kuadrant/authorino/pkg/context"
 	"github.com/kuadrant/authorino/pkg/json"
 	"github.com/kuadrant/authorino/pkg/log"
+
+	"go.opentelemetry.io/otel"
+	otel_propagation "go.opentelemetry.io/otel/propagation"
 )
 
 type providerJSON struct {
@@ -132,6 +135,7 @@ func (pat *PAT) Get(rawurl string, ctx gocontext.Context, v interface{}) error {
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Authorization", "Bearer "+pat.String())
 
+	otel.GetTextMapPropagator().Inject(ctx, otel_propagation.HeaderCarrier(req.Header))
 	// get the response
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
@@ -241,6 +245,7 @@ func (uma *UMA) requestPAT(ctx gocontext.Context, pat *PAT) error {
 
 	log.FromContext(ctx).V(1).Info("requesting pat", "url", tokenURL.String(), "data", encodedData, "headers", req.Header)
 
+	otel.GetTextMapPropagator().Inject(ctx, otel_propagation.HeaderCarrier(req.Header))
 	// get the response
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
--- a/pkg/evaluators/metadata/user_info.go
+++ b/pkg/evaluators/metadata/user_info.go
@@ -10,6 +10,9 @@ import (
 	"github.com/kuadrant/authorino/pkg/context"
 	"github.com/kuadrant/authorino/pkg/evaluators/identity"
 	"github.com/kuadrant/authorino/pkg/log"
+
+	"go.opentelemetry.io/otel"
+	otel_propagation "go.opentelemetry.io/otel/propagation"
 )
 
 type UserInfo struct {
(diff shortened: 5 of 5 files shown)


Rohith-Raju:
Working: 

In `main.go`  we first set the propagator 
```
otel.SetTextMapPropagator(otel_propagation.NewCompositeTextMapPropagator(otel_propagation.TraceContext{}, otel_propagation.Baggage{}))
```
Then  use the Getter to get `traceContext` and `Baggage` and inject them into the request headers  

```
otel.GetTextMapPropagator().Inject(ctx, otel_propagation.HeaderCarrier(req.Header))
```

Rohith-Raju:
@guicassolato Thoughts on testing this? We need to extract the information and add tracing to the external services, right?

guicassolato:
@Rohith-Raju,

I tested the changes today. In general, it's all right. The PR does what it's expected to do, mostly.

There are just a couple things that I didn't have time to investigate further but might be worth tackling.

1. Traces are only propagated into external service calls when a tracing provider is initialised, i.e. when `--tracing-service-endpoint` is provided; otherwise no trace is injected. I'm not sure if this is what we want. Maybe yes. At the same time, I can't help thinking that Authorino pushing telemetry to the trace provider is one thing and propagating to external services is another, and those things should be independent from each other to an extent. Like, maybe Authorino is not emitting traces, but the external service is. In this case, Authorino would break the chain altogether (instead of just being a jump between two external spans).
2. Baggage is not being propagated into external service calls at all. This one is more concerning. Maybe it's an issue with the instrumentation library for GRPC? Or maybe it's Envoy. Anyway, this needs investigation.

I'm pretty sure these are not this PR's fault. Nevertheless, it feels we could sync this and the fixes together, maybe?

At least about # 2 above we need to be certain before merging the PR, because users will probably expect baggage to the propagated as well.

Rohith-Raju:
> Traces are only propagated into external service calls when a tracing provider is initialised, i.e. when --tracing-service-endpoint is provided; otherwise no trace is injected. I'm not sure if this is what we want. Maybe yes. At the same time, I can't help thinking that Authorino pushing telemetry to the trace provider is one thing and propagating to external services is another, and those things should be independent from each other to an extent. Like, maybe Authorino is not emitting traces, but the external service is. In this case, Authorino would break the chain altogether (instead of just being a jump between two external spans).

@guicassolato, We could introduce another flag that takes care of this or set tracing to be _turned on_ by default unless disabled explicitly or some other way. But this could take a toll on user experience. WDYT

> At least about # 2 above we need to be certain before merging the PR, because users will probably expect baggage to the propagated as well.

Yes, let's do it. 😄

guicassolato:
> We could introduce another flag that takes care of this or set tracing to be `turned on` by default unless disabled explicitly or some other way. But this could take a toll on user experience. WDYT

IDK if I follow 100%, but if introducing new flags + enabling tracing by default is our only option here, I prefer not to propagate traces into external service calls. Let's leave it for only when Authorino is reporting traces and [this line](https://github.com/Kuadrant/authorino/blob/919be301b25960ed5a44e112405eb83f222796d6/main.go#L193) goes back inside of the `if`.

Rohith-Raju:
>Baggage is not being propagated into external service calls at all. This one is more concerning. Maybe it's an issue with the instrumentation library for GRPC? Or maybe it's Envoy. Anyway, this needs investigation.

@guicassolato I manually added baggage by 
```
bagMember, err := baggage.NewMember("baggageKey", "bagevalue")
	if err != nil {
		return nil, err
	}
	bag, err := baggage.New(bagMember)
	if err != nil {
		return nil, err
	}
	ctx = baggage.ContextWithBaggage(ctx, bag)
	otel.GetTextMapPropagator().Inject(ctx, otel_propagation.HeaderCarrier(req.Header))
```
And the Response 

```
{
  "status": {
    
  },
  "okResponse": {
    "headers": [
      {
        "header": {
          "key": "echo",
          "value": "{\"response\":{\"args\":\"\",\"body\":\"\",\"headers\":{\"CONTENT_TYPE\":\"text/plain\",\"HTTP_ACCEPT_ENCODING\":\"gzip\",\"HTTP_BAGGAGE\":\"baggageKey=bagevalue\",\"HTTP_HOST\":\"echo-api.3scale.net\",\"HTTP_TRACEPARENT\":\"00-739d0ae53ea0feac8c84ce0908e02f21-24d6a23e876f319a-01\",\"HTTP_USER_AGENT\":\"Go-http-client/1.1\",\"HTTP_VERSION\":\"HTTP/1.1\",\"HTTP_X_ENVOY_EXPECTED_RQ_TIMEOUT_MS\":\"15000\",\"HTTP_X_ENVOY_EXTERNAL_ADDRESS\":\"103.92.103.133\",\"HTTP_X_FORWARDED_FOR\":\"103.92.103.133\",\"HTTP_X_FORWARDED_PROTO\":\"https\",\"HTTP_X_REQUEST_ID\":\"b193b8df-133d-4dbf-88ff-314581fbebb9\"},\"method\":\"GET\",\"path\":\"/\",\"uuid\":\"6240a839-a9a1-4108-af4a-02847480b230\"}}"
        }
      }
    ]
  },
  "dynamicMetadata": {
    }
}
```
Baggage is being propagated 🎉
```
"HTTP_BAGGAGE\":\"baggageKey=bagevalue\"
```
Maybe the `baggage` is not populated in the previous scenario? WDYT

guicassolato:
> @guicassolato If your concern was whether we could propagate `baggage` data? Then worry no more 😄 . I manually added baggage by

@Rohith-Raju, we need to make sure the instrumentation<sup>1,2,3</sup> is taking care of extracting the `Baggage` data automatically and injecting it into the request to the external service. If it's not, then likely the problem is in the extraction part.

<sup>1</sup> https://github.com/Kuadrant/authorino/blob/a6cc9063eee73bd619667fda71f8a559593c7894/main.go#L193
<sup>2</sup> https://github.com/Kuadrant/authorino/blob/a6cc9063eee73bd619667fda71f8a559593c7894/main.go#L311-L312
<sup>3</sup> https://github.com/Kuadrant/authorino/blob/a6cc9063eee73bd619667fda71f8a559593c7894/main.go#L369

Rohith-Raju:
Sure, that makes sense.

guicassolato (approved):
@Rohith-Raju ,

I did some debugging now focusing on what's propagated by Envoy versus how instrumentation in Authorino will both extract the tracing data from the incoming requests and inject them into the outgoing ones.

_TL;DR –_ Envoy does not inject baggage data into the requests to the external authorization service. It does inject the traceparent, but not the baggage. Both are propagated to a routed upstream service, but unfortunately only the traceparent to the external authorization.

There's a section specifically about baggage data in the Envoy [docs](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/observability/tracing#baggage) where it says it works only for selected trace providers. However, I tried with [Jaeger](https://www.jaegertracing.io/) (where baggage should not be propagated) and with [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/) (where I expected it to be propagated), but it wasn't propagated in either.

Another important finding was that tracing does not have to be enabled in Authorino for traceparents and baggage data detected in the incoming authorization requests to be injected into the outgoing ones.

I can share the steps to reproduce if you're interested.

Anyway, the bottom line is we can merge this PR. Thanks again for the hard work!

Rohith-Raju:
@guicassolato, Thanks for merging the Pr
> Another important finding was that tracing does not have to be enabled in Authorino for traceparents and baggage data detected in the incoming authorization requests to be injected into the outgoing ones.

Sure, I'd love to. I'm caught up this week with my college exams. will take it up next week

PR #399, opened 2023-05-26
Control over overwriting values in the identity object

### LINKED ISSUE #394: Enhanced TokenReview identity object
When Authorino performs a [TokenReview](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#kubernetes-tokenreview-identitykubernetes), the structure of the extracted identity object varies depending whether the token is a JWT (e.g. a Kubernetes SA token) or an opaque token (e.g. OpenShift OAuth token). While for JWTs, the identity object is the JWT payload, for opaque tokens, it is the `status.user` from the TokenReview response.

This behaviour makes it hard to write authorisation rules based on attributes of the identity object, especially in hybrid authentication systems where both kinds of tokens are simultaneously accepted. Differences such as the name of the property that carries the subject ID (i.e., either `sub` or `username`) and presence/absence of relevant properties (e.g. `iss`, `groups`) can be hard to workaround, often requiring extra requests to the cluster authentication system to retrieve further user info – when such endpoint is available! –, and/or repetition of configuration under `when` conditions (only to control the expected identity object format).

Authorino [extended properties](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#extra-identity-extension-extendedproperties) feature doesn't play nice to normalise tokens in such cases where the source value of new (extension) property may not be present.

To exemplify the issue...

The payload of a Kubernetes JWT issued for a SA looks like the following:

```json
{
  "aud": [
    "https://kubernetes.default.svc"
  ],
  "exp": 1684835538,
  "iat": 1684834938,
  "iss": "https://kubernetes.default.svc",
  "kubernetes.io": {
    "namespace": "talker-api",
    "serviceaccount": {
      "name": "talker-api-bff",
      "uid": "bd2d95ad-5ca8-42fc-a6f8-e11600094b44"
    }
  },
  "nbf": 1684834938,
  "sub": "system:serviceaccount:talker-api:talker-api-bff"
}
```

The `status` subresource returned with the response to the TokenRequest request in an OpenShift cluster looks like the following:

```json
{
  "audiences": [
    "https://kubernetes.default.svc"
  ],
  "authenticated": true,
  "user": {
    "groups": [
      "system:serviceaccounts",
      "system:serviceaccounts:talker-api",
      "system:authenticated"
    ],
    "uid": "bd2d95ad-5ca8-42fc-a6f8-e11600094b44",
    "username": "system:serviceaccount:talker-api:talker-api-bff"
  }
}
```

- While JWT claims such as `iss` (from the SA JWT) as well as user info such as `user.groups` (from the TokenReview status) can both be useful for writing authorisation rules; nevertheless either only one or the other may be present.
- Writing an authorisation check (e.g. SubjectAccessReview) with user ID fetched from either `auth.identity.sub` or `auth.identity.username` is not straightforward.

<br/>

The user should have a way to tell if the JWT payload or the returned user info is desired to fill the identity object. By default, I'd personally go with the user info, but only if the keeping of other attributes, such as `audiences`, was possible (not the case today).

A few alternatives considered:

1. Both objects (JWT and TokenRequest status) could be merged together when the token is a JWT, such as in a structure like the following:
    ```go
    type KubeTokenIdentity struct {
      Token struct {
        Opaque string `json:"opaque,omitempty"`
        JWT    any    `json:"jwt,omitempty"`
      } `json:"token"`

      Status kubeauthnv1.TokenReviewStatus `json:"status"` // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#tokenreviewstatus-v1-authentication-k8s-io
    }
    ```
2. The API for the TokenReview identity verification method in the AuthConfig could include a field to use the JWT payload:
    ```yaml
    identity:
    - name: k8s-tokens
      kubernetes:
        jwt: true # default: false
    ```
3. Possibility to extend the identity object conditionally, e.g. by setting a (default) value only in case the property is absent in the source object – _only for the issue of `sub` vs `username`_:
    ```yaml
    identity:
    - name: k8s-tokens
      kubernetes: {}
      extendedProperties:
      - name: username
        valueFrom:
          authJSON: auth.identity.sub.@default:{"value":auth.identity.username}
    ```

### LINKED ISSUE #395: Do not overwrite identity extended properties
RFE: option to decide when to overwrite an existing extended property from an identity object.

Currently, Authorino does not check on the existence of the property in the identity object and simply extends the object with the new value, occasionally overwriting an existing one. 

**Example of use case where this is problematic:**

Given the following identity config:

```yaml
identity:
- name: k8s-tokens
  kubernetes: {}
  extendedProperties:
  - name: username
    valueFrom:
      authJSON: auth.identity.sub
```

For Kubernetes opaque tokens whose resolved identity object equals the `status.user` info from the TokenReview response, the `username` claim will be replaced with a `null` value.

This issue was described as well as part of #394.

**API implementation:**

```yaml
identity:
- name: k8s-tokens
  kubernetes: {}
  extendedProperties:
  - name: username
    valueFrom:
      authJSON: auth.identity.sub
    overwrite: true # default: false
```

guicassolato:
Adds new property `overwrite: bool` (default: `false`) to the extended properties.

The default set to `false` is a breaking change, since overwriting is right now the only option.

## Verification steps

```sh
make local-setup
```

Create the AuthConfig:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  identity:
  - name: plain
    plain:
      authJSON: context.request.http.headers.x-user|@fromstr
    extendedProperties:
    - name: name
      value: bar
      overwrite: false
    - name: age
      value: 35
      overwrite: true
    - name: groups
      value: ["admin"]
  response:
  - name: x-auth-data
    json:
      properties:
      - name: identity
        valueFrom:
          authJSON: auth.identity
EOF
```

Send a request:

```sh
curl -H 'x-user: {"name":"foo","age":30}' http://talker-api-authorino.127.0.0.1.nip.io:8000 -i
# …
#     "X-Auth-Data": "{\"identity\":{\"age\":35,\"groups\":[\"admin\"],\"name\":\"foo\"}}",
```

---

Closes #395.


eguzki:
Verification steps working as expected

eguzki (approved):

PR #406, opened 2023-06-12
test: generate test CA and client certs via code

guicassolato:
Modify the tests to generate the CA and client certs via code. This will avoid running again into issue such as the one fixed by #405


guicassolato:
Thanks to @pehal, I've also checked https://github.com/cloudflare/cfssl.

It seems to be a good option for CLI automation. To import it as a Go package into another code base, however, I’m afraid it won’t offer much gain compared to Go’s standard lib.

Happy to be convinced otherwise if any of you have a different opinion.

PR #407, opened 2023-06-12
Check 'client auth' key usage on mtls identity

guicassolato:
Make the mTLS identity verification method to check for 'Client authentication' extended key usage, instead of default 'Server authentication' ([ref](https://access.redhat.com/documentation/en-us/red_hat_certificate_system/9/html/administration_guide/standard_x.509_v3_certificate_extensions#Standard_X.509_v3_Certificate_Extensions-extKeyUsage)).

### DIFF
--- a/docs/user-guides/mtls-authentication.md
+++ b/docs/user-guides/mtls-authentication.md
@@ -108,7 +108,14 @@ kubectl apply -f https://raw.githubusercontent.com/kuadrant/authorino-examples/m
 Create a CA (Certificate Authority) certificate to issue the client certificates that will be used to authenticate clients that send requests to the Talker API:
 
 ```sh
-openssl req -x509 -sha256 -days 365 -nodes -newkey rsa:2048 -subj "/CN=talker-api-ca" -keyout /tmp/ca.key -out /tmp/ca.crt
+openssl req -x509 -sha256 -nodes \
+  -days 365 \
+  -newkey rsa:2048 \
+  -subj "/CN=talker-api-ca" \
+  -addext basicConstraints=CA:TRUE \
+  -addext keyUsage=digitalSignature,keyCertSign \
+  -keyout /tmp/ca.key \
+  -out /tmp/ca.crt
 ```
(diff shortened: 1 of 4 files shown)


docs/user-guides/mtls-authentication.md
```diff
@@ -118,6 +125,16 @@ kubectl create secret tls talker-api-ca --cert=/tmp/ca.crt --key=/tmp/ca.key
...
 ```
 
+Prepare an extension file for the client certificate signing requests:
+
+```sh
+cat > /tmp/x509v3.ext << EOF
+authorityKeyIdentifier=keyid,issuer
+basicConstraints=CA:FALSE
+keyUsage=digitalSignature,nonRepudiation,keyEncipherment,dataEncipherment
+extendedKeyUsage=clientAuth
```
KevFan, line 135:
```suggestion
```sh
cat > /tmp/x509v3.ext << EOF
authorityKeyIdentifier=keyid,issuer
basicConstraints=CA:FALSE
keyUsage=digitalSignature,nonRepudiation,keyEncipherment,dataEncipherment
extendedKeyUsage=clientAuth
EOF
```

KevFan (commented):
Code looks good, but following the mtls-authentication guide gives me an error with `certificate specifies an incompatible key usage` 🤔

guicassolato:
> Code looks good, but following the mtls-authentication guide gives me an error with `certificate specifies an incompatible key usage` 🤔

@KevFan, with which version of Authorino did you run the guide? The guide was written (including with the changes in this PR) assuming that it will work _after_ we merge the PR. This is because in the PR we're also changing the code that make the guide work of course.

To try the guide _before_ merging the PR, you should start with `make local-setup FF=1`, and then go straight to step ❹ of the guide.

Please let me know if this works for you.

KevFan (approved):
@guicassolato Oh, that was it 👍  Ran through the guide again and works as expected ! 🎖️ 

Looks good to me 👍

Issue #357, opened 2022-10-27
In and not-in operators for JSON pattern-matching


guicassolato:
It would be nice to have `in` and `not-in` operators amongst the ones supported by Authorino's [JSON pattern-matching](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#json-pattern-matching-authorization-rules-authorizationjson), so values fetched from the Authorization JSON can be checked if included/not included in a particular static list, without requiring an OPA policy for that.

This is more or less the opposite of the `incl` and `excl` operations, which check if a list fetched from the Authorization JSON includes/does not include a particular static value.

Rohith-Raju:
Hey @guicassolato, I'd like to work on this 😄

Rohith-Raju:
Hey @guicassolato, I thought about this and here's what I came up with. 

1. Add a new field in `StaticList []string` in `JSONPatternMatchingRule`.
```
type JSONPatternMatchingRule struct {
	Selector string
	Operator string
	Value    string
	StaticList []string
}
``` 
2. Add a new `switch` case `in` to check if values `fetched from the Authorization JSON` contain elements in the `StaticList`. Opposite can be done to `out`.
```
case operatorIn:
		exists := make(map[string]bool)
		for _, value := range expectedList {
			exists[value] = true
		}
		for _, value := range obtainedValue.Array() {
			if !exists[value.String()] {
				return false, nil
			}
		}
		return true, nil
```
WDYT?

guicassolato:
@Rohith-Raju,

We can add `StaticList []string` to `JSONPatternMatchingRule` but 2 things come to mind:

1. We still need to expose this option at the level of the API, i.e. https://github.com/Kuadrant/authorino/blob/ee10686d3f7c7553e7a4615cfa591d325d2ff8d2/api/v1beta1/auth_config_types.go#L170

   At the API, instead of introducing a new field, I thought about changing the type of `Value` to `runtime.RawExtension`. Depending on the operator, the value could be unmarshalled as `string` into `JSONPatternMatchingRule.Value` or as `[]string` into `JSONPatternMatchingRule.StaticList`.

2. Why just `string` and `[]string`, I wonder. Should we change the type of `JSONPatternMatchingRule.Value` to `any` and open for other JSON types as well?

<br/>

For the implementation (ignoring (2) above, thus thinking only `string` and `[]string`), I think `obtainedValue` would be a simple `string`, while `expectedValue` would be the slice. I.e.

```go
[...]

PR #427, opened 2023-09-20
feat: OR operator

### LINKED ISSUE #424: Logical disjunction (aka "OR" operator) in conditions
RFE: Add support for Logical disjunction (aka "OR" operator) in `when` conditions.

### Possible steps for implementation

- [x] Add 2 fields to the [`JSONPattern`](https://github.com/Kuadrant/authorino/blob/5fc204af85235548a6ba4e791c7ef86f08d3caa7/api/v1beta1/auth_config_types.go#L150-L153) (v1beta1) and [`PatternExpressionOrRef`](https://github.com/Kuadrant/authorino/blob/5fc204af85235548a6ba4e791c7ef86f08d3caa7/api/v1beta2/auth_config_types.go#L169-L172) (v1beta2) types:
  ```
  All []JSONPattern / []PatternExpressionOrRef `json:"all,omitempty"`
  Any []JSONPattern / []PatternExpressionOrRef `json:"any,omitempty"`
  ```
- [x] Modify all internal `Conditions` field of the evaluators to a new type that withholds the new logical operation `all` and `any`
- [x] Modify the [`*AuthPipeline. evaluateConditions`](https://github.com/Kuadrant/authorino/blob/5fc204af85235548a6ba4e791c7ef86f08d3caa7/pkg/service/auth_pipeline.go#L377-L387) function to support the new operations – recursion will be needed
- [x] (Optional) Define a new `pkg/conditions` package for everything related to the implementation of conditions (types, interfaces and functions)

guicassolato:
Adds support for logical disjunction (aka "OR" operator) in 'when' conditions and pattern-matching authorization rules.

Closes #424.

<br/>

Introduces a new field `any: Array<expression>` to the AuthConfig API to express logical disjunction, along with `all: Array<expression>` to enforce the AND operation (default if omitted). Where `expression` can be any of the following:
- a pattern composed of `selector`, `operator` and `value`;
- a reference to a named pattern (`patternRef: String`);
- a nested `any` field;
- a nested `all` field.

For backward compatibility, the AND operator is assumed for all lists of patterns not grouped under an explicit `any` field declaration. I.e., it is equivalent to grouping the patterns under a single `all` field declaration.

### Examples

**Example 1.** To allow anonymous access (no authentication required) for all requests to path `/test/*` OR method `GET` (ocasionally both):

```yaml
spec:
  authentication:
    anonymous-request:
      when:
      - any:
          operator: matches
          value: ^/test/.*
        - selector: context.request.http.method
          operator: eq
          method: GET
      anonymous: {}
```

**Example 2.** In a more complex condition, with nested logical operators, to express `host == 'foo.apis.io' AND ((path =~ '/test*' AND (method == 'POST' OR method == 'PUT')) OR method == 'GET')`:

```yaml
spec:
  authentication:
    anonymous-request:
      when:
      - selector: context.request.http.host
        operator: eq
        value: foo.apis.io
      - any:
        - all:
            operator: matches
            value: ^/test/.*
          - any:
            - selector: context.request.http.method
              operator: eq
              value: POST
            - selector: context.request.http.method
              operator: eq
              value: PUT
        - selector: context.request.http.method
          operator: eq
          value: GET
      anonymous: {}
```

#### Deploy

```sh
make local-setup FF=1
```

#### Try the new feature

①  Implicit AND (i.e. the existing syntax, which continues to be supported):

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  authorization:
    deny-specific-endpoints:
      when:
      - selector: context.request.http.method
        operator: eq
        value: POST
        operator: eq
        value: /foo
      opa:
        rego: "allow = false"
EOF
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X POST -i
# HTTP/1.1 403 Forbidden
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X GET -i
# HTTP/1.1 200 OK
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/bar -i
# HTTP/1.1 200 OK
```

②  Same as above with explicit AND operator (`all`):

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  authorization:
    deny-specific-endpoints:
      when:
      - all:
        - selector: context.request.http.method
          operator: eq
          value: POST
          operator: eq
          value: /foo
      opa:
        rego: "allow = false"
EOF
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X POST -i
# HTTP/1.1 403 Forbidden
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X GET -i
# HTTP/1.1 200 OK
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/bar -i
# HTTP/1.1 200 OK
```

③  Logical disjunction (`any`):

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  authorization:
    deny-specific-endpoints:
      when:
      - any:
        - selector: context.request.http.method
          operator: eq
          value: POST
        - selector: context.request.http.method
          operator: eq
          value: PUT
      opa:
        rego: "allow = false"
EOF
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X POST -i
# HTTP/1.1 403 Forbidden
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X GET -i
# HTTP/1.1 200 OK
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/bar -X PUT -i
# HTTP/1.1 403 Forbidden
```

④  Nested OR and AND expression:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  authorization:
    deny-specific-endpoints:
      when:
      - any:
        - all:
          - selector: context.request.http.method
            operator: eq
            value: POST
            operator: eq
            value: /foo
        - all:
          - selector: context.request.http.method
            operator: eq
            value: PUT
            operator: eq
            value: /foo
      opa:
        rego: "allow = false"
EOF
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X POST -i
# HTTP/1.1 403 Forbidden
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X PUT -i
# HTTP/1.1 403 Forbidden
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X GET -i
# HTTP/1.1 200 OK
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/bar -X PUT -i
# HTTP/1.1 200 OK
```

⑤  Same as above with implicit AND:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  authorization:
    deny-specific-endpoints:
      when:
        operator: eq
        value: /foo
      - any:
        - selector: context.request.http.method
          operator: eq
          value: POST
        - selector: context.request.http.method
          operator: eq
          value: PUT
      opa:
        rego: "allow = false"
EOF
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X POST -i
# HTTP/1.1 403 Forbidden
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X PUT -i
# HTTP/1.1 403 Forbidden
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X GET -i
# HTTP/1.1 200 OK
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/bar -X PUT -i
# HTTP/1.1 200 OK
```

⑥ Edge case: Invalid condition:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  authorization:
    foo:
      when:
      - all:
        - invalid: object
      opa:
        rego: "allow = false"
EOF
```

Because CRDs defined containing recursive types (e.g. an object/struct that defines a field whose type is the same type as the object itself) result in invalid OAS, the new fields `all` and `any` have to be marked with `x-kubernetes-preserve-unknown-fields`. Therefore, Kube API won't validate the scheme of these field, effectively accepting any array. The AuthConfig controller should not fail nonetheless, but just ignore the invalid condition instead. In the future, this can be improved with further validation.

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X POST -i
# HTTP/1.1 200 OK
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/foo -X PUT -i
# HTTP/1.1 200 OK
```

```sh
curl http://talker-api-authorino.127.0.0.1.nip.io:8000/bar -X PUT -i
# HTTP/1.1 200 OK
```


pkg/evaluators/authorization/json_test.go
```diff
@@ -251,33 +229,33 @@ func TestCall(t *testing.T) {
...
 			},
-			{
+			jsonexp.Pattern{
 				Selector: "auth.metadata.letters",
-				Operator: "incl",
+				Operator: jsonexp.IncludesOperator,
 				Value:    "c",
 			},
-			{
+			jsonexp.Pattern{
```
maleck13, line 253: These tests look good, but wondering is there an exsting or needed test around nested any and all patterns?

pkg/jsonexp/expressions.go
```diff
@@ -0,0 +1,178 @@
...
+			if expectedValue == item.String() {
+				return false, nil
+			}
+		}
+		return true, nil
+
+	case RegexOperator:
+		if re, err := regexp.Compile(expectedValue); err != nil {
+			return false, err
+		} else {
```
maleck13, line 89: prob can drop the else here?

pkg/jsonexp/expressions.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (o *Or) Matches(json string) (bool, error) {
+	if o.Left != nil {
+		left, err := o.Left.Matches(json)
+		if err != nil {
+			return false, err
+		}
+		if left {
+			return true, nil
```
maleck13, line 143: could this be simplified like the one below? ``` return left, nil``` and remove the if?

pkg/evaluators/authorization/json_test.go
maleck13: I might be missing someting, but I don't see test cases around `any` ??

maleck13: I see there are more test cases below in a different file

pkg/jsonexp/expressions_test.go
```diff
@@ -0,0 +1,551 @@
...
+			Operator: EqualOperator,
+			Value:    "wrong-value",
+		},
+	}
+	ok, err = exp.Matches(testJsonData)
+	assert.NilError(t, err)
+	assert.Check(t, !ok)
+}
+
+func TestEmptyOr(t *testing.T) {
```
maleck13, line 263: perhaps being stupid, but not getting why an empty `or` returns false but an empty `and` returns true?

maleck13 (commented):
The changes and test all look good to me, caveat I am not an expert on Authorino. One or two small questions but generally it is a looks good to me

pkg/jsonexp/expressions.go
```diff
@@ -0,0 +1,178 @@
...
+	if len(expressions) == 0 {
+		return &And{}
+	}
+	return &And{
+		Left:  expressions[0],
+		Right: All(expressions[1:]...),
+	}
+}
+
+func Any(expressions ...Expression) Expression {
```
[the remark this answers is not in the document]
alexsnaps: This has surprised more than one, but yes... See https://en.wikipedia.org/wiki/Vacuous_truth

pkg/jsonexp/expressions.go
guicassolato: It requires expanding the scope of `re` and `err`, but yes.

pkg/jsonexp/expressions.go
guicassolato: This is a safe-guard clause not to have to assert the right side of the expression, something we can only do if `left` is known to be true, but not otherwise.

maleck13 (dismissed):

docs/features.md
```diff
@@ -874,30 +874,100 @@ For the `AuthConfig` above,
...
+      value: ^/resources/.*
+```
+
+iii) OR condition (i.e., `any` block):
+
+```yaml
+spec:
+  when: # auth enforced only on requests with HTTP method equals to POST or PUT
+  - any:
+    - selector: context.request.http.method
```
maleck13, line 923: side point: Have you considered and `in` operator as this would (at least from an API perspective) improve the verbosity. 
```
 - selector: context.request.http.method
    operator: in
    values: ["PUT","POST"]
```

maleck13: I guess you could also do this with `regexp`?

docs/features.md
```diff
@@ -874,30 +874,100 @@ For the `AuthConfig` above,
...
 spec:
-  when: # no authn/authz required on requests to /status
+  when: # auth enforced only on requests to /resources/* path with method equals to POST or PUT
   - selector: context.request.http.path
-    operator: neq
-    value: /status
+    operator: matches
+    value: ^/resources/.*
+  - any:
+    - selector: context.request.http.method
```
maleck13, line 962: nit: is this the best example? Seems it could be achieved also with a `matches` operator? without too complex of a regex?

docs/features.md
```diff
@@ -956,47 +1046,61 @@ spec:
...
-      - selector: context.request.http.headers.authorization
-        operator: matches
-        value: APIKEY .+
-      apiKey: {…}
+      - all:
+        - patternRef: api-base-path
+        - any:
+          - selector: context.request.http.method
+            operator: eq
+            value: POST
```
maleck13, line 1082: just a note I do find myself wanting to write this as a matches operator :)

docs/features.md
```diff
@@ -956,47 +1046,61 @@ spec:
...
+          - selector: context.request.http.method
+            operator: eq
+            value: PUT
+          - selector: context.request.http.method
+            operator: eq
+            value: DELETE
+      opa:
+        patternMathing:
```
maleck13, line 1090:
```suggestion
        patternMatching:
```

docs/features.md
guicassolato: https://github.com/Kuadrant/authorino/issues/357

adam-cattermole (approved):
I've gone through the verification and code and the changes here look good and make sense from my point of view

maleck13 (approved):

docs/features.md
guicassolato: Indeed, many cases of disjunction can be solved with the `matches` operator, and this is definitely true for patterns limited to finite sets (and many other infinite ones as well!), such as the values of the HTTP method. For other cases, users may want/have to go with `any`.

So I wanted to present `any` also as an alternative to `matches`.

<details>
<summary>Click here if you want to learn more about the limitations of the <code>matches</code> operator.</summary>

Because Golang [`regexp`](https://pkg.go.dev/regexp) package (behind our `matches` operator) is a strict implementation of regular expressions, in the theoretical computer science sense of the term (i.e., the kind that defines a [regular language](https://en.wikipedia.org/wiki/Regular_language)), some patterns often supported by other programming languages are simply not available in Go by default.

For details, see https://github.com/google/re2/wiki/Syntax.
</details>

But that was not the only reason though. There are yet other cases where the disjunction involves patterns with different selectors. E.g.: _If (method == &lt;X&gt;) OR (path =~ &lt;Y&gt;.*)_.

I could have used any one of those more complex cases here. I was trying to reduce the cognitive complexity of the example by going with something simple and straightforward. I guess I didn't account for the more skilled readers, like you, who often knows a little bit more about the suject and therefore also other ways to solve the problem 🙂

PR #434, opened 2023-09-29
Enable/disable host name collision prevention for strict host subsets

### LINKED ISSUE #433: Enable/disable host name collision prevention
Authorino currently does not allow AuthConfigs to take control over the security scheme of hosts already protected by other AuthConfigs created previously. The behavior is described at [Avoiding host name collision](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#avoiding-host-name-collision) in the docs, and it is in general a good safe default.

There are use cases nonetheless where host ownership can be managed upfront by other means, such as by validating the AuthConfig CRs beforehand, at the level of the kube API requests (e.g. via a [validating webhook](https://youtu.be/JHRthOkTksg?si=VNWhBSaaovWlAdrL), or by enforcing more strict RBAC that constricts write permissions over the CRs to highly privileged cluster users only, amongst others), or when the host information and corresponding auth scheme association is known to be reconciled always from a known trusted source of truth. For such cases, disabling the prevention of host name collision mechanism may be desirable.

A concrete use case is [Kuadrant](https://docs.kuadrant.io/kuadrant-operator/doc/proposals/authpolicy-crd). The Kuadrant Operator manages Authorino AuthConfigs created out of its watched AuthPolicy CRs. Often, an AuthPolicy resource that targets a more generic network structure (e.g. a Kubernetes [Gateway](https://gateway-api.sigs.k8s.io/api-types/gateway/) resource, or a wildcard [HTTPRoute](https://gateway-api.sigs.k8s.io/api-types/httproute/)) is followed by a more specific one with a different auth scheme. Leveraging Authorino's ["most specific host wins" policy](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#host-lookup), both schemes can live side-by-side, but **only if the more specific AuthConfig is created first**. Otherwise, the more specific hosts in the AuthConfig created after will be rejected, causing this resource to be reported as not ready.

To enable such use cases, Authorino could put the mechanism that prevents host name collisions behind a switch.

The switch could be set at the level of the Authorino instance or for each individual AuthConfig, and would effectively allow an AuthConfig to take control over the auth scheme (i.e. to be "indexed") for hosts already protected by other AuthConfigs created previously.

Instead of a simple ON/OFF switch, it could be set at levels:
- Level 0 - No override
- Level 1 - Partial overrides only (hosts that are strict subsets of the ones already protected by other AuthConfigs)
- Level 2 - Full overrides (including identical hosts)

For level 2 in particular, one caveat to take into consideration is the status update of the resources that are fully overridden by other AuthConfigs created after. This is currently not a concern due to the assurance that the resource with the earliest creation timestamp always prevails, even in the cases of (re)building the index from preexisting AuthConfig stored in etcd. By allowing full overrides, some (older) AuthConfigs will be de-indexed. Their status stanzas should reflect that, including for the sake of reliable [readiness check](https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/observability.md#readiness-check).

guicassolato:
Adds a new command-line flag `--allow-superseding-host-subsets` that disables the [host name collision prevention](https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#avoiding-host-name-collision) for strict subsets of hosts attempted to be linked after a superset already taken.

Closes #433 under a more conservative approach where the switch for the host name collision mechanism does not come in levels, but as a simple ON/OFF switch, whose semantics is limited to strict subsets. I reckon this is a reasonable compromise as there are only few use cases I can think of for fully superseding the auth scheme a given host is linked to, based only on the order of creation of the resources.

Moreover, this approach exempts the reconciler from having to update the status of the (partially) superseded AuthConfigs, consistently with the current behaviour, when the order of creation goes: more specific first, then more generic host. The status of the more generic (wider) AuthConfig has never reflected that a subset of its hosts was actually under another auth scheme declared first.

### Verification steps

①  Setup

```sh
make local-setup FF=1
```

②  Activate the switch

Avoid the Authorino Operator to reconcile the changes back:

```sh
kubectl scale --replicas=0 deployment/authorino-operator -n authorino-operator
```

Modify the Authorino deployment activating the switch:

```sh
kubectl edit deployment/authorino
```

Add the `--allow-superseding-host-subsets` command-line flag to the list of `args` of the pod.

③  Create AuthConfigs in an order that otherwise would result in one of them being rejected

More generic:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: deny-all
spec:
  hosts:
  - "*.127.0.0.1.nip.io"
  authorization:
    deny-all:
      opa:
        rego: "allow = false"
EOF
```

More specific:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    friends:
      apiKey:
        selector:
          matchLabels:
            app: friends
        allNamespaces: true
      credentials:
        authorizationHeader:
          prefix: APIKEY
EOF
```

Verify both AuthConfigs have been accepted:

```sh
kubectl get authconfigs
# NAME         READY   HOSTS
# deny-all     true    1/1
# talker-api   true    1/1
```

④  Create another AuthConfig that tries to fully supersede a hostname already taken

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: allow-all
spec:
  hosts:
  - "*.127.0.0.1.nip.io"
EOF
```

Verify all the right AuthConfigs are accepted and rejected:

```sh
kubectl get authconfigs
# NAME         READY   HOSTS
# allow-all    false   0/1
# deny-all     true    1/1
# talker-api   true    1/1
```

⑤  Send requests to the sample API and confirm the right AuthConfigs are being enforced

Expose the proxy:

```sh
```

More specific auth scheme:

```sh
curl http://talker-api.127.0.0.1.nip.io:8000 -i
# HTTP/1.1 401 Unauthorized
```

More generic auth scheme:

```sh
curl http://other.127.0.0.1.nip.io:8000 -i
# HTTP/1.1 403 Forbidden
```

⑥  (Optional) Verify that bootstrapping the index with preexisting resources keeps the state consistsent

```sh
kubectl rollout restart deployment/authorino
```

```sh
kubectl get authconfigs
# NAME         READY   HOSTS
# allow-all    false   0/1
# deny-all     true    1/1
# talker-api   true    1/1
```

### DIFF
--- a/controllers/auth_config_controller.go
+++ b/controllers/auth_config_controller.go
@@ -58,12 +58,13 @@ const (
 // AuthConfigReconciler reconciles an AuthConfig object
 type AuthConfigReconciler struct {
 	client.Client
-	Logger        logr.Logger
-	Scheme        *runtime.Scheme
-	Index         index.Index
-	StatusReport  *StatusReportMap
-	LabelSelector labels.Selector
-	Namespace     string
+	Logger                      logr.Logger
+	Scheme                      *runtime.Scheme
+	Index                       index.Index
+	AllowSupersedingHostSubsets bool
+	StatusReport                *StatusReportMap
+	LabelSelector               labels.Selector
+	Namespace                   string
 
 	indexBootstrap sync.Mutex
 }
@@ -608,7 +609,7 @@ func (r *AuthConfigReconciler) addToIndex(ctx context.Context, resourceNamespace
 
 	for _, host := range hosts {
 		// check for host name collision between resources
-		if indexedResourceId, found := r.Index.FindId(host); found && indexedResourceId != resourceId {
+		if r.hostTaken(host, resourceId) {
 			looseHosts = append(looseHosts, host)
 			logger.Info("host already taken", "host", host)
 			continue
@@ -623,6 +624,15 @@ func (r *AuthConfigReconciler) addToIndex(ctx context.Context, resourceNamespace
 	}
 
 	return
+}
+
+func (r *AuthConfigReconciler) hostTaken(host, resourceId string) bool {
+	indexedResourceId, found := r.Index.FindId(host)
+	return found && indexedResourceId != resourceId && !r.supersedeHostSubset(host, indexedResourceId)
+}
+
+func (r *AuthConfigReconciler) supersedeHostSubset(host, supersetResourceId string) bool {
+	return r.AllowSupersedingHostSubsets && !utils.SliceContains(r.Index.FindKeys(supersetResourceId), host)
 }
 
 func (r *AuthConfigReconciler) bootstrapIndex(ctx context.Context) error {
--- a/controllers/auth_config_controller_test.go
+++ b/controllers/auth_config_controller_test.go
@@ -230,7 +230,7 @@ func TestTranslateAuthConfig(t *testing.T) {
 	// TODO
 }
 
-func TestPreventHostCollision(t *testing.T) {
+func TestPreventHostCollisionExactMatches(t *testing.T) {
 	mockController := gomock.NewController(t)
 	defer mockController.Finish()
 	indexMock := mock_index.NewMockIndex(mockController)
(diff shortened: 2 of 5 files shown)


controllers/auth_config_controller.go
```diff
@@ -608,7 +609,7 @@ func (r *AuthConfigReconciler) addToIndex(ctx context.Context, resourceNamespace
 
 	for _, host := range hosts {
 		// check for host name collision between resources
-		if indexedResourceId, found := r.Index.FindId(host); found && indexedResourceId != resourceId {
+		if indexedResourceId, found := r.Index.FindId(host); found && indexedResourceId != resourceId && (!r.AllowSupersedingHostSubsets || utils.SliceContains(r.Index.FindKeys(indexedResourceId), host)) {
```
alexsnaps, line 612: I'd maybe extract that one bool condition into a method of the reconciler... really a minor thing, but I find it makes it really hard to read when `&&` and `||` are mixed, and when there is a `!` in there just to screws with my brain ;)

alexsnaps (dismissed):
Really one minor comment...

controllers/auth_config_controller.go
guicassolato: Done.

alexsnaps (approved):

PR #436, opened 2023-10-16
fix: flaky tests

guicassolato:
- [x] authconfig conversion tests with random orderring of elements within array
- [x] weak concurrency coordination using sleep directive

Popped up more often after #435.


Boomatang (approved):
Changes look straight forward and good to me, as this is fixing flaky tests I guess time will tell if it resolves the issue.

PR #449, opened 2023-12-08
Add issues/PRs to new Project

alexsnaps:


.github/workflows/issues-workflow.yaml
```diff
@@ -4,6 +4,9 @@ on:
   issues:
     types:
       - opened
+  pull_request:
```
guicassolato, line 7: Triggering this both for issues and PRs may cause some duplicates in the board. I hope that's OK.

It is also true that sometimes we open PRs for which there are no tracking issues associated. I.e., the PRs _are_ the tracking issues themselves.

alexsnaps: Right, I prefer dups than blind spots.

guicassolato (approved):

PR #458, opened 2024-03-05
fix: extra metrics handler

guicassolato:
Options for the extra metrics handler of the controller manager need to be set before creating the manager.
Otherwise, the `/server-metrics` endpoint will not be included and return a 404.

### Verification steps

```sh
make local-setup FF=1
kubectl port-forward service/authorino-controller-metrics 8080:8080 2>&1 >/dev/null &
curl http://localhost:8080/server-metrics
```

You should see the server metrics and not a `404 Not Found`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -416,13 +416,13 @@ func setupTelemetryServices(opts telemetryOptions) {
 }
 
 func setupManager(options ctrl.Options) (ctrl.Manager, error) {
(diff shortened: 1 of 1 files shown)


adam-cattermole (approved):
lgtm, verified on main receive a `404 page not found`, whereas from these changes successfully retrieves the metrics.

PR #467, opened 2024-05-30
Added support for s390x & ppcl64e architecture

modassarrana89:
Updated Github workflow to support multi-arch


modassarrana89:
@jasonmadigan @thomasmaas @maleck13 @alexsnaps Please review & do the needful

guicassolato:
Thanks for the PR @modassarrana89.

I'm running the workflow now: https://github.com/Kuadrant/authorino/actions/runs/9303477739/job/25605860414

Let's see how it goes.

modassarrana89:
> Thanks for the PR @modassarrana89.
> 
> I'm running the workflow now: https://github.com/Kuadrant/authorino/actions/runs/9303477739/job/25605860414
> 
> Let's see how it goes.

Thanks. It has run successfully by the way.

.github/workflows/build-images.yaml
```diff
@@ -38,17 +38,21 @@ jobs:
...
-          sudo apt-get update
-          sudo apt-get install -y qemu-user-static
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.IMG_REGISTRY_USERNAME }}
+          password: ${{ secrets.IMG_REGISTRY_TOKEN }}
```
guicassolato, line 48: Any particular issue you hit that may have motivated adding this step?

modassarrana89: Didn't face any issue . I have added the login before we build/pushed the image. so that it should fail before build action get invoked

modassarrana89: @guicassolato Please let me know if i have to do any further changes. I am going to raise similar kind of changes for authorino operator as well

guicassolato: I see the point. Authentication is hardly the step we fail here though. I'm tempted to say we don't really need it.

I see some repetition with the parameters we have to pass to `redhat-actions/push-to-registry@v2` further down as well.

modassarrana89: ok. I will update it

modassarrana89: @guicassolato Removed it & updated the code changes

modassarrana89:
@guicassolato I have raised another PR with similar changes for authorino operator

guicassolato (approved):

Issue #471, opened 2024-06-12
Add commit hash dirty implementation, similar to kuadrantctl, to authorino. 


ehearneRedHat:
After speaking with @eguzki in [kuadrantctl #85](https://github.com/Kuadrant/kuadrantctl/pull/85) , the idea of adding a commit hash to the version output of authorino, similar to kuadrantctl, would be a great way for users to see where exactly their version of authorino is coming from. 

This can be done through a Makefile target, such as in `build` . Refer to the above issue to see how it was implemented in kuadrantctl for context. 

### Additional Context 

In addition to the already implemented version and hash being displayed, it would be great to show whether the version of authorino contains just the work on the main branch, or if there were additional changes made to the repository before being committed. 

Within kuadrantctl, when `./kuadrantctl version` is used, the output when there is additional changes made before being compiled, looks like

` kuadrantctl version: <version> (<git-hash>-dirty) `

A hash, while beneficial to the keen eye, can easily be overlooked, where as by showing the the branch in question has changes unknown to the user, they can go ahead and remove those changes so their binary is not dirty.

Therefore, within authorino, this additional feature, when the branch has new changes, can be implemented something like

```bash
$ ./authorino version
Authorino dc90ac837c3babc6afe6e1e5aacb64fe84c58088 (dirty)
```

Otherwise, it will remain the same if no new changes were made to the branch

```bash
$ ./authorino version
Authorino dc90ac837c3babc6afe6e1e5aacb64fe84c58088
```

guicassolato:
@ehearneRedHat, the Authorino binary already includes the version, which is either a SHA (latest builds) or a semantic version (releases.)

Apart from printed to the logs when the auth server starts:

```
{"level":"info","ts":"2024-06-12T13:48:57Z","logger":"authorino","msg":"booting up authorino","version":"dc90ac837c3babc6afe6e1e5aacb64fe84c58088","cmd":"server"}
```

one can also check it by running:

```
$ ./authorino version
Authorino dc90ac837c3babc6afe6e1e5aacb64fe84c58088
```

ehearneRedHat:
@guicassolato I was speaking with @eguzki again, and I was made aware that additional context may be needed to understand what the proposed feature was. 

I have added that context to the first comment. Thank you for the additional information. :)

PR #473, opened 2024-07-01
add commit hash implementation similar to kuadrantctl to authorino

ehearneRedHat:
### What:

- Makefile now contains implementation of git commit hash dirty implementation.
- Dockerfile has fallback option for when git is not installed such as in a container running the image.
- build arg changed in build images action to reflect implementation.
- New `build.yaml` file to add build information for authorino compile. Currently just supports `version` . 
- RELEASE docs have been updated with instructions on how to push a release . 

### Verify:

**NOTE:** for running the binaries use `authorino version` to check implementation. Within container in `default` namespace, check the logs at the top for the version. 

1. Run `make build`  and `make local-setup` locally. Without any file changes made the hash should be displayed without appendixed `dirty` flag. Then add a comment to a file to have git detect new changes and re-run the commands checking binary and in-cluster for both respectively.
2. Run the `build-images.yaml` workflow and check that the binaries show the hash.


.gitignore
```diff
@@ -25,3 +25,8 @@ vendor
 tmp
 target
 .scannerwork
+api/v1beta1/zz_generated.deepcopy.go
+api/v1beta2/zz_generated.deepcopy.go
+install/manifests.yaml
+install/crd/authorino.kuadrant.io_authconfigs.yaml
+install/rbac/role.yaml
```
guicassolato, line 32: I don't think these should be here.

.github/workflows/build-images.yaml
```diff
@@ -48,7 +40,7 @@ jobs:
           tags: ${{ env.IMG_TAGS }}
           platforms: linux/amd64,linux/arm64,linux/s390x,linux/ppc64le
           build-args: |
-            version=${{ env.VERSION }}
+            GITHUB_SHA=${{ github.sha }}
```
guicassolato, line 43: Why "GITHUB_SHA"? It should be "GIT_SHA", no?

.github/workflows/build-images.yaml
```diff
@@ -30,14 +30,6 @@ jobs:
         id: add-branch-tag
         run: |
           echo "IMG_TAGS=${GITHUB_REF_NAME/\//-} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Set Authorino version
-        id: authorino-version
-        run: |
-          if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
```
guicassolato, line 37: So users who run `./authorino version` will no longer see a semantic version not even on official releases? Not sure if I like this.

What about aiming for something more structured? E.g.

```json
{"version":"latest","commit":"abcd1234","dirty":true}
```

```json
{"version":"v0.17.0","commit":"efgh4567","dirty":false}
```

.gitignore
ehearneRedHat: Makes sense.

.github/workflows/build-images.yaml
ehearneRedHat: That makes total sense, Gui. I will aim to resolve this.

.github/workflows/build-images.yaml
ehearneRedHat: I can see why this would make sense, Gui. If it was a local deployment, `git` is how version control is managed, and even if on GitHub, the version control is also done with git. 

The reason why I wrote the variable as `GITHUB_SHA` was simply a foobar mistake. :) I will also make changes to have the name reflect its context. :)

main.go
```diff
@@ -73,6 +73,7 @@ const (
 var (
 	// ldflags
 	version string
+	dirty   string
```
eguzki, line 76: you would need yet another variable called `gitSHA`

main.go
```diff
@@ -373,7 +374,7 @@ func runWebhookServer(cmd *cobra.Command, _ []string) {
 func setup(cmd *cobra.Command, log logOptions, telemetry telemetryOptions) {
 	setupLogger(log)
 
-	logger.Info("booting up authorino", "version", version, "cmd", cmd.Use)
```
eguzki, line 376: expected format: `{"version":"v0.17.0","commit":"efgh4567","dirty":false}`

main.go
```diff
@@ -562,5 +563,9 @@ func timeoutMs(timeout int) time.Duration {
 }
 
 func printVersion(_ *cobra.Command, _ []string) {
-	fmt.Println("Authorino", version)
+	if dirty == "true" {
```
eguzki, line 566: expected format

```
<version> (<git-hash>[-dirty])
```

From https://github.com/Kuadrant/authorino/issues/471#issuecomment-2165087353

eguzki (changes requested):

main.go
eguzki, line 76:
```suggestion
	dirty   bool
```

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -123,7 +122,8 @@ func (in *AuthConfigSpec) DeepCopyInto(out *AuthConfigSpec) {
 			if val == nil {
 				(*out)[key] = nil
 			} else {
-				in, out := &val, &outVal
+				inVal := (*in)[key]
```
eguzki, line 125: these changes should not be here. Checkout go version. It should be 1.21.

Then, delete `bin` directory and run `make generate`. There should not be any change in this generated files.

main.go
ehearneRedHat: The reason why this is set to `string` is that ldflags cannot take bool .

main.go
```diff
@@ -562,5 +564,9 @@ func timeoutMs(timeout int) time.Duration {
...
+		fmt.Printf("%s (%s-dirty)", version, gitSHA)
```
eguzki, line 568: missing newline?

api/v1beta1/zz_generated.deepcopy.go
eguzki: I still see the changes

Makefile
```diff
@@ -1,8 +1,30 @@
...
+		else \
+			DIRTY="false"; \
+		fi; \
+	fi; \
+	echo $$DIRTY
+endef
+
 # Authorino version
-VERSION = $(shell git rev-parse HEAD)
+AUTHORINO_VERSION = v0.17.3-dev # change as version increments
```
eguzki, line 21: I would keep versioning at golang level

wdyt @guicassolato ?

Makefile
```diff
@@ -1,8 +1,30 @@
 # Use bash as shell
 SHELL = /bin/bash
 
+# Check for dirty :D 
+define check_dirty
```
eguzki, line 5: why not create `check-git-dirty.sh` like in https://github.com/Kuadrant/limitador-operator/blob/main/utils/check-git-dirty.sh?

Makefile
```diff
@@ -1,8 +1,30 @@
...
+	fi; \
+	echo $$DIRTY
+endef
+
 # Authorino version
-VERSION = $(shell git rev-parse HEAD)
+AUTHORINO_VERSION = v0.17.3-dev # change as version increments
+
+# Authorino commit hash
+GIT_SHA = $(shell git rev-parse HEAD)
```
eguzki, line 24: this is run for every make target, but only used in few of them. Check out https://github.com/Kuadrant/limitador-operator/blob/main/Makefile#L270-L275

Makefile
```diff
@@ -1,8 +1,30 @@
...
+
+# Authorino dirty
+DIRTY = $(shell $(check_dirty))
```
eguzki, line 27: same as above

main.go
```diff
@@ -373,7 +375,7 @@ func runWebhookServer(cmd *cobra.Command, _ []string) {
 func setup(cmd *cobra.Command, log logOptions, telemetry telemetryOptions) {
 	setupLogger(log)
 
-	logger.Info("booting up authorino", "version", version, "cmd", cmd.Use)
+	logger.Info("", "version", version, "commit", gitSHA, "dirty", dirty)
```
eguzki, line 378: I would not leave that empty. `authorino` is not needed, but something else?

main.go
```diff
@@ -562,5 +564,9 @@ func timeoutMs(timeout int) time.Duration {
 }
 
 func printVersion(_ *cobra.Command, _ []string) {
-	fmt.Println("Authorino", version)
+	if dirty == "true" {
+		fmt.Printf("%s (%s-dirty)\n", version, gitSHA)
```
eguzki, line 568:
```suggestion
		fmt.Printf("Authorino %s (%s-dirty)\n", version, gitSHA)
```

main.go
```diff
@@ -562,5 +564,9 @@ func timeoutMs(timeout int) time.Duration {
...
+	} else {
+		fmt.Printf("%s (%s)\n", version, gitSHA)
```
eguzki, line 570:
```suggestion
		fmt.Printf("Authorino %s (%s)\n", version, gitSHA)
```

eguzki (changes requested):

Makefile
ehearneRedHat: I'll hold on implementing here until Gui gets back to us :)

Makefile
ehearneRedHat: I think that's a good suggestion, @eguzki ! Will seek to implement.

main.go
guicassolato: `cmd` must stay too.

Makefile
ehearneRedHat: Absolutely, @eguzki ! Target level implementation is the way to go. :)

Makefile
guicassolato: I'm OK with "X.Y.Z" version, i.e. without leading "v".

Be careful with other other possible uses of this variable, like in the tagging of the Docker image, which must continue to follow the pattern `vX.Y.Z`.

I'm not sure about hard coding the version inside the Makefile. We didn't need that before. It was either a SHA for latest or an actual semantic version number (from the name of the branch/tag.)

Also, no need to rename the variable `AUTHORINO_VERSION` IMO. If it's inside Authorino's Makefile, `VERSION` must refer to the Authorino version.

main.go
ehearneRedHat: I can add `"build information"` ?

main.go
ehearneRedHat: Makes sense.

main.go
guicassolato: > I can add `"build information"` ?

If you're referring to "cmd", it's not that. cmd is if the binary is being execute as `server`, or `webhooks`, etc

Makefile
ehearneRedHat: @guicassolato all good suggestions, and thank you for the insight on the docker tagging. 

> I'm OK with "X.Y.Z" version, i.e. without leading "v".

OK, makes sense, for semantics. I'll move the v down to the string that displays the version.

> I'm not sure about hard coding the version inside the Makefile. We didn't need that before. It was either a SHA for latest or an actual semantic version number (from the name of the branch/tag.)

OK sure, I'll move `VERSION` back to global level and change to latest.

main.go
ehearneRedHat: @guicassolato let me know if this is what you were looking for.

ehearneRedHat: @guicassolato just checking in if the new changes satisfy your comments ?

Makefile
```diff
@@ -112,22 +112,31 @@ manifests: controller-gen kustomize ## Generates the manifests in $PROJECT_DIR/i
...
+VERSION ?= $(GIT_SHA)
 using_semantic_version := $(shell [[ $(VERSION) =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.+)?$$ ]] && echo "true")
 ifdef using_semantic_version
 IMAGE_TAG=v$(VERSION)
 else
 IMAGE_TAG=local
 endif
 AUTHORINO_IMAGE ?= $(IMAGE_REPO):$(IMAGE_TAG)
+
+docker-build: AUTHORINO_VERSION=v0.17.3-dev # change as version increments
```
guicassolato, line 135: This is going to be very easy to overlook and forget, especially at line 135. I wonder if we should read it from a file [`build.yaml`](https://github.com/Kuadrant/authorino-operator/blob/7e2fb1b5a7ce863d417cddcb3abb1083aba5a12f/Makefile#L87) file in the root of the project?

@didierofrivia @adam-cattermole, you both have experience with reading this kind of parameters for the release process. WDYT?

adam-cattermole: I would tend to agree that it could be easy to forget this and is another extra manual maintenance step for releases. That being said, if I'm not mistaken, the make target isn't used by the release image build workflow so perhaps not the end of the world if it gets missed - although that does make me wonder what is the value in setting it explicitly like this for local dev only?

I don't mind the use of the config file (build.yaml) that we used in the operator where it was autogenerated on release branches to ensure consistency but I'm not as keen on one that might sit around on main indefinitely

ehearneRedHat: Maybe we could check the latest tag on the repo, increment by 1 and append `-dev` ? It would require a bit of string formatting but I think it could be possible.

.github/workflows/build-images.yaml
```diff
@@ -32,12 +32,26 @@ jobs:
...
+        run: |
+          echo "VERSION=latest" >> $GITHUB_ENV
+      - name: Set Authorino git sha
+        id: authorino-git-sha
         run: |
           if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
           echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
           else
-          echo "VERSION=${{ github.sha }}" >> $GITHUB_ENV
+          echo "GIT_SHA=${{ github.sha }}" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
ehearneRedHat: Maybe we could change the name from `VERSION` to `AUTHORINO_VERSION` so this doesn't happen ? As for the `latest` tag, we could have the version set to whatever the latest tag on the repo is, `+1-dev` ?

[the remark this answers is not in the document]
guicassolato: I'm fine with `version: latest|vX.Y.Z, commit: 1ac43d69d..., dirty: true|false`. I understand this is how other Kuadrant components are doing, right?

`version: <next>-dev` I don't like much because half of the times it will be wrong, due to `<next>` sometimes being a minor bump, sometimes a patch.

If there is one improvement I could suggest is using the name of the branch in the version, when it's not `main` (in this case we would continue injecting `latest`), nor a tag matching `vX.Y.Z` (that suggests a stable release.) IOW,

```
version: latest|vX.Y.Z|branch-name, commit: 1ac43d69d..., dirty: true|false
```

WDYT?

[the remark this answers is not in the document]
ehearneRedHat: Sounds good to me. Will do!

Makefile
eguzki: what about `version/version.go` like we have in other go projects? I wish this would be in go.mod like in rust. But since it is not in go.mod, version.version.go looks best approach to me.

guicassolato: @eguzki, do you mean the static version in the go file? If so, no longer inject the value at build time?

ehearneRedHat: I believe so, here is an example of that in effect  https://github.com/Kuadrant/kuadrantctl/blob/main/version/version.go .

guicassolato: We can't do that, I'm afraid. It has implication downstream as well.

ehearneRedHat: OK sure, I will stick to the above plan, of implementing a `build.yaml` file at project root with the version there...

Something like:

```yaml
# Build params for compiling build image . 
build:
   version: <version>
   ... : ...
   ...
```

guicassolato: We won't change source code files downstream, @eguzki. I prefer not to touch this on a component that is actively being productised. The version number is merely informative and not worth the trouble.

eguzki: up to you, obviously. But this is not a blocking thing at all. Changing things on downstream is regular work being done daily basis. 

This effort is on pushing to have consistency across all golang components in kuadrant.

guicassolato: Thanks, @eguzki. Understood.

Let's please not hard code the version number of Authorino inside a Golang file.

Personally, I would prefer this to be the policy setting consistency across all Kuadrant components, but I understand it can be tricky for some; the ones written in Rust in particular. Either way, let's take the discussion outside of this PR, so it can progress.

For now, I vote moving forward with your [proposed](https://github.com/Kuadrant/authorino/pull/473#discussion_r1692951848) `build.yaml`, @ehearneRedHat.

.github/workflows/build-images.yaml
```diff
@@ -32,12 +32,26 @@ jobs:
           echo "IMG_TAGS=${GITHUB_REF_NAME/\//-} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
       - name: Set Authorino version
         id: authorino-version
+        run: |
+          echo "VERSION=latest" >> $GITHUB_ENV
```
guicassolato, line 36: I suppose we'll read from `build.yaml` here, right?

ehearneRedHat: Yep, exactly !

ehearneRedHat:
@guicassolato re-requesting your review... I didn't add the version to the workflow file as requested .

guicassolato:
@ehearneRedHat, I think I'm missing something...

I still see the Authorino version with `-dev` suffix hard-coded in the Makefile for the docker-build arg. I thought we'd be reading it from the `build.yaml`. 🤔

The build-images workflow on the other hand seems to be setting `VERSION` always to `latest`, even when we build for feature branches and tag releases!

I guess what I expected was:

<table>
<tr>
<th></th>
<th>main</th>
<th>feature branch</th>
<th>vX.Y.Z</th>
</tr>
<tr>
<th>Version</th>
<td>latest</td>
<td>&lt;branch-name&gt;</td>
<td>X.Y.Z</td>
</tr>
<tr>
<th>Image tags</th>
<td>latest, &lt;git-sha&gt;</td>
<td>&lt;branch-name&gt;, &lt;git-sha&gt;</td>
<td>vX.Y.Z, &lt;git-sha&gt;</td>
</tr>
</table>

It should not matter if we build the image locally (dev env) or in a GHA. The outcome should be the same, depending only on what we are building, i.e., the code currently in `main`, in a feature branch, or "vX.Y.Z"-tagged (released version).

The `build.yaml` file should be just a resource to achieve that, if needed and to avoid hard-coding version numbers inside a Makefile or a Golang file.

Another scenario – which I do not like, but could live with if the rest of Kuadrant so decides:

<table>
<tr>
<th></th>
<th>main</th>
<th>feature branch</th>
<th>vX.Y.Z</th>
</tr>
<tr>
<th>Version</th>
<td>&lt;curr+1&gt;-dev</td>
<td>&lt;curr+1&gt;-dev</td>
<td>X.Y.Z</td>
</tr>
<tr>
<th>Image tags</th>
<td>latest, &lt;git-sha&gt;</td>
<td>&lt;branch-name&gt;, &lt;git-sha&gt;</td>
<td>vX.Y.Z, &lt;git-sha&gt;</td>
</tr>
</table>

ehearneRedHat:
> @ehearneRedHat, I think I'm missing something...
> 
> I still see the Authorino version with `-dev` suffix hard-coded in the Makefile for the docker-build arg. I thought we'd be reading it from the `build.yaml`. 🤔
> 
> The build-images workflow on the other hand seems to be setting `VERSION` always to `latest`, even when we build for feature branches and tag releases!
> 
> I guess what I expected was:
> 
> main	feature branch	vX.Y.Z
> Version	latest	<branch-name>	X.Y.Z
> Image tags	latest, <git-sha>	<branch-name>, <git-sha>	vX.Y.Z, <git-sha>
> It should not matter if we build the image locally (dev env) or in a GHA. The outcome should be the same, depending only on what we are building, i.e., the code currently in `main`, in a feature branch, or "vX.Y.Z"-tagged (released version).
> 
> The `build.yaml` file should be just a resource to achieve that, if needed and to avoid hard-coding version numbers inside a Makefile or a Golang file.
> 
> Another scenario – which I do not like, but could live with if the rest of Kuadrant so decides:
> 
> main	feature branch	vX.Y.Z
> Version	<curr+1>-dev	<curr+1>-dev	X.Y.Z
> Image tags	latest, <git-sha>	<branch-name>, <git-sha>	vX.Y.Z, <git-sha>

@guicassolato you're absolutely right! I did forget to amend the `docker-build` target. 

My understanding was that you would prefer not touching the version within the `build-images.yaml` workflow . That's why I didn't adjust. 

I can see now what you were expecting, and how the current state isn't aligned with that! I think it sounds like a good idea. There would need to be some logic to check for the branches but I'll implement it!

ehearneRedHat:
@guicassolato I believe I made the requested changes. Feel free to review when you have the chance :)

Makefile
```diff
@@ -2,7 +2,7 @@
 SHELL = /bin/bash
 
 # Authorino version
-VERSION = $(shell git rev-parse HEAD)
+VERSION ?= 0.0.0
```
guicassolato, line 5: Is this being used in any case?

.github/workflows/build-images.yaml
```diff
@@ -48,7 +62,9 @@ jobs:
           tags: ${{ env.IMG_TAGS }}
           platforms: linux/amd64,linux/arm64,linux/s390x,linux/ppc64le
           build-args: |
-            version=${{ env.VERSION }}
+            GIT_SHA=${{ env.GIT_SHA }}
+            version=${{ env.version }}
+            DIRTY=${{ env.DIRTY }}
```
guicassolato, line 67: nitpick: Can we please keep the case consistent here?

.github/workflows/build-images.yaml
```diff
@@ -32,12 +32,26 @@ jobs:
...
         id: authorino-version
+        run: |
+          echo "version=latest" >> $GITHUB_ENV
+      - name: Set Authorino git sha
+        id: authorino-git-sha
         run: |
           if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+          echo "version=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
           else
```
guicassolato, line 42: It looks like we're setting the version to `latest` always; then, in some cases, setting it again to the value of `$GITHUB_REF_NAME`. I wonder if `echo ... >> $GITHUB_ENV` replaces the existing entry with a new one or (as I would expect) adds another. Which entry will `${{ env.version }}` read? Probably what we want is to set `version=latest` only when it's not a semantic version and to `$GITHUB_REF_NAME` otherwise?

Makefile
```diff
@@ -111,23 +119,37 @@ generate: vendor controller-gen ## Generates types deepcopy code
...
+build:DIRTY=$(shell $(PROJECT_DIR)/hack/check-git-dirty.sh || echo "unknown")
+build: $(shell $(PROJECT_DIR)/hack/check-version.sh)
+build:VERSION=$(shell $(YQ) '.build.version' build.yaml)
 build: generate ## Builds the manager binary
-	CGO_ENABLED=0 GO111MODULE=on go build -a -ldflags "-X main.version=$(VERSION)" -o bin/authorino main.go
+	CGO_ENABLED=0 GO111MODULE=on go build -a -ldflags "-X main.version=$(VERSION) -X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" -o bin/authorino main.go
 
 IMAGE_REPO ?= authorino
+VERSION ?= $(GIT_SHA)
 using_semantic_version := $(shell [[ $(VERSION) =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.+)?$$ ]] && echo "true")
```
guicassolato, line 139: It looks like `using_semantic_version` will never be true here when VERSION is defined equal to `$(GIT_SHA)`. What use case does this cover?

Makefile
ehearneRedHat: Besides for semantic versioning, I suppose not.

Makefile
ehearneRedHat: We can remove that line since it doesn't cover the use case of semantic versioning

.github/workflows/build-images.yaml
ehearneRedHat: We could, but there is reasoning behind why `version` is lowercase in comparison to the other build args... `VERSION` is already defined in the [parent Dockerfile](https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?container-tabs=dockerfile).  

I made the choice of making that one lowercase as was seen in the previous version of the Dockerfile. I suppose since one of the args is already lowercase, we can make all lowercase so it's less confusing.

.github/workflows/build-images.yaml
ehearneRedHat: You're right - we should add more logic so that `version` isn't in the github environment more than once at any given stage.

.github/workflows/build-images.yaml
```diff
@@ -30,14 +30,25 @@ jobs:
         id: add-branch-tag
         run: |
           echo "IMG_TAGS=${GITHUB_REF_NAME/\//-} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Set Authorino version
-        id: authorino-version
+      - name: Set Authorino git sha
+        id: authorino-git-sha
```
guicassolato, line 34:
```suggestion
      - name: Set Authorino build info
        id: authorino-build-info
```

.github/workflows/build-images.yaml
```diff
@@ -30,14 +30,25 @@ jobs:
...
         run: |
           if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+          echo "version=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
           else
-          echo "VERSION=${{ github.sha }}" >> $GITHUB_ENV
+          echo "git_sha=${{ github.sha }}" >> $GITHUB_ENV
```
guicassolato, line 39: When it's a release tag, we do not set the Git SHA?

.github/workflows/build-images.yaml
```diff
@@ -30,14 +30,25 @@ jobs:
...
         run: |
           echo "IMG_TAGS=${GITHUB_REF_NAME/\//-} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Set Authorino version
-        id: authorino-version
+      - name: Set Authorino git sha
+        id: authorino-git-sha
         run: |
           if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+          echo "version=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
```
guicassolato, line 37: Do we need to remove the "v" prefix from the version? See e.g. https://github.com/Kuadrant/authorino-operator/pull/198/files#diff-6ec9d6ffc2c8ebd369edcc698ad16020762b6616d45849d4b539557e59563884R114

Makefile
```diff
@@ -111,12 +118,20 @@ generate: vendor controller-gen ## Generates types deepcopy code
 manifests: controller-gen kustomize ## Generates the manifests in $PROJECT_DIR/install
 	controller-gen crd:crdVersions=v1 rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=install/crd output:rbac:artifacts:config=install/rbac && $(KUSTOMIZE) build install > $(AUTHORINO_MANIFESTS)
 	$(MAKE) patch-webhook
-
+	
+run:GIT_SHA=$(shell git rev-parse HEAD)
+run:DIRTY=$(shell $(PROJECT_DIR)/hack/check-git-dirty.sh || echo "unknown")
+run:$(shell $(PROJECT_DIR)/hack/check-version.sh)
+run:VERSION=$(shell $(YQ) '.build.version' build.yaml)
```
guicassolato, line 125: For consistency only:

```suggestion
run:$(shell $(PROJECT_DIR)/hack/check-version.sh)
run:version=$(shell $(YQ) '.build.version' build.yaml)
run:git_sha=$(shell git rev-parse HEAD)
run:dirty=$(shell $(PROJECT_DIR)/hack/check-git-dirty.sh || echo "unknown")
```

Makefile
```diff
@@ -111,12 +118,20 @@ generate: vendor controller-gen ## Generates types deepcopy code
...
 run: generate manifests ## Runs the application against the Kubernetes cluster configured in ~/.kube/config
-	go run -ldflags "-X main.version=$(VERSION)" ./main.go server
+	go run -ldflags "-X main.version=$(VERSION) -X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" ./main.go server
 
+build:GIT_SHA=$(shell git rev-parse HEAD)
+build:DIRTY=$(shell $(PROJECT_DIR)/hack/check-git-dirty.sh || echo "unknown")
+build: $(shell $(PROJECT_DIR)/hack/check-version.sh)
+build:VERSION=$(shell $(YQ) '.build.version' build.yaml)
```
guicassolato, line 132: Same as https://github.com/Kuadrant/authorino/pull/473#discussion_r1726904992

.github/workflows/build-images.yaml
ehearneRedHat: I can ensure it happens in all scenarios ?

Makefile
```diff
@@ -111,12 +118,20 @@ generate: vendor controller-gen ## Generates types deepcopy code
...
 	$(MAKE) patch-webhook
-
+	
+run:$(shell $(PROJECT_DIR)/hack/check-version.sh)
+run:version=$(shell $(YQ) '.build.version' build.yaml)
+run:git_sha=$(shell git rev-parse HEAD)
+run:dirty=$(shell $(PROJECT_DIR)/hack/check-git-dirty.sh || echo "unknown")
 run: generate manifests ## Runs the application against the Kubernetes cluster configured in ~/.kube/config
-	go run -ldflags "-X main.version=$(VERSION)" ./main.go server
+	go run -ldflags "-X main.version=$(VERSION) -X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" ./main.go server
```
guicassolato, line 127:
```suggestion
	go run -ldflags "-X main.version=$(version) -X main.gitSHA=$(git_sha) -X main.dirty=$(dirty)" ./main.go server
```

ehearneRedHat:
@guicassolato feel free to re-review again, but I need to check the verification steps first. You are too fast ! :D

.github/workflows/build-images.yaml
```diff
@@ -30,14 +30,25 @@ jobs:
...
         run: |
           echo "IMG_TAGS=${GITHUB_REF_NAME/\//-} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Set Authorino version
-        id: authorino-version
+      - name: Set Authorino build info
+        id: authorino-build-info
         run: |
           if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+            echo "VERSION=${tag#v}" >> $GITHUB_ENV
```
guicassolato, line 37: I haven't tested it, but maybe...

```suggestion
          tag=${GITHUB_REF_NAME/\//-}
          echo "VERSION=${tag#v}" >> $GITHUB_ENV
```

ehearneRedHat: I must have missed that part of the code... thanks!

ehearneRedHat:
https://github.com/Kuadrant/authorino/actions/runs/10507966761

.github/workflows/build-images.yaml
```diff
@@ -30,14 +30,26 @@ jobs:
...
           echo "IMG_TAGS=${GITHUB_REF_NAME/\//-} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Set Authorino version
-        id: authorino-version
+      - name: Set Authorino build info
+        id: authorino-build-info
         run: |
           if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+            tag=${GITHUB_REF_NAME/\//-}
+            echo "VERSION=${tag#v}" >> $GITHUB_ENV
```
guicassolato, line 38: Lower case?

```suggestion
            echo "version=${tag#v}" >> $GITHUB_ENV
```

Makefile
```diff
@@ -111,12 +118,20 @@ generate: vendor controller-gen ## Generates types deepcopy code
...
+	
+run:$(shell $(PROJECT_DIR)/hack/check-version.sh)
+run:version=$(shell $(YQ) '.build.version' build.yaml)
+run:git_sha=$(shell git rev-parse HEAD)
+run:dirty=$(shell $(PROJECT_DIR)/hack/check-git-dirty.sh || echo "unknown")
 run: generate manifests ## Runs the application against the Kubernetes cluster configured in ~/.kube/config
-	go run -ldflags "-X main.version=$(VERSION)" ./main.go server
+	go run -ldflags "-X main.version=$(version) -X main.gitSHA=${git_sha} -X main.dirty=${dirty}" ./main.go server
 
+build:$(shell $(PROJECT_DIR)/hack/check-version.sh)
```
guicassolato, line 129: I've tried `make build` locally, but got:

```
make: /bin/yq: Command not found
```

We probably need to add `yq` as a dependency of the target. Same for `run` and `docker-build`.

guicassolato: Actually the error seems to be `go-install-tool` instead of `go-get-tool`.

Makefile
```diff
@@ -42,6 +41,14 @@ help:
 
 ##@ Dependencies
 
+YQ=$(PROJECT_PATH)/bin/yq
+YQ_VERSION := v4.34.2
+$(YQ):
+	$(call go-install-tool,$(YQ),github.com/mikefarah/yq/v4@$(YQ_VERSION))
```
guicassolato, line 47: Actually

```suggestion
	$(call go-get-tool,$(YQ),github.com/mikefarah/yq/v4@$(YQ_VERSION))
```

Makefile
```diff
@@ -42,6 +41,14 @@ help:
 
 ##@ Dependencies
 
+YQ=$(PROJECT_PATH)/bin/yq
```
guicassolato, line 44:
```suggestion
YQ=$(PROJECT_DIR)/bin/yq
```

Makefile
guicassolato: Hmm. Both things are needed: `go-get-tool` and adding the dependency here.

hack/check-version.sh
```diff
@@ -0,0 +1,15 @@
...
+# Check if the branch is "main"
+if [[ "$branch" == "main" ]]; then
+    version="latest"
+# Otherwise, use the branch name as the version
+else
+    version="$branch"
+fi
+
+# Use yq to set build.version to version variable
+yq eval ".build.version = \"$version\"" -i build.yaml
```
guicassolato, line 15: I see this modifies the `build.yaml` file, setting the version to the name of the branch when it's not `main`. This is indeed what I had in mind. What I missed though is that this will leave a dirty change we do not want to commit.

Not sure the best to solve this. We want the branch name version in the binary, but not in the branch itself.

ehearneRedHat: Actually, this is accounted for within `hack/check-git-dirty.sh` . https://github.com/Kuadrant/authorino/blob/b2311c1825c1006cf977648ef72949144b6998b2/hack/check-git-dirty.sh#L11

guicassolato: Indeed it is but only for the purpose of setting the "dirty" flag. The dirty change to the file will remain on disk. It's a tracked file. People adding it to stage, committing the change, others overlooking it in the PRs and the thing ending up in main is only a matter of time. Believe me.

main.go
```diff
@@ -562,5 +564,9 @@ func timeoutMs(timeout int) time.Duration {
 }
 
 func printVersion(_ *cobra.Command, _ []string) {
-	fmt.Println("Authorino", version)
+	if dirty == "true" {
+		fmt.Printf("Authorino %s (%s-dirty)\n", version, gitSHA)
```
guicassolato, line 568: Am I wrong or there was a suggestion for printing this as JSON?

```json
{"version":"latest","commit":"abcd1234","dirty":true}
```

Was there any Kuadrant-wide decision about the format?

hack/check-version.sh
ehearneRedHat: I can see what you mean... if this is something you would rather not see, should we revert back to using a variable instead? 

I know before you said you didn't want to as it would have implications upstream, but to avoid this from becoming an issue it could be the better option? 

Let me know your thoughts...

main.go
ehearneRedHat: I don't believe so... 

I know for that format, that is the one we use in the logs. Example:
```bash
{"level":"info","ts":"2024-08-26T10:24:20+01:00","logger":"authorino","msg":"build information","version":"471-add-commit-hash-implementation-similar-to-kuadrantctl-to-authorino","commit":"2164661fe2f879fe07b5a1fa11ece02c75e8cef2","dirty":"false","cmd":"server"}
```

As for when we use `authorino version` , that one is not JSON, and I don't believe there was a suggestion for doing this just yet.

```bash
evanhearne@fedora:~/Documents/projects/authorino$ ./bin/authorino version
Authorino 471-add-commit-hash-implementation-similar-to-kuadrantctl-to-authorino (2164661fe2f879fe07b5a1fa11ece02c75e8cef2)
```

Sometime in the future, I believe there will be a PR opened to suggest how the version command will look like universally across all operators.

hack/check-version.sh
eguzki, line 15: instead of the name of the file `check-version.sh` , I would call this `write-version.sh` or `commit-version.sh` or something consistent with what it is doing.

eguzki: why do we need a (git checked) file that contains the branch name (calling `latest` to `main`)?

ehearneRedHat: Hey, @eguzki ! I will leave your last comment for @guicassolato who could probably answer it best. 

As for your first comment, makes sense. Will do !

guicassolato: > why do we need a (git checked) file that contains the branch name 

The idea of this file is to store all info needed to build the component that typically is passed as argument to the build command tools, thus avoiding otherwise hard-coding this info within the source, nor losing it somewhere within the Makefile. IOW, it's to separate build _data_ from build _logic_.

By "build" I mean the binary as well as the container image.

As proposed [here](https://github.com/Kuadrant/authorino/pull/473#issuecomment-2273351222), the `version` info assumes one of 3 possible values:
1. `latest` → for the latest builds based on the `main` branch – it should be constantly moving with the new SHAs being pushed to main
2. `<branch-name>` → for short-lived builds based on feature branches, mainly triggered manually during development, for testing, etc – typically lasting until a PR merged is only
3. `X.Y.Z` → for long-lived, stable builds based on release tags (named `vX.Y.Z`) – typically associated with one SHA that will never change

I guess 2 is the tricky one here. We want to be able to build binaries and tag container images after a branch name, so we identify these artifacts properly during dev/testing. However, this particular `version` info will only last until the feature branch gets merged. So, we do not want to commit `version: <branch-name>`. We want to brand the (short-lived) artifact like that, but without risking merging it. Instead, we want to commit `version: latest` (or whatever base the feature branch is targeting.)

Make sense?

ehearneRedHat: I guess in order to achieve that, we could set the version for building, and after build, revert the change. 

The downside with that is the user would not be able to see what their build info was post-build. 

Another solution might be to have a `recent` build info file which has all the info used to build at that time, and could be added to `.gitignore` since it would only exist post-build i.e. it would not be there before build. 

This would allow the user to see their most recent build config, and would allow for developers not to worry about git tracking unnecessary changes.

eguzki: This `version` information needs to be checked in git. Some projects do in the makefile, others do in `version.[go|java|..] and yet another approach is the build.yaml.  I do not really care about "how".  The "what" is what matters: version is specified somewhere. Consistency has a value, but can be sacrificed for greater good. 

All projects need some (documented) manual procedure or automated script to update the version and commit it.

What it does not make sense is to change is when image is building or executing "run" command. Image building or execute run command read build.yaml and with ldflags the information is embedded in the binary, available for future reference.

guicassolato: > What it does not make sense is to change is when image is building or executing "run" command. Image building or execute run command read build.yaml and with ldflags the information is embedded in the binary, available for future reference.

Completely agree. The `build.yaml` should be the input here, not the output.

In general, the value of `version` inside `build.yaml` should be statically set to `latest`. I can see value in allowing make targets to use `build.yaml` or override it "on-the-fly" with something like the name of the branch, when running the workflow from a PR (via GHA) or locally, but without touching `build.yaml` then.

Separately, we need to update the [release process](https://github.com/Kuadrant/authorino/blob/main/RELEASE.md) to now include:
1. creating a `release-vX.Y.Z` branch
2. updating and commit `build.yaml` in the branch
3. Tag the release branch instead of `main`

We've never needed `release-vX.Y.Z` branches in Authorino and have always managed to simply tag a commit in `main` to cut a release, but that was because the version number never had to be committed along with the code before.

ehearneRedHat: OK, so I will go ahead with ensuring the build.yaml `.version` is statically kept as `latest` , having a recent build file ignored in `.gitignore`, and update the release process with this new information.

ehearneRedHat:
@guicassolato I have amended changes - feel free to review. :)

main.go
guicassolato: No problem. I probably got that wrong then. It works for me the way it is now.

.github/workflows/build-images.yaml
```diff
@@ -30,14 +30,28 @@ jobs:
...
         run: |
           if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
-          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+            tag=${GITHUB_REF_NAME/\//-}
+            echo "version=${tag#v}" >> $GITHUB_ENV
+          elif [[ ${GITHUB_REF_NAME/\//-} == "main" ]]; then
+            echo "version=latest" >> $GITHUB_ENV
           else
-          echo "VERSION=${{ github.sha }}" >> $GITHUB_ENV
+            echo "version=${{ github.ref_name }}" >> $GITHUB_ENV
```
guicassolato, line 42: It looks like we still have 2 ways to do the same thing more or less. Even though GHA and Makefile both seem to use the git ref, for some reason one does it straightforward and the other uses a file (two files!) in the process.

This GHA workflow (used to build and push the images to registry on pushes to main, releases and for PR testing) uses only the git ref straight away to infer the version info. The Makefile (used locally for dev/testing workflows), on the other hand, reads the version info from `build-recent.yaml`, while still generating this file (from a `build.yaml` "template") before every command, based on the git ref.

I see two paths from here to fix this. We need to decide whether `build.yaml` will be the source for the version info or not.

**OPTION A:** `build.yaml` is the source of version info:

1. The version info should NOT be updated before every command, but simply read from the `build.yaml` file locally (this is the base line, but keep reading to see the special case)
2. Updating `build.yaml` must be a manual step only performed in preparation for releases
3. We need to start doing release branches for Authorino, so these can contain the version of `build.yaml` with the altered version info inside (i.e. "X.Y.Z"), unlike all other branches (including main) whose `build.yaml` will always say "latest"
4. The GHA must read the version info also from `build.yaml`
5. There must be a way for the user to manually override the info in `build.yaml` without modifying the file. This will be the case mostly of when building from feature branches/PRs, both using the Makefile or via GHA. We probably don't need `build-recent.yaml` for that.

**OPTION B:** `build.yaml` is NOT the source:

1. version info must be inferred from the git ref directly, both in the GHA and in the make targets:

| git ref | version info |
|---------|--------------|
| main    | latest       |
| feature | branch-name  |
| vX.Y.Z  | X.Y.Z        |

IMO, OPTION B looks much simpler. On the flip side, it doesn't give us the traceability, along with the code, about which version info would be stamped when building from that source. However, we're also introducing git shas and the `dirty` flag here. So I tend to think that storing the version info along with the code is not needed.

But please let me know what you think @ehearneRedHat @eguzki @didierofrivia

ehearneRedHat: @guicassolato I think `OPTION B` looks to be easier but you would lose the traceability as you mentioned, but there would be less files to look to fix. 

I will await other responses before implementing.

eguzki: The `version` info in logs gives high level view of the version being used.  Useful to know if some feature should be there or not. 

For us, developers, `version` is not accurate enough to debug and investigate issues. We want the git sha revision to know exactly the source code revision being run. 

I leave `version` management up to you guys. 

My take? version lives in `version.go`. In `main` branch it is `vX.Y.Z-dev` (where vX.Y.Z has not been released yet). Note that it is not `latest` or `main` as it is meaningless. In release branches it is `vX.Y.Z`. In RELEASE.md file, there is an additional step to update this `version.go` file. This is the approach we are taking in other kuadrant projects.

ehearneRedHat: @didierofrivia what are your thoughts ?

ehearneRedHat:
I pass the baton over to @guicassolato to complete this task, as today marks my last day of the internship in Kuadrant at Red Hat. 

Thank you to everyone who has helped me over the eight months of my being here. It was an amazing experience and hope to be back someday soon. :v:

guicassolato:
> I pass the baton over to @guicassolato to complete this task, as today marks my last day of the internship in Kuadrant at Red Hat.
> 
> Thank you to everyone who has helped me over the eight months of my being here. It was an amazing experience and hope to be back someday soon. ✌️

Thank you so much for this, @ehearneRedHat! I will own the PR now, though the credit is still all yours, if you ask me!

.github/workflows/build-images.yaml
guicassolato: Going with OPTION B here for simplicity.

alexsnaps (approved):
Minor questions... but this should do the job

Dockerfile
```diff
@@ -4,8 +4,13 @@ FROM registry.access.redhat.com/ubi9/go-toolset:1.21 AS builder
 USER root
 WORKDIR /usr/src/authorino
 COPY ./ ./
-ARG version=latest
-RUN CGO_ENABLED=0 GO111MODULE=on go build -a -ldflags "-X main.version=${version}" -o /usr/bin/authorino main.go
+ARG version
+ENV version=${version:-unknown}
```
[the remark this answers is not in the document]
alexsnaps: Ignore me... the `-` is part of the `:-` sequence, not the value to "default to"...

PR #474, opened 2024-07-05
fixup: some broken docs links

jasonmadigan:
We've a decent amount of 404'ing content across our docs content.

https://github.com/Kuadrant/docs.kuadrant.io/actions/runs/9807810746/job/27082317585

Some noticed in Authorino's docs have a similar pattern which should be simple/regexable to fix.

Example Broken links: https://docs.kuadrant.io/0.8.0/authorino/docs/user-guides/resource-level-authorization-uma/

External auth metadata → User-Managed Access (UMA) resource registry => 
https://docs.kuadrant.io/0.8.0/authorino/docs/user-guides/features.md#user-managed-access-uma-resource-registry-metadatauma

Should be:

https://docs.kuadrant.io/0.8.0/authorino/docs/features#user-managed-access-uma-resource-registry-metadatauma

With markdown links, the relative link above are resolved and pre-processed before render. HTML is generally passed straight through, untouched, so these relative links end up broken unfortunately. 

This is probably a bit of a PITA to review: sorry!

### DIFF
--- a/docs/user-guides/anonymous-access.md
+++ b/docs/user-guides/anonymous-access.md
@@ -3,17 +3,12 @@
 Bypass identity verification or fall back to anonymous access when credentials fail to validate
 
 <details markdown="1">
-  <summary>
-    <strong>Authorino capabilities featured in this guide:</strong>
-    <ul>
-      <li>Identity verification & authentication → <a href="./../features.md#anonymous-access-authenticationanonymous">Anonymous access</a></li>
-    </ul>
-  </summary>
+  <summary>Authorino capabilities featured in this guide:</summary>
 
-  For further details about Authorino features in general, check the [docs](./../features.md).
+  - Identity verification & authentication → [Anonymous access](./../features.md#anonymous-access-authenticationanonymous)
+
+    For further details about Authorino features in general, check the [docs](./../features.md).
 </details>
-
-<br/>
 
 ## Requirements
 
--- a/docs/user-guides/caching.md
+++ b/docs/user-guides/caching.md
@@ -17,21 +17,16 @@ Cases where one will **NOT** want to enable caching, due to relatively cheap com
 - Anonymous access
 
 <details markdown="1">
-  <summary>
-    <strong>Authorino capabilities featured in this guide:</strong>
-    <ul>
-      <li>Common feature → <a href="../features.md#common-feature-caching-cache">Caching</a></li>
-      <li>Identity verification & authentication → <a href="../features.md#anonymous-access-authenticationanonymous">Anonymous access</a></li>
-      <li>External auth metadata → <a href="../features.md#http-getget-by-post-metadatahttp">HTTP GET/GET-by-POST</a></li>
-      <li>Authorization → <a href="../features.md#open-policy-agent-opa-rego-policies-authorizationopa">Open Policy Agent (OPA) Rego policies</a></li>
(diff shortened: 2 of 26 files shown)


docs/user-guides/anonymous-access.md
```diff
@@ -6,7 +6,7 @@ Bypass identity verification or fall back to anonymous access when credentials f
   <summary>
     <strong>Authorino capabilities featured in this guide:</strong>
     <ul>
-      <li>Identity verification & authentication → <a href="./../features.md#anonymous-access-authenticationanonymous">Anonymous access</a></li>
+      <li>Identity verification & authentication → [Anonymous access](./../features.md#anonymous-access-authenticationanonymous)</li>
```
guicassolato, line 9: I believe the reason why this is broken in the Docs website is not the HTML vs MD format, but because of the relative path – which BTW currently does not break the link when the page is rendered in Github:

Screenshot 2024-07-06 at 03 35 04
Screenshot 2024-07-06 at 03 32 45

<br/><br/>

Using MD format does NOT work in Github in this case where the content is still wrapped within a HTML tag (`details > summary > ul > li`).

Before this PR:
Screenshot 2024-07-06 at 03 29 29


After this PR:
Screenshot 2024-07-06 at 03 29 23

Ideally, we'd find a fix that makes it work both in Github and in the Docs website. Maybe try something focused on the relative path?

guicassolato: If it helps, we can change to avoid this pattern of an anchor link within a detail summary. It's probably cumbersome for the user anyway; depending if click on the link or its surrounding you either follow the link or open the details box.

I think it's repeated at all Authorino User guides, so some work to do, but probably worthy. WDYT?

jasonmadigan: That would probably help. Tedious to do, but might be worth it. Happy to help with it

jasonmadigan: OK done - works good for both mkdocs rendered content + GH.

jasonmadigan:
Lint action seems to be linting binary files (I wish it good luck!):

Screenshot 2024-07-09 at 12 55 37

docs/user-guides/caching.md
```diff
@@ -17,22 +17,17 @@ Cases where one will **NOT** want to enable caching, due to relatively cheap com
...
-      <li>Common feature → <a href="../features.md#common-feature-caching-cache">Caching</a></li>
-      <li>Identity verification & authentication → <a href="../features.md#anonymous-access-authenticationanonymous">Anonymous access</a></li>
-      <li>External auth metadata → <a href="../features.md#http-getget-by-post-metadatahttp">HTTP GET/GET-by-POST</a></li>
-      <li>Authorization → <a href="../features.md#open-policy-agent-opa-rego-policies-authorizationopa">Open Policy Agent (OPA) Rego policies</a></li>
-      <li>Dynamic response → <a href="../features.md#json-injection-responsesuccessheadersdynamicmetadatajson">JSON injection</a></li>
-    </ul>
-  </summary>
+  <summary>Authorino capabilities featured in this guide</summary>
+
+  - Common feature → [Caching](../features.md#common-feature-caching-cache)
```
david-martin, line 22: I came across this today and hit a 404 on the docs site at https://docs.kuadrant.io/0.11.0/authorino/docs/user-guides/caching/

@guicassolato I think this should be merged asap as there's lots of broken/404 links

PR #483, opened 2024-08-28
Make v1beta2 the stored version

alexsnaps:
See #479

PR #488, opened 2024-09-18
feat: configurable API Key K8s secret key name

### LINKED ISSUE #360: Configurable API Key K8s secret key name
API Key values currently need to be stored in a key necessarily named as `api_key` within the Kubernetes Secret resource. It would be nice:
1. to be able to customise the name of this key within the secret; perhaps
2. making possible to provide multiple name options.
 
E.g.:

```yaml
kind: AuthConfig
metadata:
  name: my-api-protection
spec:
  hosts: [...]
  identity:
  - name: friends
    apiKey:
      selector:
        matchLabels:
          group: friends
      keySelectors:
      - first-api-key-value
      - second-api-key-value
    credentials:
      in: authorization_header
      keySelector: APIKEY
```

So a Kubernetes as such could be defined:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: api-key-1
  labels:
    group: friends
stringData:
  first-api-key-value: secret
  second-api-key-value: other-secret
type: Opaque
```

Point number 1 above helps with avoiding the strict constraint on the name and implications such as in some case not being able to use the same Kubernetes Secret resource that stores the API key for other purposes or for different AuthConfigs (with different API key secret values).

And, by supporting multiple valid key names (point number 2 above), that Authorino would try in order when reading the secret value of the API key (stopping when the first valid key name is found within the Kubernetes Secret), this change would also make it easier to implement key rotation, which otherwise could only be done by creating a new Kubernetes Secret.

KevFan:
# Description
Closes: https://github.com/Kuadrant/authorino/issues/360

Adds Key Selectors support for API Key k8s secret

# Verification
* Create local cluster
```sh
make local-setup
```
* Apply secret
```sh
kubectl apply -f -<<EOF
apiVersion: v1
kind: Secret
metadata:
  name: api-key-1
  labels:
    group: friends
stringData:
  api_key:  key_1
  api_key_2: key_2
  key3: key_3
EOF
```
* Apply AuthConfig
```sh
kubectl apply -f -<<EOF                                                           
kind: AuthConfig   
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "friends":
      apiKey:
        selector:
          matchLabels:
            group: friends
      credentials:
        authorizationHeader:
          prefix: APIKEY
EOF
```
* Port forward envoy
```sh
```
* Curl using `api_key` key (default key as not key selector was defined) should succeed
```sh
curl -H 'Authorization: APIKEY key_1' http://talker-api.127.0.0.1.nip.io:8000/hello
# HTTP/1.1 200 OK
```
* Curl using `api_key_2` and `key 3` key should fail with `401 Unauthorized`
```sh
curl -H 'Authorization: APIKEY key_2' http://talker-api.127.0.0.1.nip.io:8000/hello -v
curl -H 'Authorization: APIKEY key_3' http://talker-api.127.0.0.1.nip.io:8000/hello -v
```
* Update AuthConfig to only selector for `api_key_2`
```sh
kubectl apply -f -<<EOF                                                           
kind: AuthConfig   
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "friends":
      apiKey:
        selector:
          matchLabels:
            group: friends
        keySelector: "['api_key_2']"
      credentials:
        authorizationHeader:
          prefix: APIKEY
EOF
```
* Curling with `api_key_2` should succeed
```sh
curl -H 'Authorization: APIKEY key_2' http://talker-api.127.0.0.1.nip.io:8000/hello -v
# HTTP/1.1 200 OK
```
* Curl using `api_key` and `key 3` key should fail with `401 Unauthorized`
```sh
curl -H 'Authorization: APIKEY key_1' http://talker-api.127.0.0.1.nip.io:8000/hello -v
curl -H 'Authorization: APIKEY key_3' http://talker-api.127.0.0.1.nip.io:8000/hello -v
```
* Update AuthConfig to selector for keys matching `api_key*`
```sh
kubectl apply -f -<<EOF                                                           
kind: AuthConfig   
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "friends":
      apiKey:
        selector:
          matchLabels:
            group: friends
        keySelector: "keys.filter(k, k.startsWith('api_key'))"
      credentials:
        authorizationHeader:
          prefix: APIKEY
EOF
```
* Curling with `api_key` and `api_key_2` should succeed
```sh
curl -H 'Authorization: APIKEY key_1' http://talker-api.127.0.0.1.nip.io:8000/hello 
curl -H 'Authorization: APIKEY key_2' http://talker-api.127.0.0.1.nip.io:8000/hello 
```
* Curling with `key_3`  key should fail with `401 Unauthorized`
```sh
curl -H 'Authorization: APIKEY key_3' http://talker-api.127.0.0.1.nip.io:8000/hello -v
```

* Update secret to remove `api_key_2`
```sh
kubectl patch secret api-key-1 -p '{"data":{"api_key_2": null}}'
```
* Curling with `api_key_2` should fail
```sh
curl -H 'Authorization: APIKEY key_2' http://talker-api.127.0.0.1.nip.io:8000/hello -v
# HTTP/1.1 401 Unauthorized
```
* Curling with `api_key` should still succeed 
```sh
curl -H 'Authorization: APIKEY key_1' http://talker-api.127.0.0.1.nip.io:8000/hello
# HTTP/1.1 200 OK
```
* Create second secret
```sh
kubectl apply -f -<<EOF
apiVersion: v1
kind: Secret
metadata:
  name: api-key-2
  labels:
    group: friends
stringData:
  api_key_2: another_2
EOF
```
* Curling with `api_key_2` from second secret should succeed
```
curl -H 'Authorization: APIKEY another_2' http://talker-api.127.0.0.1.nip.io:8000/hello 
# HTTP/1.1 200 OK
```
* Delete first secet
```sh
kubectl delete secret api-key-1
```
* Curling with `api_key`, `api_key_2` and `api_key_3` should fail 
```sh
curl -H 'Authorization: APIKEY key_1' http://talker-api.127.0.0.1.nip.io:8000/hello 
curl -H 'Authorization: APIKEY key_2' http://talker-api.127.0.0.1.nip.io:8000/hello 
curl -H 'Authorization: APIKEY key_3' http://talker-api.127.0.0.1.nip.io:8000/hello 
```
* Update second secret with new key
```sh
kubectl apply -f -<<EOF
apiVersion: v1
kind: Secret
metadata:
  name: api-key-2
  labels:
    group: friends
stringData:
  api_key_2: another_2
  api_key_3: another_3
EOF
```
* Curling with `api_key_2` and `api_key_3` should succeed 
```sh
curl -H 'Authorization: APIKEY another_2' http://talker-api.127.0.0.1.nip.io:8000/hello 
curl -H 'Authorization: APIKEY another_3' http://talker-api.127.0.0.1.nip.io:8000/hello 
# HTTP/1.1 200 OK
```


guicassolato (commented):
Thank you so much for taking the time to address this issue, @KevFan!

I've left a few comments. Nothing that invalidates the change, but mainly to double check on the design we want for the feature and some implementation detail.

One general wondering I have though is whether we want to allow this change into v1beta2 or v1beta3 (addressing #480 first). Although the `keySelectors` field not being backwards compatible with v1beta1 is not much of a concern (since #483 was merged), it feels like breaking a promise (internal agreement maybe?) that all changes to the API now would go into v1beta3.

Do you have an opinion on this, @alexsnaps?

pkg/evaluators/identity/api_key.go
```diff
@@ -26,18 +26,20 @@ type APIKey struct {
...
 }
 
-func NewApiKeyIdentity(name string, labelSelectors k8s_labels.Selector, namespace string, authCred auth.AuthCredentials, k8sClient k8s_client.Reader, ctx context.Context) *APIKey {
+func NewApiKeyIdentity(name string, labelSelectors k8s_labels.Selector, namespace string, keySelectors []string, authCred auth.AuthCredentials, k8sClient k8s_client.Reader, ctx context.Context) *APIKey {
 	apiKey := &APIKey{
 		AuthCredentials: authCred,
 		Name:            name,
 		LabelSelectors:  labelSelectors,
 		Namespace:       namespace,
+		KeySelectors:    append(keySelectors, apiKeySelector),
```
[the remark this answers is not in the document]
KevFan: Sounds good 👍 

On a related note, what's your thoughts on instead of doing this `append`, use the kubebuilder default annotation in the API struct instead? The `append` works but this default value is hidden from the CRD, but not sure is this want we want in general
```
	// +kubebuilder:default:={"api_key"}
```

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1,5 +1,4 @@
 //go:build !ignore_autogenerated
-// +build !ignore_autogenerated
```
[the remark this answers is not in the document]
KevFan: Hmm, not sure about this one, this still give this diff on go1.21 locally. This seems to be more related `controller-gen` v0.15.0 instead requiring go1.22. So people's `authorino/bin/controller-gen` might be outdated if it's already been downloaded before since I believe it won't be redownloaded by the makefile command if it's already present 🤔 

image

alexsnaps:
>all changes to the API now would go into v1beta3

So... @guicassolato, you'd bump the version with this change, is that it? I'm growing concerned I won't get the CEL stuff done by the EOW tho... So the question becomes more of _when_ would `v1beta3` be releasable? If we miss the mark of next week's release, can we keep authorino at `v0.18` for another iteration? I think we could, hoping no need for a `v0.18.1` 🙈 !

api/v1beta3/auth_config_types.go
```diff
@@ -355,6 +355,12 @@ type ApiKeyAuthenticationSpec struct {
 	// +optional
 	// +kubebuilder:default:=false
 	AllNamespaces bool `json:"allNamespaces,omitempty"`
+
+	// List of keys within the selected Kubernetes secret that contain valid API credentials.
+	// Authorino will attempt to authenticate using the first key that matches.
+	// If no match is found, authentication will fail.
```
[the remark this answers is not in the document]
KevFan: That makes sense, I'll update 👍

pkg/evaluators/identity/api_key.go
```diff
@@ -103,17 +105,19 @@ func (a *APIKey) AddK8sSecretBasedIdentity(ctx context.Context, new k8s.Secret)
...
-				delete(a.secrets, oldAPIKeyValue)
-				logger.V(1).Info("api key updated")
-			} else {
-				logger.V(1).Info("api key unchanged")
+	for _, key := range a.KeySelectors {
+		newAPIKeyValue := string(new.Data[key])
+		for oldAPIKeyValue, current := range a.secrets {
+			if current.GetNamespace() == new.GetNamespace() && current.GetName() == new.GetName() {
+				if oldAPIKeyValue != newAPIKeyValue {
+					a.appendK8sSecretBasedIdentity(new)
```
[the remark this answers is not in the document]
KevFan: Yeah, I was doing some thinking on this. That makes sense and I agree that all API key values whose key match any of the `APIKey.KeySelectors` should be valid and added to the  `APIKey.secrets` map 👍

pkg/evaluators/identity/api_key.go
```diff
@@ -134,7 +144,6 @@ func (a *APIKey) RevokeK8sSecretBasedIdentity(ctx context.Context, deleted k8s_t
 		if secret.GetNamespace() == deleted.Namespace && secret.GetName() == deleted.Name {
 			delete(a.secrets, key)
 			log.FromContext(ctx).WithName("apikey").V(1).Info("api key deleted")
-			return
```
KevFan, line 137: Should no longer return early since the change now allows for the possibility of multiple keys pointing to the same secret

pkg/evaluators/identity/api_key.go
```diff
@@ -103,17 +105,19 @@ func (a *APIKey) AddK8sSecretBasedIdentity(ctx context.Context, new k8s.Secret)
...
-		if current.GetNamespace() == new.GetNamespace() && current.GetName() == new.GetName() {
-			if oldAPIKeyValue != newAPIKeyValue {
-				a.appendK8sSecretBasedIdentity(new)
-				delete(a.secrets, oldAPIKeyValue)
-				logger.V(1).Info("api key updated")
-			} else {
-				logger.V(1).Info("api key unchanged")
+	for _, key := range a.KeySelectors {
+		newAPIKeyValue := string(new.Data[key])
+		for oldAPIKeyValue, current := range a.secrets {
```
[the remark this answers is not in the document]
KevFan: Not sure does this still apply with the latest changes, but let me know if you still would like this. I'm assuming you mean `k8s secret namespaced name → list of api key value`. Initial thoughts is that maintaining both maps would add more complexity since they both need to be kept consistent. Not sure is it worth it unless we're expecting the possibility of a large list of keys per secret? 🤔

api/v1beta3/auth_config_types.go
```diff
@@ -355,6 +355,12 @@ type ApiKeyAuthenticationSpec struct {
 	// +optional
 	// +kubebuilder:default:=false
 	AllNamespaces bool `json:"allNamespaces,omitempty"`
+
+	// List of keys within the selected Kubernetes secret that contain valid API credentials.
+	// Authorino will attempt to authenticate using any matching key, including "api-key".
+	// If no match is found, the Kubernetes secret is not considered a valid Authorino API Key secret and is ignored.
+	// +optional
+	KeySelectors []string `json:"keySelectors,omitempty"`
```
KevFan, line 363: Do we want to define the kubebuilder default here also? 🤔 
```
	// +kubebuilder:default:={"api_key"}
```

Do we want to set a maximum?
```
+kubebuilder:validation:MaxItems
```

pkg/evaluators/identity/api_key.go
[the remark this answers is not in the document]
KevFan: > But I guess the part of making it easier to implement key rotation is not really true, is it?

Thinking on this, I don't think this PR enables/implements key rotation per se. Can this not be already achieved by having multiple secrets that an AuthConfig selects? If so, then this to me enables users another way which this can be achieved (aside from allowing them to define the key selectors) 🤔

pkg/evaluators/identity/api_key.go
```diff
@@ -26,18 +28,21 @@ type APIKey struct {
...
 }
 
-func NewApiKeyIdentity(name string, labelSelectors k8s_labels.Selector, namespace string, authCred auth.AuthCredentials, k8sClient k8s_client.Reader, ctx context.Context) *APIKey {
+func NewApiKeyIdentity(name string, labelSelectors k8s_labels.Selector, namespace string, keySelectors []string, authCred auth.AuthCredentials, k8sClient k8s_client.Reader, ctx context.Context) *APIKey {
 	apiKey := &APIKey{
 		AuthCredentials: authCred,
 		Name:            name,
 		LabelSelectors:  labelSelectors,
 		Namespace:       namespace,
+		KeySelectors:    append(keySelectors, defaultAPIKeySelector),
```
guicassolato, line 45: This means the user cannot invalidate `api_key` as a key then, no? Should we only `append` when `len(keySelectors) == 0` maybe?

api/v1beta3/auth_config_types.go
guicassolato: This is a good question. I guess the only reason I can think of to do that is to help users not to shoot themselves in the foot when tempted to add hundreds, thousands of valid keys as a workaround to store multiple valid API keys inside of a same Kubernetes secret.

If we ignore performance (and let users figure that out by themselves), arguably storing multiple valid identities in the same Secret may not be necessarily such a bad idea－even though it isn't how this feature was originally designed to work and users would lose capabilities in the process (e.g. storing one identity's metadata in the secret's metadata).

Say we do want to let users store multiple API keys in a Secret (now that it will be technically possible anyway). Probably, having to come up with a different key name for each identity would be a PITA. I can see requests of extending the key selector to support things like regexes, globs, etc coming up.

So I wonder, should we embrace that we make the key selector to be a CEL expression?

cc @alexsnaps @maleck13

pkg/evaluators/identity/api_key.go
guicassolato: Yep. I guess so.

api/v1beta3/auth_config_types.go
KevFan: I think that is a good point, moving to CEL expression would make a more consistent experience while giving users more capabilities to how they want to select the secret keys 👍

KevFan: I've a working implementation based on how it's currently done for the AuthJSON and implementing https://github.com/Kuadrant/authorino/blob/b2bde91bfc64b1d140f740069ba36ce81a3f3349/pkg/expressions/types.go#L3-L5 

Interested in @alexsnaps opinions on this

api/v1beta3/auth_config_types.go
```diff
@@ -355,6 +355,15 @@ type ApiKeyAuthenticationSpec struct {
...
 	AllNamespaces bool `json:"allNamespaces,omitempty"`
+
+	// A Common Expression Language (CEL) expression that evaluates to a list of string keys within the selected Kubernetes
+	// secret that contain valid API credentials. The keys of the selected Kubernetes secret are available for evaluation
+	// in the following structure: `{"keys": ["key1", "key2"]}`.
+	// Authorino will attempt to authenticate using any matching key. If no keys are defined, the default "api_key" will be used.
+	// If no match is found, the Kubernetes secret is not considered a valid Authorino API Key secret and is ignored.
+	// String expressions are supported (https://pkg.go.dev/github.com/google/cel-go/ext#Strings).
+	// +optional
+	KeySelector CelExpression `json:"keySelector,omitempty"`
```
guicassolato, line 366: I guess we're OK with not kicking off on a new version of the API here, being this an additive change and optional field anyway, or are we not?

Interested in your thoughts @maleck13. Managing CRD versions has been proven to be tricky, to say the least. We probably want to get this right especially given the change will eventually bubble up to the level of the AuthPolicy, already marked as v1.

guicassolato (dismissed):
Verification steps work.

nit: we may want to fix perhaps only where it says:

> - Curling with api_key_2 should succeed
>
> ```sh
> curl -H 'Authorization: APIKEY key_2' http://talker-api.127.0.0.1.nip.io:8000/hello -v
> # HTTP/1.1 401 Unauthorized
> ```

Last line (commented) should probably read `# HTTP/1.1 200 OK`.

<br/>

Let's just make sure we're all in agreement regarding https://github.com/Kuadrant/authorino/pull/488/files#r1958525812 before merging this (or making the proper adjustments.)

Nice job, @KevFan!

alexsnaps:
I wondered if binding the `keys` root to the set of keys could have become an issue. Looking at the [API](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L6145) tho, there is not much we'd miss from the whole secret tho. Most importantly tho, afaict if I read this properly, it's _always_ a unique set of keys (whether the secret defines `stringData` or `data`... is both possible?). So I think there isn't much issues with "only" exposing `keys` for now and possibly expose more (through an additional root) when needed, e.g. `type` or `immutable`, or even `secret` entirely _when needed_. 
Thanks for the PR!

pkg/evaluators/identity/api_key_test.go
```diff
@@ -65,6 +65,27 @@ func TestNewApiKeyIdentitySingleNamespace(t *testing.T) {
...
+	defer ctrl.Finish()
+
+	selector, _ := k8s_labels.Parse("planet=coruscant")
+	apiKey := NewApiKeyIdentity("jedi", selector, "ns1", "['no_op','api_key_2']", mock_auth.NewMockAuthCredentials(ctrl), testAPIKeyK8sClient, context.TODO())
+
+	assert.Equal(t, apiKey.Name, "jedi")
+	assert.Equal(t, apiKey.LabelSelectors.String(), "planet=coruscant")
+	assert.Equal(t, apiKey.Namespace, "ns1")
+	assert.Equal(t, len(apiKey.secrets), 1)
+	_, exists := apiKey.secrets["ObiWanKenobiLightSaber"]
```
didierofrivia, line 79: <3

api/v1beta3/auth_config_types.go
KevFan: Just to note since I think there was a question on this during the demo. If a user just defines a string value `keySelector: api_key_2` instead of ` keySelector: "['api_key_2']"`. There will be an error in the log saying 
```log
2025-02-19T10:46:50Z	ERROR	authorino.controller-runtime.manager.controller.authconfig.identity.apikey	failed to create key selector expression	{"authconfig": "default/talker-api-protection", "error": "ERROR: <input>:1:1: undeclared reference to 'api_key_2' (in container '')\n | api_key_2\n | ^"}
github.com/kuadrant/authorino/pkg/evaluators/identity.NewApiKeyIdentity
	/usr/src/authorino/pkg/evaluators/identity/api_key.go:53
github.com/kuadrant/authorino/controllers.(*AuthConfigReconciler).translateAuthConfig
	/usr/src/authorino/controllers/auth_config_controller.go:265
github.com/kuadrant/authorino/controllers.(*AuthConfigReconciler).Reconcile
	/usr/src/authorino/controllers/auth_config_controller.go:116
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/usr/src/authorino/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:119
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/usr/src/authorino/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/usr/src/authorino/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/usr/src/authorino/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:227
```
And the `AuthConfig` would be still marked be marked as ready in the status due to no `err` in returned at https://github.com/KevFan/authorino/blob/ea72dd422a9db8c582178517a93e2b5423908d18/controllers/auth_config_controller.go#L265. Is this the expected behaviour that we want @guicassolato ? I'm wondering since errors were just logged already at https://github.com/KevFan/authorino/blob/main/pkg/evaluators/identity/api_key.go#L45 🤔 

I'm going to push an extra commit update the description with example to make this more obvious at least

KevFan:
> I wondered if binding the `keys` root to the set of keys could have become an issue. Looking at the [API](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L6145) tho, there is not much we'd miss from the whole secret. Most importantly, afaict if I read this properly, it's _always_ a unique set of keys (whether the secret defines `stringData` or `data`... is both possible?). So I think there isn't much issues with "only" exposing `keys` for now and possibly expose more (through an additional root) when needed, e.g. `type` or `immutable`, or even `secret` entirely _when needed_. Thanks for the PR!

Thanks @alexsnaps for raising this. Yes, I believe its always a unique set of keys also. Agreed, I'm in favour of possibly exposing more of the secret as needed in any future work 👍

api/v1beta3/auth_config_types.go
guicassolato: If the CEL expression does not resolve to the expected data type, I would invalidate the AuthConfig, instead of silently logging the error only. The owner of the AuthConfig must be aware of the issue to be able to act on it (by fixing the AuthConfig, in this case).

IMO, this is not like cases such as https://github.com/KevFan/authorino/blob/b2bde91bfc64b1d140f740069ba36ce81a3f3349/pkg/evaluators/identity/api_key.go#L45, where the AuthConfig is valid. It's just loading the secrets that failed, which could be due to some API server error, for example.

guicassolato:
> > I wondered if binding the `keys` root to the set of keys could have become an issue. Looking at the [API](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L6145) tho, there is not much we'd miss from the whole secret. Most importantly, afaict if I read this properly, it's _always_ a unique set of keys (whether the secret defines `stringData` or `data`... is both possible?). So I think there isn't much issues with "only" exposing `keys` for now and possibly expose more (through an additional root) when needed, e.g. `type` or `immutable`, or even `secret` entirely _when needed_. Thanks for the PR!
> 
> Thanks @alexsnaps for raising this. Yes, I believe its always a unique set of keys also. Agreed, I'm in favour of possibly exposing more of the secret as needed in any future work 👍

Let me start by saying that API key authentication, without proper guardrails set by a cluster admin beforehand, can easily be the least secure feature of Authorino.

On one hand, giving access to the entire secret by binding to the root of the object instead of `keys` would enable extra functionality, not yet contemplated in this PR. E.g., one could store the valid key names for any given API key secret in the annotations of the secret itself; then use this root binding to read the annotation, cast its value to the expected format (list of strings), thus achieving dynamic key names on a per secret basis.

Only this change could make label selectors obsolete (if it wasn't for the optimisation aspect involved when querying the API server of course).

On the other hand, such change marginally increases an existing known vulnerability of unsafe Authorino deployments, that allow malicious users to elevate privileges and get access to Kubernetes secrets otherwise out of reach.

While checking the listed secrets for valid API keys stored in it, [this line](https://github.com/Kuadrant/authorino/blob/01a81eb499e81660a58f410ed8ceaee27ac47846/pkg/evaluators/identity/api_key.go#L208) could be exploited to leak sensitive data to the logs, such as other values stored in the secret that are unrelated to API key authn. Combined with `allNamespaces: true` and some rather loose labels selectors, this can be a recipe for disaster in a cluster without proper RBAC rules set for the Authorino instance.

I say "marginally", however, because arguably such vulnerability already exists. It exists in 2 levels.

First, because any secret deemed to contain a valid API key secret already gets a root binding at `auth.identity`. This happens at request time, in the data plane, instead of control plane, but one who happens to know what's stored in an `api_key` entry of a valid API key secret can read the entire secret today already. This is by design and meant to let applications store user info in the secret's metadata, store other secrets in association with the user identity (that are later used in calls to fetch metadata from external HTTP services, e.g.), etc.

The second level is enabled by this PR. One who can guess the value of any key stored in a secret－it doesn't have to be necessarily the value of the `api_key` entry only－could write an AuthConfig that adds the known key to the list of valid key names, and use that to get access to the rest of the secret.

KevFan:
Closing this PR following our discussion on the security implications of this implementation. We’ve decided to revisit the feature later, as it may require a broader redesign of how Authorino handles API Key authentication

PR #495, opened 2024-10-14
First seam: DynamicCEL

alexsnaps:


install/rbac/role.yaml
```diff
@@ -41,3 +121,69 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - operator.authorino.kuadrant.io
+  resources:
+  - authorinos
```
guicassolato, line 127: There's something wrong going on here. It's somehow bringing to the operand permissions that only belong to the operator.

alexsnaps: Ah! Again... remember it happened to me already when I refactored the controller to `v1beta2` 🤔

alexsnaps:
🤦 This should in `v1beta3` ... 
I think we still miss
 - [ ] Some "CEL enabled" `HttpEndpointSpec.Url` alternative
 - [ ] `PlainIdentitySpec.CelExpression` 
 - [ ] ... and moving this to `v1beta3`

@guicassolato am I missing something else?

pkg/evaluators/authorization/kubernetes_authz.go
```diff
@@ -63,26 +63,61 @@ func (k *KubernetesAuthz) Call(pipeline auth.AuthPipeline, ctx gocontext.Context
 	}
 
 	authJSON := pipeline.GetAuthorizationJSON()
-	jsonValueToStr := func(value json.JSONValue) string {
-		return fmt.Sprintf("%s", value.ResolveFor(authJSON))
+	jsonValueToStr := func(value expressions.Value) (string, error) {
+		if value == nil {
+			return "", nil
```
alexsnaps, line 68: Ugh... Now that it is a pointer, we could get `nil` indeed... I don't know whether all these cases are now covered, hard to tell what can or not be `nil` in those specs

install/rbac/role.yaml
alexsnaps: Could you try and run it on your machine and push the changes to this branch please? I have no idea what's screwed up on my end 🤷

guicassolato: Done. https://github.com/Kuadrant/authorino/pull/495/commits/28e7fb976c96937231960c488996a2f4f99f3d10

```
❯ git st
On branch cel
Your branch is up to date with 'origin/cel'.

nothing to commit, working tree clean

❯ make generate manifests
go mod tidy
go: downloading github.com/envoyproxy/go-control-plane v0.12.1-0.20240621013728-1eb8caab5155
go: downloading github.com/prometheus/client_golang v1.20.2
go: downloading github.com/open-policy-agent/opa v0.68.0
go: downloading google.golang.org/grpc v1.66.0
go: downloading go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.53.0
go: downloading github.com/google/cel-go v0.21.0
go: downloading google.golang.org/genproto/googleapis/rpc v0.0.0-20240701130421-f6361c86f094
go: downloading github.com/klauspost/compress v1.17.9
go: downloading golang.org/x/time v0.6.0
go: downloading github.com/stoewer/go-strcase v1.2.0
go: downloading cloud.google.com/go/compute/metadata v0.3.0
go: downloading github.com/golang/glog v1.2.1
go: downloading golang.org/x/tools v0.21.1-0.20240508182429-e35e4ccd0d2d
go: downloading cloud.google.com/go/compute v1.23.0
go: downloading cloud.google.com/go v0.34.0
go mod vendor
controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
/Library/Developer/CommandLineTools/usr/bin/make fmt vet
go fmt ./...
go vet ./...
controller-gen crd:crdVersions=v1 rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=install/crd output:rbac:artifacts:config=install/rbac && /Users/guilherme.cassolato/Workspace/go/src/github.com/kuadrant/authorino/bin/kustomize build install > /Users/guilherme.cassolato/Workspace/go/src/github.com/kuadrant/authorino/install/manifests.yaml
/Library/Developer/CommandLineTools/usr/bin/make patch-webhook
envsubst \
			< /Users/guilherme.cassolato/Workspace/go/src/github.com/kuadrant/authorino/install/manifests.yaml \
			> /Users/guilherme.cassolato/Workspace/go/src/github.com/kuadrant/authorino/install/manifests.yaml.tmp && \
	mv /Users/guilherme.cassolato/Workspace/go/src/github.com/kuadrant/authorino/install/manifests.yaml.tmp /Users/guilherme.cassolato/Workspace/go/src/github.com/kuadrant/authorino/install/manifests.yaml

❯ git st
On branch cel
Your branch is up to date with 'origin/cel'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   install/manifests.yaml
	modified:   install/rbac/role.yaml

no changes added to commit (use "git add" and/or "git commit -a")

❯ grep "controller-gen@" Makefile
	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.15.0)

❯ ./bin/controller-gen --version
Version: v0.15.0
```

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1,5 +1,4 @@
 //go:build !ignore_autogenerated
-// +build !ignore_autogenerated
```
guicassolato, line 2: The `// +build` syntax has been deprecated since Go v1.17.

https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

api/v1beta3/auth_config_types.go
```diff
@@ -0,0 +1,897 @@
...
+type AuthResponseMethod int8
+type CallbackMethod int8
+type CredentialsType int8
+
+type StatusConditionType string
+
+// AuthConfig is the schema for Authorino's AuthConfig API
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
```
guicassolato, line 94: Shall we remove it from `v1beta2`?

api/v1beta3/auth_config_conversion.go
```diff
@@ -0,0 +1,4 @@
+package v1beta3
+
+// Hub marks this version as a conversion hub.
+func (a *AuthConfig) Hub() {}
```
guicassolato, line 4: I don't know for sure how this works, but can we have 2 versions marked as conversion hub? I believe right now `v1beta2` is also marked as hub, right?

api/v1beta3/auth_config_types.go
guicassolato: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions

alexsnaps:
@KevFan could you have a look at this wrt the `v1beta3` changes... my environment is acting up on these generated changes... 🙏

KevFan:
@alexsnaps Sure, I'll take a look 👀

KevFan:
@alexsnaps when running `make manifests generate`, I got a diff compared to yours. I've a PR at https://github.com/Kuadrant/authorino/pull/499 to merge to this PR if you want to merge it into this or take the changes or if you want me to push directly to this PR 🤔 Whichever works best for you anyway 👍

tests/v1beta3/authconfig.yaml
```diff
@@ -8,13 +8,9 @@ spec:
 
   patterns:
     admin-path:
-    - selector: context.request.http.path
-      operator: matches
-      value: ^/admin(/.*)?$
+    - predicate: request.http.path.matches("^/admin(/.*)?$")
```
KevFan, line 11: Out of curiosity, are these changes still WIP for this file? 🤔 
Just wondering cause running `make e2e` is erroring due to invalid AuthConfig with this since it's not expecting `predicate` currently:
- https://github.com/Kuadrant/authorino/blob/3925889e8753a51592b169535c1e5bc36a630afc/api/v1beta3/auth_config_types.go#L118
- https://github.com/Kuadrant/authorino/blob/3925889e8753a51592b169535c1e5bc36a630afc/api/v1beta3/auth_config_types.go#L152-L165
```log
Error from server (BadRequest): error when creating "/Users/chfan/dev/authorino/tests/v1beta3/authconfig.yaml": AuthConfig in version "v1beta3" cannot be handled as a AuthConfig: strict decoding error: unknown field "spec.patterns.admin-path[0].predicate", unknown field "spec.patterns.resource-path[0].predicate"
waiting authconfig readyError from server (NotFound): authconfigs.authorino.kuadrant.io "e2e-test" not found
.Error from server (NotFound): authconfigs.authorino.kuadrant.io "e2e-test" not found
```
And also, failing on the `oneOf` validations such as:
- https://github.com/Kuadrant/authorino/blob/3925889e8753a51592b169535c1e5bc36a630afc/install/crd/patches/oneof_in_authconfigs.yaml#L305-L321

```log
The AuthConfig "e2e-test" is invalid: 
* <nil>: Invalid value: "": "spec.response.success.headers.wristband.when[0]" must validate one and only one schema (oneOf). Found none valid
* spec.response.success.headers.wristband.when[0].patternRef: Required value
* <nil>: Invalid value: "": "spec.authentication.anonymous.when[0]" must validate one and only one schema (oneOf). Found none valid
* spec.authentication.anonymous.when[0].patternRef: Required value
* <nil>: Invalid value: "": "spec.authentication.anonymous.when[1]" must validate one and only one schema (oneOf). Found none valid
* spec.authentication.anonymous.when[1].patternRef: Required value
* <nil>: Invalid value: "": "spec.authorization.admin-kubernetes-rbac.when[1]" must validate one and only one schema (oneOf). Found none valid
* spec.authorization.admin-kubernetes-rbac.when[1].patternRef: Required value
* <nil>: Invalid value: "": "spec.authorization.admin-jwt-rbac.when[1]" must validate one and only one schema (oneOf). Found none valid
* spec.authorization.admin-jwt-rbac.when[1].patternRef: Required value
* <nil>: Invalid value: "": "spec.authorization.admin-jwt-rbac.patternMatching.patterns[0]" must validate one and only one schema (oneOf). Found none valid
* spec.authorization.admin-jwt-rbac.patternMatching.patterns[0].patternRef: Required value
waiting authconfig readyError from server (NotFound): authconfigs.authorino.kuadrant.io "e2e-test" not found
```

alexsnaps: Darn, looks like I screwed up my rebasing then 🤦

alexsnaps: Actually... no! I made this fail on purpose, I need @guicassolato 's opinion here... cause adding CEL there is either recursive _or_ a breaking change, or needs another `NamedCelExpression` mechanism, or...

alexsnaps: So, I'm thinking of adding a `celFunctions` at the same level as 

```go
	NamedPatterns map[string]PatternExpressions `json:"patterns,omitempty"`
```

Same "pattern", but `map[string]CelExpression` that then register the key as a fn() in cel. So that:

```yaml
celFunctions:
  userGroup: request.headers['someHeader']
```

become usable in a cel expression: `userGroup().startWith("admin")`.

If we think that makes sense, I'd add this in a subsequent PR tho and "just get rid of `NamedPattern`s equivalence for Cel for now... wdyt?

pkg/evaluators/identity.go
```diff
@@ -199,11 +199,13 @@ func (config *IdentityConfig) ResolveExtendedProperties(pipeline auth.AuthPipeli
 	authJSON := pipeline.GetAuthorizationJSON()
 
 	for _, extendedProperty := range config.ExtendedProperties {
-		resolved, err := extendedProperty.ResolveFor(extendedIdentityObject, authJSON)
-		if err != nil {
-			return nil, err
+		if extendedProperty.Value != nil {
```
alexsnaps, line 202: This took me forever to hunt down... Looks like there's always been uninitialized properties in there:

```
{"Name":"","Value":{"Static":null,"Pattern":""},"Overwrite":false},{"Name":"","Value":{"Static":null,"Pattern":""},"Overwrite":false}```
```

Which now have `"Value":nil` of course... So... I know I don't know. I _think_ this has been a bug in the making all along, but I don't know for sure... Someone with real deep knowledge and history of this might be able to make a call. I guess this is due to how the "slice" (which is effectively an array then... but I am no Go expert) is constructed [here](https://github.com/Kuadrant/authorino/pull/495/commits/a64198b704cecf80f6b1217c5a4aafadf5099ee0#diff-75405a152218cb714e868d6fe4cc86674992e73c8c0e0a9bd83fd3c475c4b7e0R184)

alexsnaps:
@KevFan Thanks for your help fixing the generated stuff... but I fear I could still use either your or @guicassolato 's help.
Integration tests fail still:

```
Test failed [#5]:
  GET   http://talker-api.127.0.0.1.nip.io:8000/greetings/1
  Authorization: Bearer eyJhbGciOiJSUzI1NiIsImtpZCI6InpTZThzUGx3YXIwOVhCc2RITTd4cU45YU1iY09uNEdMRnoyWlRUckFfMkkifQ.eyJhdWQiOlsiaHR0cHM6Ly9rdWJlcm5ldGVzLmRlZmF1bHQuc3ZjLmNsdXN0ZXIubG9jYWwiXSwiZXhwIjoxNzI5ODY3NjE3LCJpYXQiOjE3Mjk4NjcwMTcsImlzcyI6Imh0dHBzOi8va3ViZXJuZXRlcy5kZWZhdWx0LnN2Yy5jbHVzdGVyLmxvY2FsIiwia3ViZXJuZXRlcy5pbyI6eyJuYW1lc3BhY2UiOiJhdXRob3Jpbm8iLCJzZXJ2aWNlYWNjb3VudCI6eyJuYW1lIjoiYXBwLTEtc2EiLCJ1aWQiOiIwNDRmNWU4Mi01MDQxLTQwOWMtYjdlNy02YjkyZTY1YmNmYTQifX0sIm5iZiI6MTcyOTg2NzAxNywic3ViIjoic3lzdGVtOnNlcnZpY2VhY2NvdW50OmF1dGhvcmlubzphcHAtMS1zYSJ9.oCZ8tqz2mo0QbxwGulyhPz0WKdghsoWOYv3-awh_mtBbcL0ERpxfHg4OyoVnTLj2TvLGDx9ntKt-ag17zHYm8mUgdCQNA7qFFcb1DkN4CFnHkNMn7rwx7Bnv4CwyCbjNM79R5GL3jJu7X55I9PgskStFsXw_LKdV7ttn1Gw2TexYM3QLHjSc8ZXTDFAFoBc8H-19WWFsHRE3zC8erpvAqXzneCBICa9xrbEBOri62teA-yP9bcDp32ROa3MukaYq5JmH79X83Yj03ApG4lD77AiCrokPu2Px4DEUAZRMUSrFwJj7-NcS1ZXGwOh1svSN47HaTOb03j8a8Pn9qksDmg
  X-Forwarded-For: 109.69.200.56

  Expected: 403
  Actual: 200


FAIL
make: *** [Makefile:162: e2e] Error 1
```

But if I then rerun it:

```
$  ./tests/e2e-test.sh
deployment.apps/coredns condition met
deployment.apps/ip-location unchanged
service/ip-location unchanged
deployment.apps/authorino condition met
deployment.apps/envoy condition met
deployment.apps/ip-location condition met
deployment.apps/keycloak condition met
deployment.apps/talker-api condition met
waiting keycloak ready condition met
waiting oidc config ready. condition met
testing invalid authconfig [SUCCESS]
testing valid authconfig
authconfig.authorino.kuadrant.io/e2e-test configured
secret/oauth2-token-introspection-credentials-keycloak configured
secret/talker-api-uma-credentials configured
secret/wristband-signing-key configured
secret/bob-api-key configured
secret/alice-api-key configured
serviceaccount/app-1-sa unchanged
serviceaccount/app-2-sa unchanged
clusterrole.rbac.authorization.k8s.io/talker-api-admin unchanged
clusterrolebinding.rbac.authorization.k8s.io/talker-api-admins configured
waiting authconfig ready. condition met
sending multiple requests
.................................................

SUCCESS
````

They succeed. I guess this some caching issue somewhere... I'll keep on investigating, but a second set of eyes would be appreciated. This is much deeper into the inner working of Authorino than I had hoped for...

File: tests/v1beta3/authconfig.yaml [hunk 8]
```diff
@@ -130,27 +115,21 @@ spec:
           }
     admin-kubernetes-rbac:
       when:
-      - patternRef: admin-path
-      - selector: auth.identity.kubernetes-rbac
-        operator: eq
-        value: 'true'
+      - predicate: request.http.path.matches("^/admin(/.*)?$")
+      - predicate: auth.identity["kubernetes-rbac"]
```
guicassolato, line 119: It looks like this predicate is failing.

guicassolato: Well... _may_ be failing. Or at least flaky.

alexsnaps: But ... then succeeds?

guicassolato: Hmm. Maybe not. The e2e test that failed was a GET request to `/greetings/1` authenticating by K8s SA token.

Could this other predicate be the one to blame `request.http.path.matches("^/admin(/.*)?$")`? 🤔

guicassolato: Could it be that it failed to [obtain a token from the cluster](https://github.com/Kuadrant/authorino/blob/1bea62698b1e9a3485a87f17106cece0ee1ac03f/tests/e2e-test.sh#L119) by setting `$access_token` to some non-empty value that nonetheless is not a valid token, then missed all authentication methods, defaulting to `anonymous` (`priority`: 1) and thus skipping the authorisation rules that otherwise would be enforced when `auth.identity["kubernetes-rbac"]`?

I'm now inclined to blame on the e2e tests themselves, I guess. It's badly written. It shouldn't accept _any_ string as a valid access token ready to send the request.

alexsnaps:
After much headache, I think I found the issue: None of the `expression`s are being evaluated _and_ I get these warnings about them: 

```
authorino.KubeAPIWarningLogger  unknown field "spec.authentication.anonymous.defaults.username.Expression"
authorino.KubeAPIWarningLogger  unknown field "spec.authentication.api-key.defaults.kubernetes-rbac.Expression"
```

So pretty sure there is something I didn't do to declare these fields and most importantly have them evaluated!
@guicassolato @KevFan Or anyone with k8s magic could help me with this?

guicassolato (commented):
> After much headache, I think I found the issue: None of the `expression`s are being evaluated _and_ I get these warnings about them:
> 
> ```
> authorino.KubeAPIWarningLogger  unknown field "spec.authentication.anonymous.defaults.username.Expression"
> authorino.KubeAPIWarningLogger  unknown field "spec.authentication.api-key.defaults.kubernetes-rbac.Expression"
> ```
> 
> So pretty sure there is something I didn't do to declare these fields and most importantly have them evaluated! @guicassolato @KevFan Or anyone with k8s magic could help me with this?

I think you need to give the fields where the `CelExpression` type is used proper JSON names and make `CelExpression` a type of `string` instead of a complex object.

alexsnaps:
[e2e tests](https://github.com/Kuadrant/authorino/actions/runs/11577988126) now pass as well... 🎉

alexsnaps:
This below would let us get rid of all the remainder selector et al stuff: 

```diff
diff --git a/pkg/expressions/cel/expressions.go b/pkg/expressions/cel/expressions.go
index 18ba5728..40fd95b9 100644
--- a/pkg/expressions/cel/expressions.go
+++ b/pkg/expressions/cel/expressions.go
@@ -9,6 +9,7 @@ import (
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/checker/decls"
 	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/ext"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/structpb"
@@ -130,6 +131,7 @@ func Compile(expression string, expectedType *cel.Type, opts ...cel.EnvOption) (
 		decls.NewConst(RootDestinationBinding, decls.NewObjectType("google.protobuf.Struct"), nil),
 		decls.NewConst(RootAuthBinding, decls.NewObjectType("google.protobuf.Struct"), nil),
 	)}, opts...)
+	envOpts = append(envOpts, ext.Strings())
 	env, env_err := cel.NewEnv(envOpts...)
 	if env_err != nil {
 		return nil, env_err
diff --git a/pkg/expressions/cel/expressions_test.go b/pkg/expressions/cel/expressions_test.go
index 587ac4cd..6154d020 100644
--- a/pkg/expressions/cel/expressions_test.go
+++ b/pkg/expressions/cel/expressions_test.go
@@ -42,3 +42,14 @@ func TestPredicate(t *testing.T) {
 	assert.NilError(t, err)
 	assert.Equal(t, response, true)
 }
+
+func TestExpression(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	expression, err := NewStringExpression(`"hello hello".replace("", "_")`)
+	assert.NilError(t, err)
+	resolveFor, err := expression.ResolveFor("{}")
+	assert.NilError(t, err)
+	assert.Equal(t, resolveFor, "_h_e_l_l_o_ _h_e_l_l_o_")
+}
diff --git a/tests/v1beta3/authconfig.yaml b/tests/v1beta3/authconfig.yaml
index 509b93fe..554f2f62 100644
--- a/tests/v1beta3/authconfig.yaml
+++ b/tests/v1beta3/authconfig.yaml
@@ -77,10 +77,10 @@ spec:
           Accept:
             value: application/json
         method: GET
-        url: http://ip-location.authorino.svc.cluster.local:3000/{context.request.http.headers.x-forwarded-for.@extract:{"sep":","}}
+        urlExpression: "http://ip-location.authorino.svc.cluster.local:3000/" + request.headers["x-forwarded-for"].split(",")[0]
       cache:
         key:
-          selector: request.http.headers.x-forwarded-for.@extract:{"sep":","}
+          selector: request.headers["x-forwarded-for"].split(",")[0]
     user-info:
       userInfo:
         identitySource: keycloak
@@ -179,7 +179,7 @@ spec:
               uri:
                 expression: request.path
               scope:
-                selector: request.http.method.@case:lower
+                expression: request.http.method.lowerAscii()
             signingKeyRefs:
             - name: wristband-signing-key
               algorithm: ES256
```

pkg/expressions/cel/expressions.go
```diff
@@ -0,0 +1,195 @@
...
+	input, err := AuthJsonToCel(json)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return e.program.Eval(input)
+}
+
+func (e *Expression) EvaluateStringValue(json string) (string, error) {
+	fmt.Printf("   ====>>>  data: %s\n", json)
```
guicassolato, line 111:
```suggestion
```

pkg/expressions/cel/expressions.go
```diff
@@ -0,0 +1,195 @@
...
+	result, _, err := e.Evaluate(json)
+	if err != nil {
+		return nil, err
+	}
+
+	if jsonVal, err := ValueToJSON(result); err != nil {
+		return nil, err
+	} else {
+		return jsonVal, nil
+	}
```
guicassolato, line 94:
```suggestion
	return ValueToJSON(result)
```

controllers/auth_config_controller.go
```diff
@@ -552,16 +643,50 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
...
+	var err error
+	if user.Expression != "" {
+		if strValue, err = cel.NewStringExpression(string(user.Expression)); err != nil {
+			return nil, err
+		}
+	} else {
+		strValue = &json.JSONValue{Static: user.Value, Pattern: user.Selector}
+	}
+	return strValue, nil
+}
```
guicassolato, line 688: Is there any difference between these two funcs?

controllers/auth_config_controller.go
```diff
@@ -272,7 +284,15 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
 			}
 
 		case api.PlainIdentityAuthentication:
-			translatedIdentity.Plain = &identity_evaluators.Plain{Pattern: identity.Plain.Selector}
+			if identity.Plain.Expression != "" {
+				expression, err := cel.NewStringExpression(string(identity.Plain.Expression))
```
guicassolato, line 288: Not sure if understand `NewStringExpression` here. Does this mean `plain.expression` will always resolve to a string?

With `JSONValue`, I think `plain.selector` can resolve to any type, including returning complex objects to set `auth.identity`, no?

controllers/auth_config_controller.go
```diff
@@ -179,24 +181,30 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
...
 
 	for identityCfgName, identity := range authConfigIdentityConfigs {
-		extendedProperties := make([]evaluators.IdentityExtension, len(identity.Defaults)+len(identity.Overrides))
+		extendedProperties := make([]evaluators.IdentityExtension, 0)
 		for propertyName, property := range identity.Defaults {
-			extendedProperties = append(extendedProperties, evaluators.NewIdentityExtension(propertyName, json.JSONValue{
-				Static:  property.Value,
-				Pattern: property.Selector,
-			}, false))
+			if value, err := stringValueFrom(&property); err != nil {
```
guicassolato, line 186: Same as https://github.com/Kuadrant/authorino/pull/495/files#r1821989655. The value of an extended property can be of any type.

controllers/auth_config_controller.go
alexsnaps: Typo! This the latter should create a `NewExpression`! Tho I wonder if we should keep that one altogether... (by the current state of things, I'd say no!)

controllers/auth_config_controller.go
alexsnaps: So, what would be `plain.selector` value's type for `auth.identity` if one would be to select that? In Golang's type system that is...

controllers/auth_config_controller.go
alexsnaps: Also, just to be clear (and we may want to rename that function), this does return a "JSON literal", but for `String` returns a plain string... i.e. w/o the quotes.

controllers/auth_config_controller.go
alexsnaps: from gjson

```go
func (t Result) Value() interface{} {
	if t.Type == String {
		return t.Str
	}
	switch t.Type {
	default:
		return nil
	case False:
		return false
	case Number:
		return t.Num
	case JSON:
		r := t.arrayOrMap(0, true)
		if r.vc == '{' {
			return r.oi
		} else if r.vc == '[' {
			return r.ai
		}
		return nil
	case True:
		return true
	}
}
```

From gjson... So a `map[string]interface{}`, where `interface` is one of the above recursively I guess...

alexsnaps: One more question tho... what is that good for tho? I realize I took a shortcut here, and even left a note to make this better, but... we're just marshalling json back and forth, right? This only ever gets used thru `AuthPipeline.GetAuthorizationJSON`, no? i.e. _as JSON_?

alexsnaps: I think this is where the mismatch happens: 

```
	ResolveFor(jsonData string) (interface{}, error)
```

`ResolveFor` is the thing that only ever get to interact with these payloads, is that correct? So we end up spending all our CPU cycles marshaling that data back and forth, are we not? Am I missing something?

api/v1beta3/auth_config_types.go
```diff
@@ -402,6 +410,8 @@ type PlainIdentitySpec struct {
 	// Any pattern supported by https://pkg.go.dev/github.com/tidwall/gjson can be used.
 	// The following Authorino custom modifiers are supported: @extract:{sep:" ",pos:0}, @replace{old:"",new:""}, @case:upper|lower, @base64:encode|decode and @strip.
 	Selector string `json:"selector"`
```
guicassolato, line 412:
```suggestion
	Selector string `json:"selector,omitempty"`
```

api/v1beta3/auth_config_types.go
```diff
@@ -439,6 +449,8 @@ type HttpEndpointSpec struct {
 	// E.g. https://ext-auth-server.io/metadata?p={request.path}
 	Url string `json:"url"`
```
guicassolato, line 450:
```suggestion
	Url string `json:"url,omitempty"`
```

controllers/auth_config_controller.go
alexsnaps: > > So, what would be `plain.selector` value's type for `auth.identity` if one would be to select that? In Golang's type system that is...
> 
> Not sure if I follow. `plain.selector` is not for selecting _from_ `auth.identity`; it selects for setting the value _of_ `auth.identity`. What it selects can be of any type.

Not going to elaborate on this... as it's orthogonal and I found the answer, as [per this answer](https://github.com/google/cel-spec/blob/master/tests/simple/testdata/macros.textproto#L209-L213). But the follow up questions remain. I mean I can go thru the whole process of remapping to golang native types... to only marshal them back into JSON literals later... but it makes no sense, other than something _needing to use that in golang_. Tho even then I'd argue it might be better to optimize for the most frequent data access, which seems to be JSON based of my scanning of the code base. 

I do think that, if we ever 🔥 the gjson stuff, as per my comment in the code, ideally the pipeline stores all that data as protobuf... which would make it much lighter weight to inject it into the CEL interpreter. 

So _I can_ have the resulting value of a CEL expression be returned from `ResolveFor`, be expressed in golang's type system, but I really think this isn't sensible... as, again afaict! but I could be missing something, nothing actually needs that, other than to re marshal it back into JSON... 

Is my reasoning sensible? Am I missing something that renders it inapplicable to the problem space?

alexsnaps: Also, I think `ResolveFor(jsonData string) (interface{}, error)` should probably _not_ expect json in and golang out, but have a clear contract, i.e. `json string` in & out _or_ `interface{}` in & out... It's not really symmetrical. But _not arguing to fix this as part of this PR or even now neither_

alexsnaps: For future us... Not sensible for this PR I think

<details>

Thinking out loud further... 

`ResolveFor` should probably be `ResolveFor(pipeline *AuthPipeline) (interface{}, error)`. 
That now would let us control what the expression takes in as argument _format_. gjson stuff can call `GetAuthorizationJSON()`, while CEL could bypass it go straight to the protobuf format. I'd still argue for storing things in protobuf tho... So in order to remain "compatible", the result would be... "undefined" which probably isn't good... 🤔 

</details>

guicassolato: Let me show I mean with an example...

Requisite:
1. Add the `omitempty` suggested at https://github.com/Kuadrant/authorino/pull/495#discussion_r1822380561

Run the service:

```sh
export LOG_LEVEL=debug; make cluster install run
```

Create a couple of AuthConfigs expected to behave exactly the same:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: gjson
spec:
  hosts: ["gjson"]
  authentication:
    "envoy-verified-jwt":
      plain:
        selector: metadata.filter_metadata.envoy\.filters\.http\.jwt_authn|verified_jwt
  authorization:
    "must_be_admin":
      patternMatching:
        patterns:
          operator: incl
          value: admin
---
kind: AuthConfig
metadata:
  name: cel
spec:
  hosts: ["cel"]
  authentication:
    "envoy-verified-jwt":
      plain:
        expression: |
          metadata.filter_metadata["envoy.filters.http.jwt_authn"].verified_jwt
  authorization:
    "must_be_admin":
      patternMatching:
        patterns:
        - predicate: |
            "admin" in auth.identity.roles
EOF
```

Send a request to the AuthConfig using `plain.selector`:

```sh
grpcurl -plaintext -d @ localhost:50051 envoy.service.auth.v3.Authorization.Check <<EOF
{
  "attributes": {
    "request": {
      "http": {
        "host": "gjson"
      }
    },
    "metadata_context": {
      "filter_metadata": {
        "envoy.filters.http.jwt_authn": {
          "verified_jwt": {
            "aud": "toystore",
            "exp": 1730284979,
            "iat": 1730371379,
            "iss": "auth-server.io",
            "sub": "55474041-c8c6-48bb-9398-34c2dbfd88b3",
            "roles": ["admin"]
          }
        }
      }
    }
  }
}
EOF
```
```
{
  "status": {},
  "okResponse": {},
  "dynamicMetadata": {}
}
```

Now send a request to the AuthConfig using `plain.expression`:

```sh
grpcurl -plaintext -d @ localhost:50051 envoy.service.auth.v3.Authorization.Check <<EOF
{
  "attributes": {
    "request": {
      "http": {
        "host": "cel"
      }
    },
    "metadata_context": {
      "filter_metadata": {
        "envoy.filters.http.jwt_authn": {
          "verified_jwt": {
            "aud": "toystore",
            "exp": 1730284979,
            "iat": 1730371379,
            "iss": "auth-server.io",
            "sub": "55474041-c8c6-48bb-9398-34c2dbfd88b3",
            "roles": ["admin"]
          }
        }
      }
    }
  }
}
EOF
```
```
{
  "status": {
    "code": 7
  },
  "deniedResponse": {
    "status": {
      "code": "Forbidden"
    },
    "headers": [
      {
        "header": {
          "key": "X-Ext-Auth-Reason",
          "value": "no such key: roles"
        }
      }
    ]
  }
}
```

If you check the logs, both requests succeed in the authentication, however...

The first request evaluates authorisation based on the following Authorization JSON:

```json
{
  "auth": {
    "identity": {
      "aud": "toystore",
      "exp": 1730284979,
      "iat": 1730371379,
      "iss": "auth-server.io",
      "roles": [
        "admin"
      ],
      "sub": "55474041-c8c6-48bb-9398-34c2dbfd88b3"
    }
  },
  "context": {
    "metadata_context": {
      "filter_metadata": {
        "envoy.filters.http.jwt_authn": {
          "verified_jwt": {
            "aud": "toystore",
            "exp": 1730284979,
            "iat": 1730371379,
            "iss": "auth-server.io",
            "roles": [
              "admin"
            ],
            "sub": "55474041-c8c6-48bb-9398-34c2dbfd88b3"
          }
        }
      }
    },
    "request": {
      "http": {
        "host": "gjson",
        "id": "b59827c5-ec62-422c-972c-b86f1d29ec52"
      }
    }
  },
  "destination": {},
  "metadata": {
    "filter_metadata": {
      "envoy.filters.http.jwt_authn": {
        "verified_jwt": {
          "aud": "toystore",
          "exp": 1730284979,
          "iat": 1730371379,
          "iss": "auth-server.io",
          "roles": [
            "admin"
          ],
          "sub": "55474041-c8c6-48bb-9398-34c2dbfd88b3"
        }
      }
    }
  },
  "request": {
    "host": "gjson",
    "id": "b59827c5-ec62-422c-972c-b86f1d29ec52"
  },
  "source": {}
}
```

The second request evaluates authorisation based on the following Authorization JSON:

```json
{
  "auth": {
    "identity": "{\"aud\":\"toystore\", \"exp\":1730284979, \"iat\":1730371379, \"iss\":\"auth-server.io\", \"roles\":[\"admin\"], \"sub\":\"55474041-c8c6-48bb-9398-34c2dbfd88b3\"}"
  },
  "context": {
    "metadata_context": {
      "filter_metadata": {
        "envoy.filters.http.jwt_authn": {
          "verified_jwt": {
            "aud": "toystore",
            "exp": 1730284979,
            "iat": 1730371379,
            "iss": "auth-server.io",
            "roles": [
              "admin"
            ],
            "sub": "55474041-c8c6-48bb-9398-34c2dbfd88b3"
          }
        }
      }
    },
    "request": {
      "http": {
        "host": "cel",
        "id": "04eb5f71-ec3c-40f1-a941-a7a5f0b3390b"
      }
    }
  },
  "destination": {},
  "metadata": {
    "filter_metadata": {
      "envoy.filters.http.jwt_authn": {
        "verified_jwt": {
          "aud": "toystore",
          "exp": 1730284979,
          "iat": 1730371379,
          "iss": "auth-server.io",
          "roles": [
            "admin"
          ],
          "sub": "55474041-c8c6-48bb-9398-34c2dbfd88b3"
        }
      }
    }
  },
  "request": {
    "host": "cel",
    "id": "04eb5f71-ec3c-40f1-a941-a7a5f0b3390b"
  },
  "source": {}
}
```

guicassolato (approved):
Amazing work, @alexsnaps!

I've manually tried this against a couple use cases myself and it seems to work like a charm.

Issue #500, opened 2024-11-02
Well-known attribute `request.time` is not really usable as it


alexsnaps:
This has become more painful because of the introduction of CEL, but `request.time` is made available as the struct that backs it:
`map[nanos:68000 seconds:1.730319299e+09]`. With CEL having first class support it is probably expected (and documented) to be available as a proper `Timestamp`. Even in JSON tho, it would be probably be best represented in RFC3339 format.

## Proposal

[`protojson`](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#pkg-types) deals with these conversions properly. Which is already used to produce the intermediary JSON result coming back from a CEL Value, as well as feeding out `WellKnownAttributes` into the interpreter. `request` itself _is protobuf_ as it comes from `envoy/service/auth/v3`'s `AttributeContext` which is autogenerated from a `.proto`. 

I'm thinking we should have our own `WellKnownAttributes` spec'ed as protobuf and use `protojson` to marshal that data. With that approach, no json marshaling would actually to inject the `WellKnownAttributes` into the CEL interpreter, but they would be made available as protobuf `Message`s directly. 

So the proposal is to refactor all of `WellKnownAttributes` into `.proto`, and use that generated code to produce proper typed JSON/CEL Values. wdyt?

alexsnaps:
As part of the refactoring `Value.ResolveFor` would be changed to _not_ take a JSON literal in anymore, but the `WellKnownAttributes` directly... so they still can be "cached" (kept around) on the call site, but the callee would now be the one to decide how to marshal that data...

```go
type Value interface {
	ResolveFor(attributes *WellKnownAttributes) (interface{}, error)
}

// where WellKnownAttributes are obtained from the AuthPipeline, just as the JSON currently is with GetAuthorizationJSON
value.ResolveFor(authPipeline.GetWellKnowAttributes)
```

alexsnaps:
Also note that currently, because of how data "gets out of CEL", this

```go
exp, err := NewExpression("timestamp(\"2024-12-25T12:00:00Z\") + duration('123456m')")
v, err := exp.ResolveFor("{}") 
// v is a `string` of value `2025-03-21T05:36:00Z` , i.e. a RFC3339 date format
```

PR #501, opened 2024-11-04
fix: RawExtension to string conversion

guicassolato:
Fixes conversion of `ValueOrSelector.Value` (based on [`runtime.RawExtension`](https://pkg.go.dev/k8s.io/apimachinery@v0.28.3/pkg/runtime#RawExtension)) to string, used at the following configs:
- SubjectAccessReview authorization
- External HTTP requests (metadata, external Rego policies, etc)
- SpiceDB check permissions

This bug was introduced in [v0.18.0](https://github.com/Kuadrant/authorino/releases/tag/v0.18.0), when the AuthConfig controller was updated to work with `v1beta2` type (previously (`v1beta1`), thus activating the option to set static values to JSON/YAML types other than strings. The conversion functions used in the features listed above were overlooked in the process and remain naively treating static values (stored in Golang `interface{}` variables) as if they were always strings.

## Verification steps

```sh
make local-setup FF=1
```

```sh
```

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authorization:
    "sar":
        user:
          value: john
        resourceAttributes:
          resource:
            value: secrets
          name:
            selector: request.path.@extract:{"sep":"/","pos":1}
          verb:
            expression: request.method
  response:
    unauthorized:
      message:
        value: Access denied by the Kubernetes RBAC
EOF
```

```sh
curl http://talker-api.127.0.0.1.nip.io:8000/my-secret -i
```

Check the Authorino logs. You should spot en entry like the following:

```sj
{
  "level": "debug",
  "ts": "2024-11-04T11:06:44Z",
  "logger": "authorino.service.auth.authpipeline.authorization.kubernetesauthz",
  "msg": "calling kubernetes subject access review api",
  "request id": "b6471552-9154-41fb-87e4-843f5db9255a",
  "subjectaccessreview": {
    "metadata": {
      "creationTimestamp": null
    },
    "spec": {
      "resourceAttributes": {
        "verb": "GET",
        "resource": "secrets",
        "name": "my-secret"
      },
      "user": "john"
    },
    "status": {
      "allowed": false
    }
  }
}
```

Before the fix, attributes such as `subjectaccessreview.spec.user` were being stringified as `"{\"john\" <nil>}"`, which is the direct print out of the `RawExtension` type in Golang as string.

### DIFF
--- a/pkg/evaluators/authorization/authzed.go
+++ b/pkg/evaluators/authorization/authzed.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/kuadrant/authorino/pkg/auth"
 	"github.com/kuadrant/authorino/pkg/expressions"
+	"github.com/kuadrant/authorino/pkg/json"
 	"google.golang.org/grpc"
 	insecuregrpc "google.golang.org/grpc/credentials/insecure"
 
@@ -60,10 +61,11 @@ func (a *Authzed) Call(pipeline auth.AuthPipeline, ctx gocontext.Context) (inter
 	if err != nil {
 		return nil, err
 	}
+	permissionStr, err := json.StringifyJSON(permission)
 	resp, err := client.CheckPermission(ctx, &authzedpb.CheckPermissionRequest{
 		Resource:   resource,
 		Subject:    &authzedpb.SubjectReference{Object: object},
-		Permission: fmt.Sprintf("%s", permission),
+		Permission: permissionStr,
 	})
 	if err != nil {
 		return nil, err
@@ -91,12 +93,20 @@ func authzedObjectFor(name, kind expressions.Value, authJSON string) (*authzedpb
 	if err != nil {
 		return nil, err
 	}
+	objectIdStr, err := json.StringifyJSON(objectId)
+	if err != nil {
+		return nil, err
+	}
 	objectType, err := kind.ResolveFor(authJSON)
 	if err != nil {
 		return nil, err
 	}
+	objectTypeStr, err := json.StringifyJSON(objectType)
+	if err != nil {
+		return nil, err
+	}
 	return &authzedpb.ObjectReference{
-		ObjectId:   fmt.Sprintf("%s", objectId),
-		ObjectType: fmt.Sprintf("%s", objectType),
+		ObjectId:   objectIdStr,
+		ObjectType: objectTypeStr,
 	}, nil
 }
--- a/pkg/evaluators/authorization/authzed.go
+++ b/pkg/evaluators/authorization/authzed.go
@@ -62,6 +62,9 @@ func (a *Authzed) Call(pipeline auth.AuthPipeline, ctx gocontext.Context) (inter
 		return nil, err
 	}
 	permissionStr, err := json.StringifyJSON(permission)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/evaluators/metadata/generic_http.go
```diff
@@ -142,7 +142,11 @@ func (h *GenericHttp) buildRequest(ctx gocontext.Context, endpoint, authJSON str
 		if err != nil {
 			return nil, err
 		}
-		req.Header.Set(header.Name, fmt.Sprintf("%s", headerValue))
+		headerValueStr, err := json.StringifyJSON(headerValue)
```
alexsnaps, line 145: Is that _always_ true tho? Couldn't this be "just a plain string" ever?

alexsnaps (approved):
Tho... unsure about the last one, as per my comment...

pkg/evaluators/authorization/authzed.go
```diff
@@ -60,10 +61,11 @@ func (a *Authzed) Call(pipeline auth.AuthPipeline, ctx gocontext.Context) (inter
 	if err != nil {
 		return nil, err
 	}
+	permissionStr, err := json.StringifyJSON(permission)
```
KevFan, line 64: Do we care for doing any handling for the `err` here? Since nothing is done with it, it just gets reassigned in the next line 🤔

pkg/evaluators/metadata/generic_http.go
guicassolato: One thing we can do to mitigate the impact of this PR over the other affected sources for the conversion is to make `json.StringifyJSON` return earlier if the `interface{}` input is a `string`.

pkg/evaluators/authorization/authzed.go
guicassolato: Fixed.

pkg/evaluators/metadata/generic_http.go
guicassolato: Just pushed a commit: https://github.com/Kuadrant/authorino/pull/501/commits/113717d144f6d3349be47759bf6a63b3bc782e0d

KevFan (approved):
Looks good to me!

pkg/evaluators/metadata/generic_http.go
alexsnaps: Ok, so it was _not_ always the case? And now a "plain string" remains a plain string... do I get this right? In anycase... I think looking into something like what's proposed in #500 wrt to `Value.ResolveFor` signature would be a good refactor for ... future us. Today's `interface{}` (and I know that's sourced in golang's type system) is just too broad I think. Further more the asymmetry of the function wrt its input & output is definitively a code smell to me.

alexsnaps (approved):
🚀

pkg/evaluators/metadata/generic_http.go
alexsnaps: It was the case, I played with these functions a bit... I get what's going on now

PR #507, opened 2024-11-15
fix: add authorizationGroups property to K8s SAR authorization, fixes #506

### LINKED ISSUE #506: Kubernetes SubjectAccessReview authz doesn't support using user's groups from Kubernetes TokenReview authn
**Describe the bug**
Kubernetes TokenReview Authn retrieves user group information. This group information is ignored when making Kubernetes SubjectAccessReview (SAR) call for authz. However, the API doc for SAR suggests that the groups passed in that call should be the list of user's groups. Otherwise it acts as an OR condition. I.e. access is authorized if the user has rolebindings that grant permissions, OR a static list of groups in AuthConfig has rolebindings, irrespective of whether the user is a member of the static list of groups or not. 

**Help us Reproduce it**

Given the following in the namespace `odh-model-registries`:
1. Service `modelregistry-sample`
2. User Group `modelregistry-sample-users`
3. A K8s Role `registry-user-modelregistry-sample` with a GET permission for the service `modelregistry-sample`
4. And the following AuthConfig

```yaml
spec:
  authentication:
    cluster-users:
      credentials:
        authorizationHeader: {}
      kubernetesTokenReview:
        audiences:
        - https://*********
  authorization:
    k8s-rbac:
        groups:
        - modelregistry-sample-users
        resourceAttributes:
          group:
            value: ""
          name:
            value: modelregistry-sample
          namespace:
            value: odh-model-registries
          resource:
            value: services
          subresource:
            value: ""
          verb:
            value: get
        user:
          selector: auth.identity.username
```

The following behavior is observed:

1. Authorization against this service with a User not in the allowed Group modelregistry-sample-users is allowed
2. The call is incorrectly authorized as the Group itself is allowed
3. Remove the static group `modelregistry-sample-users` from AuthConfig above
4. Add User to Group to grant it GET permission
5. Authorization fails despite User having permission to call GET indirectly through `modelregistry-sample-users` Group

**Expected behavior**
AuthConfig should add support for user defined expressions to get group information from a prior AuthN step in the pipeline. Which, will allow end users to create AuthConfig like:

```yaml
spec:
  authentication:
    cluster-users:
      credentials:
        authorizationHeader: {}
      kubernetesTokenReview:
        audiences:
        - https://**********
  authorization:
    k8s-rbac:
        authorizationGroups:
         selector: auth.identity.groups
        resourceAttributes:
          group:
            value: ""
          name:
            value: modelregistry-sample
          namespace:
            value: odh-model-registries
          resource:
            value: services
          subresource:
            value: ""
          verb:
            value: get
        user:
          selector: auth.identity.username
```

- Cluster Version: OpenShift 4.15
- Authorino version: Red Hat - Authorino (Technical Preview) 1.0.2

**Additional context**

The test service and AuthConfig can be be replicated using the opendatahub modelregistry component at https://github.com/opendatahub-io/model-registry-operator. Or, any test service and group along with an AuthConfig similar to above can replicated this issue.

dhirajsb:


api/v1beta3/auth_config_types.go
```diff
@@ -614,6 +614,9 @@ type KubernetesSubjectAccessReviewAuthorizationSpec struct {
 	// Groups the user must be a member of or, if `user` is omitted, the groups to check for authorization in the Kubernetes RBAC.
 	Groups []string `json:"groups,omitempty"`
 
+	// AuthorizationGroups is a value or selector to use as groups to check for authorization in the Kubernetes RBAC.
+	AuthorizationGroups *ValueOrSelector `json:"authorizationGroups,omitempty"`
```
guicassolato, line 618: Perhaps we should deprecate `groups`, as `authorizationGroups` suffices, covering as well for the case of static values.

controllers/auth_config_controller.go
```diff
@@ -503,7 +503,14 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
...
 
-			translatedAuthorization.KubernetesAuthz, err = authorization_evaluators.NewKubernetesAuthz(authorinoUser, authorization.KubernetesSubjectAccessReview.Groups, authorinoResourceAttributes)
+			var authorinoGroups expressions.Value
+			if authorization.KubernetesSubjectAccessReview.AuthorizationGroups != nil {
+				authorinoGroups, err = valueFrom(authorization.KubernetesSubjectAccessReview.AuthorizationGroups)
+				if err != nil {
+					return nil, err
+				}
+			}
+			translatedAuthorization.KubernetesAuthz, err = authorization_evaluators.NewKubernetesAuthz(authorinoUser, authorization.KubernetesSubjectAccessReview.Groups, authorinoGroups, authorinoResourceAttributes)
```
guicassolato, line 513: I was thinking... What about resolving `groups` (if provided) or `authorizationGroups` (if provided) in the control plane?

Instead of passing both to the constructor, the constructor could expect always and only an `expressions.Value`. If `authorizationGroups != nil`, it's straightforward and we pass `valueFrom(authorizationGroups)`. Otherwise and if `len(groups) > 0`, we build a `json.JSONValue` (implements `expressions.Value`) setting by `Static` to `groups`.

api/v1beta3/auth_config_types.go
dhirajsb: I agree, we can deprecate `groups`. What's the process for doing so in Authorino? Is it just documenting it as deprecated?

controllers/auth_config_controller.go
dhirajsb: Thanks @guicassolato, I made the suggested change to accept a single groups arg.

api/v1beta3/auth_config_types.go
guicassolato: Right now, just adding the deprecation notice to the description of the field is good enough. I.e.

```go
	// Groups the user must be a member of or, if `user` is omitted, the groups to check for authorization in the Kubernetes RBAC.
	// Deprecated: Use authorizationGroups instead.
	Groups []string `json:"groups,omitempty"`
```

guicassolato:
> @guicassolato I was thinking back to an authorization requirement we had in managed openshift connectors, where we were authorizing against an org field in JWT, and a groups claim. I'm wondering whether we should support the following scenarios as well:
> 
>     * Groups selector could resolve to a single scalar string
> 
>     * Multiple groups selectors are needed to aggregate a list of groups

@dhirajsb, I think it should be straightforward and match the SAR API, i.e., `authorizationGroups` always resolves to an array of strings.

For cases like the one you described, one can always use CEL. E.g., the following resolves to a list containing a single string equal to the org name used as group.

```yaml
authorizationGroups:
  expression: [auth.identity.org_name]
```

Because it's CEL, we can do fancy stuff. E.g.: `[auth.identity.org_name] + auth.identity.groups`.

guicassolato (dismissed):

api/v1beta3/auth_config_types.go
dhirajsb: I added the suggested doc string.

dhirajsb:
> Because it's CEL, we can do fancy stuff. E.g.: [auth.identity.org_name] + auth.identity.groups.

I was thinking about CEL after I wrote that comment, but wasn't sure how capable it was. Good to know we can cover fairly complex use cases with it. :+1:

api/v1beta3/auth_config_types.go
```diff
@@ -612,6 +612,7 @@ type KubernetesSubjectAccessReviewAuthorizationSpec struct {
 	User *ValueOrSelector `json:"user,omitempty"`
 
 	// Groups the user must be a member of or, if `user` is omitted, the groups to check for authorization in the Kubernetes RBAC.
+	// Deprecated: Use authorizationGroups instead.
```
guicassolato, line 615: Can you please `make manifests` again, so we catch this in the OAS too?

dhirajsb: Sorry about that. Fixed it by running `make manifests`. I also added a check for uncommitted changes in go-test.yaml action.

guicassolato (approved):

.github/workflows/go-test.yaml
```diff
@@ -26,6 +26,17 @@ jobs:
...
+        run: |
+          make generate manifests
+          clean=$(git status --porcelain)
+          if [[ -z "$clean" ]]; then
+              echo "Empty git status --porcelain: $clean"
+          else
+              echo "Uncommitted file changes detected: $clean"
+              git diff
+              exit 1
+          fi
```
guicassolato, line 39: I like this, but maybe we should have it in a separate PR. I don't see it as part of the 'Unit Tests' workflow.

dhirajsb: No problem, should I take it out of this PR and open another one?

guicassolato: Open another one is up to you of course. For now, if you can please take it out of this PR, I think it's cleaner. Thanks!

dhirajsb: I removed that commit from the PR history.

guicassolato (approved):
Good job, @dhirajsb! Thank you so much for this!

Leaving here a few verification steps for the record. We should probably add some docs too at some point.

### Verification steps

Setup the environment:

```sh
make local-setup DEPLOY_KEYCLOAK=1 FF=1

```

In the Keycloak Admin UI (http://keycloak.127.0.0.1.nip.io:8080 | admin:p):

1. Define a `talker-api-members` role in the `demo` client of the `kuadrant` realm
2. Create a Token Mapper of 'User Client Role' type to set the `groups` token claim
3. Map the realm user `john` to the `demo:talker-api-members` role

Create permissions in the Kubernetes RBAC:

```sh
kubectl apply -f -<<EOF
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  namespace: default
  name: reader
rules:
- apiGroups: ["talker-api.127.0.0.1.nip.io"]
  resources: ["messages"]
  verbs: ["get"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: readers
subjects:
- kind: Group
  name: talker-api-members
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: Role
  name: reader
  apiGroup: rbac.authorization.k8s.io
EOF
```

#### Try the deprecated API

Create the AuthConfig:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    keycloak:
      jwt:
  authorization:
    "k8s-rbac":
        groups: # <== this will grant access to all requests given the rolebinding created before; not specifying this would require users to be bound directly to the role (not the group)
        - talker-api-members
        user:
          expression: auth.identity.preferred_username
        resourceAttributes:
          group:
            value: talker-api.127.0.0.1.nip.io
          resource:
            value: messages
          namespace:
            value: default
          verb:
            expression: request.method.lowerAscii()
EOF
```

Consumer the API as a user who _is_ a member of `talker-api-members`:

```sh
ACCESS_TOKEN=$(kubectl run token --attach --rm --restart=Never -q --image=curlimages/curl -- http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=john' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

```sh
# HTTP/1.1 200 OK (due to the SAR request with static groups)
```

Consume the API as a user who is _not_ a member of `talker-api-members`:

```sh
ACCESS_TOKEN=$(kubectl run token --attach --rm --restart=Never -q --image=curlimages/curl -- http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=jane' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

```sh
# HTTP/1.1 200 OK (due to the SAR request with static groups)
```

#### Try the new API

Create the AuthConfig:

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    keycloak:
      jwt:
  authorization:
    "k8s-rbac":
        authorizationGroups: # <== checks whether any of the actual groups the user is a member of has permission
          expression: auth.identity.groups
        user:
          expression: auth.identity.preferred_username
        resourceAttributes:
          group:
            value: talker-api.127.0.0.1.nip.io
          resource:
            value: messages
          namespace:
            value: default
          verb:
            expression: request.method.lowerAscii()
EOF
```

Consume the API as a user who _is_ a member of `talker-api-members`:

```sh
ACCESS_TOKEN=$(kubectl run token --attach --rm --restart=Never -q --image=curlimages/curl -- http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=john' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

```sh
# HTTP/1.1 200 OK
```

Consumer the API as a user who is _not_ a member of `talker-api-members`:

```sh
ACCESS_TOKEN=$(kubectl run token --attach --rm --restart=Never -q --image=curlimages/curl -- http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=jane' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

```sh
# HTTP/1.1 403 Forbidden
```

dhirajsb:
Thanks for testing a built version @guicassolato . I'm also working on testing this locally with a modified model registry. I'll let you know if I run into any issues.

PR #509, opened 2024-11-25
docs: Common Expression Language (CEL)

guicassolato:


docs/features.md
```diff
@@ -250,7 +264,7 @@ spec:
   authentication:
     "pre-validated-jwt":
       plain:
-        selector: context.metadata_context.filter_metadata.envoy\.filters\.http\.jwt_authn|verified_jwt
+        expression: metadata.filter_metadata.envoy['filters.http.jwt_authn'].verified_jwt
```
alexsnaps, line 267: We've seen issues getting to that data in the wasm-shim when directly requesting that from envoy. 
But would that even work on the authorino side? There is no way the entire filter chain's metadata is sent over... is it? Did that _ever_ worked? 

Also, this should be `metadata.filter_metadata['envoy.filters.http.jwt_authn'].verified_jwt`

alexsnaps: /cc @adam-cattermole

guicassolato: It works in Authorino.

guicassolato: > Also, this should be metadata.filter_metadata['envoy.filters.http.jwt_authn'].verified_jwt

Nice catch! Fixing it now. Thanks!

docs/features.md
```diff
@@ -277,14 +291,10 @@ spec:
   authorization:
     "read-only-access-if-authn-fails":
       when:
-        - selector: auth.identity.anonymous
-          operator: eq
-          value: "true"
+      - predicate: auth.identity.anonymous
```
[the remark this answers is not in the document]
guicassolato: Good point. No, it is not _always_ set. In the example, if the other authentication method wins, then there will be no `anonymous` in `auth.identity`.

That wasn't a problem before. I guess now our options are:
1. `has(auth.identity.anonymous) && auth.identity.anonymous`
2. `authentication: { "jwt": { …, defaults: { anonymous: { expression: "false" } } }, "anonymous": {…} }`

docs/features.md
```diff
@@ -475,9 +486,9 @@ authorization:
         resource:
           value: pets # the resource kind
         name:
-          selector: context.request.http.path.@extract:{"sep":"/","pos":2} # resource name – e.g., the {id} in `/pets/{id}`
+          expression: request.path.split('/')[2] # resource name – e.g., the {id} in `/pets/{id}`
```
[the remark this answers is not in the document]
guicassolato: - 0: `` (empty string)
- 1: `pets`
- 2: `{id}`

alexsnaps (approved):
🙏

PR #523, opened 2025-02-26
docs: move using kuadrant from guides into readme for guides

### LINKED ISSUE #519: docs: move "Using Kuadrant" portion in user guides into a README of user-guides directory
# Description
Resulting from a converstaion to decouple Kuadrant `AuthPolicy` from the authorino user guides, it's suggested that we move the "Using Kuadrant" section from the user guides to a `README.md` under `authorino/docs/user-guides` in the github repo since Kuadrant Operator `AuthPolicy` is a consumer of the `AuthConfig` . 

This would link users to Kuadrant if they're considering Gateway API while also separating out that section from the authorino user guides

KevFan:
# Description
Closes: https://github.com/Kuadrant/authorino/issues/519

Move out the "Using Kuadrant" from the user guides to keep the user guides focused on `AuthConfig`. Instead a `README.md` is introduced describing Kuadrant `AuthPolicy` as a consumer of the `AuthConfig`

### DIFF
--- a/docs/user-guides/README.md
+++ b/docs/user-guides/README.md
@@ -15,13 +15,19 @@ to define authentication and authorization rules directly in their network polic
 Rather than defining `AuthConfig` resources directly, users can define an `AuthPolicy`, and Kuadrant will automatically
 generate the corresponding `AuthConfig` for Authorino to consume.
 
-Key details about `AuthPolicy` as an `AuthConfig` Consumer:
+### Key Details About `AuthPolicy` as an `AuthConfig` Consumer
 
 - **Same Specification**: The spec schema of `AuthPolicy` is effectively the same as `AuthConfig`.
 - **Automatic Hostname Handling**: Unlike `AuthConfig`, `AuthPolicy` does not explicitly define `spec.host`. Instead, hostnames are
 inferred from the Kubernetes network object in `spec.targetRef` and the route selectors in the policy.
 - **Authorino Integration**: Kuadrant translates `AuthPolicy` into an `AuthConfig`, which is then processed by Authorino to enforce
 authentication and authorization.
+
+### Authorino Features Not Available via Kuadrant
+Some Authorino-specific features may not be available when using Authorino through Kuadrant:
+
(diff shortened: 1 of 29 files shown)


guicassolato (commented):
While revisiting this, I had doubts about centralising the disclaimer for Kuadrant users in a separate doc that one could easily overlook. However, I now think it's worthwhile the trade-offs, which I'd probably summarise as follows:

**✅ Pros:**

- **Reduced noise for direct Authorino users** - The guides become cleaner and more focused on their core purpose (demonstrating AuthConfig features). The current boxes interrupt the flow twice per guide.
- **Better discoverability of limitations** - The centralised README explicitly lists features NOT available via Kuadrant (Sharding, Host Override), which is harder to communicate in repeated inline boxes.
- **Single source of truth** - Information about Kuadrant integration lives in one place, making it easier to:
  - Keep consistent across all guides
  - Update when the relationship changes
  - Add other AuthConfig consumers in the future
- **Improved maintenance** - Fewer places to update when Kuadrant's API or relationship changes.
- **Scalability** - If other projects become AuthConfig consumers – at the same as Kuadrant, we add one section to README.md instead of modifying tens of other files.

**⚠️ Cons:**

- **Discoverability** (the other side) - Kuadrant users following a specific guide might miss the README and get confused when instructions mention AuthConfig instead of AuthPolicy.
- **Context switch** - Users need to jump to another doc to understand the Kuadrant-specific adaptations. The inline boxes provided just-in-time reminders.
- **First-time user experience** - Someone coming from Kuadrant docs might land on a specific guide via search/link and not see the README context.

If I may suggest only a couple of enhancements before we go ahead and merge this PR:
1. Keep a single brief callout at the top of the guides:
    > **Using AuthConfig via Kuadrant or other tools?** See the [AuthConfig Consumers guide](README.md) for adaptation notes.
2. Ensure the README is prominently featured in the navigation – E.g., add a link to it from the main README of the repo and user guide index page.

If there's anybody interested in taking ownership of this and implement these suggestions, I'll be glad to approve them.

PR #527, opened 2025-04-10
Explicitly state Z stream go version in go.mod, version set to 1.22.5

Patryk-Stefanski:
Same as https://github.com/Kuadrant/authorino-operator/pull/249


guicassolato (approved):
Thanks!

PR #529, opened 2025-04-16
Upgrade Go version to 1.23

guicassolato:


maleck13:
We are stuck at the moment as this is not a version we can use downstream also

eguzki:
> We are stuck at the moment as this is not a version we can use downstream also

yes we can https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690

eguzki (approved):

PR #535, opened 2025-05-28
fix: use ubuntu-24.04 for image build action

KevFan:
# Description
Image build github action is broken as ubuntu-20.04 runner is deprecated and fully unsupported - https://github.com/actions/runner-images/issues/11101


.github/workflows/build-images.yaml
```diff
@@ -16,7 +16,7 @@ env:
 jobs:
   build:
     name: Build and push image
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
```
eguzki, line 19: why not `ubuntu-latest` ?

eguzki (approved):

PR #537, opened 2025-05-29
workflow: enable golangci

KevFan:
# Description
`golangc-lint` was always skipped in the code style github action due to an additional check of the presence of the `.golangci.yaml` configuration file, which was never present. 

This PR enables `golangc-lint` by removing this check, where the [standard](https://golangci-lint.run/usage/linters/#enabled-by-default) linters are run and fixes any linter issues detected by these checks


guicassolato:
That's a lot of bad code we're fixing. Thanks, @KevFan!

adam-cattermole (approved):
Nice, these changes look good to me

PR #543, opened 2025-06-17
dep: upgrade to jwt/v5

KevFan:
# Description
Upgrade to jwt/v5 to potentially fix dependabot alert

# Verification
Passing e2e should be enough as there's tests that use JWT auth
- https://github.com/Kuadrant/authorino/actions/runs/15826543810


adam-cattermole (approved):
Changes lgtm!

PR #544, opened 2025-07-01
refactor: cross compile and distribute build image

KevFan:
# Description
Image building is slow due to using QEMU and building the various platforms sequentially, taking about [54 mins](https://github.com/Kuadrant/authorino/actions/runs/15826379972) to complete the image build workflow.

This PR refactors the image build by:
1. Cross compiling instead of using emulating the targetted architecture 
2. Distribute the image build per architecture to individual github runners and merging the manifests at the end

This approach decreases the image build time to approximately **[3 mins](https://github.com/Kuadrant/authorino/actions/runs/15997307660)**, an **18** fold decrease.

This will allow quicker releases and potentially testings where users may not waiting/depend of the build images

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -15,17 +15,13 @@ env:
 
 jobs:
   prepare:
-    name: Prepare build metadata
+    name: Prepare build info
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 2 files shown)


maleck13 (approved):
Not familiar with the workflow but it looks reasonable and so happy to approve

Issue #553, opened 2025-08-26
Implement in Authorino


adam-cattermole:
Add required labels to all the metrics

PR #554, opened 2025-09-05
feat: custom labels on metrics via heuiristic path

KevFan:
# Description
Closes: https://github.com/Kuadrant/authorino/issues/553
 
* Allows for prometheus metrics with dynamic labels to be emitted
* Evaulates labels from `io.kuadrant.metrics.labels` and emit as part of auth config metrics that are resolvable 

# Verification
* Create cluster
```
kind create cluster
```
* Install CRDs
```
make install
```
* Apply AuthConfig
```
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - localhost:5001
  authentication:
    "friends":
      apiKey:
        selector:
          matchLabels:
            group: friends
      credentials:
        authorizationHeader:
          prefix: APIKEY
---
apiVersion: v1   
kind: Secret
metadata:
  name: api-key-1
  labels:
    group: friends
stringData:
type: Opaque
EOF
```
* Run authorino
```
make run
```
* gRPC curl to authorino
```
grpcurl -plaintext -d '{
  "attributes": {
    "metadata_context": {
      "filter_metadata": {
        "io.kuadrant.metrics.labels": {
          "methodLabel": { "cel_expr": "request.method" },
          "hostLabel": { "cel_expr": "request.host" },
          "stringLabel": "foo",
          "intLabel": 123,
          "boolLabel": true,
          "unsupportedNullLabel": null,
          "unResolvableLabel": { "cel_expr": "auth.nonexistent.value" },
          "authIdenityKindLabel": { "cel_expr": "auth.identity.kind" }
        }
      }
    },
    "request": {
      "http": {
        "host": "localhost:5001",
        "path": "/some-path",
        "method": "GET",
        "headers": {
          "authorization": "APIKEY ndyBzreUzF4zqDQsqSPMHkRhriEOtcRx"
        }
      }
    }
  }
}' localhost:50051 envoy.service.auth.v3.Authorization/Check
```
* Curl `/server-metrics` to view labels attached
  * `methodLabel`, `hostLabel` values ares dynamically resolved
  * `stringLabel` value is taken as string literal and set directly as`foo` string
  * `intLabel` value is taken as a int literal and set as a casted string
  * `intLabel` value is taken as a int literal and set as a casted string
  * `boolLabel` value is taken as a bool literal and set as a casted string
  * `unsupportedNullLabel` value is not a supported type so it ignored
  * `unResolvableLabel` cannot be resolved, so ignored
  * `authIdenityKindLabel` is resolvable only for metrics when it is present in the authJSON (should see only one metric emitted with this label)
```
curl localhost:8080/server-metrics | grep staticLabel
```


pkg/service/auth_pipeline.go
```diff
@@ -498,19 +473,44 @@ func (pipeline *AuthPipeline) Evaluate() auth.AuthResult {
...
+
+			if value, err := expr.ResolveFor(pipeline.GetAuthorizationJSON()); err != nil {
+				pipeline.Logger.Error(err, "failed to evaluate custom label expression", "expression", v.String())
+			} else if strValue, ok := value.(string); ok {
+				labels[k] = strValue
+			} else {
+				labels[k] = fmt.Sprintf("%v", value)
+			}
+		}
+	}
```
alexsnaps, line 511: I fear I may have changed that a bit this morning. See [these changes](https://github.com/Kuadrant/wasm-shim/pull/210/commits/47a1a8264787f5671f7e45f29640a648e69665cb?diff=unified&w=0), open to change it back tho. I sorta like that we don't conflate it all... but maybe that's a stupid idea... wdyt?

pkg/service/auth_pipeline.go
```diff
@@ -498,19 +475,63 @@ func (pipeline *AuthPipeline) Evaluate() auth.AuthResult {
...
 func (pipeline *AuthPipeline) reportStatusMetric(rpcStatusCode rpc.Code) {
-	metrics.ReportMetricWithStatus(authServerAuthConfigResponseStatusMetric, rpc.Code_name[int32(rpcStatusCode)], pipeline.metricLabels()...)
+	metrics.ReportMetricWithStatus(metrics.Registry.GetAuthServerAuthConfigResponseStatusMetric(), rpc.Code_name[int32(rpcStatusCode)], pipeline.metricLabels())
 }
 
-func (pipeline *AuthPipeline) metricLabels() []string {
-	labels := pipeline.AuthConfig.Labels
-	return []string{labels["namespace"], labels["name"]}
+func (pipeline *AuthPipeline) metricLabels() map[string]string {
+	labels := maps.Clone(pipeline.AuthConfig.Labels)
```
KevFan, line 489: This change causes a difference for the field name used for `AuthConfig.name`. Previous to this it would have been `authconfig`, but now it's `name` as the field name

For example: 
Before
```
auth_server_authconfig_total{authconfig="talker-api-protection",namespace="default"} 1
```
https://github.com/Kuadrant/authorino/blob/ddfd7e68f964d77ce939a45b038392708411503a/pkg/metrics/auth_config_metrics.go#L14

After:
```
auth_server_authconfig_total{name="talker-api-protection",namespace="default"} 1
```
https://github.com/Kuadrant/authorino/blob/ddfd7e68f964d77ce939a45b038392708411503a/controllers/auth_config_controller.go#L665

Is this an issue? Do we want to keep the pre-existing field name for `AuthConfig.Name`?

pkg/metrics/metrics.go
```diff
@@ -18,44 +20,130 @@ func Register(metrics ...prometheus.Collector) {
...
+		help:    help,
+		buckets: prometheus.LinearBuckets(0.001, 0.05, 20),
+	}
+}
+
+func (dh *DynamicHistogram) Observe(labels map[string]string, value float64) {
+	dh.mu.Lock()
+	defer dh.mu.Unlock()
+
+	key := fmt.Sprintf("%v", labels)
```
guicassolato, line 94: Same https://github.com/Kuadrant/authorino/pull/554/files#r2329734746

pkg/metrics/registry.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (r *MetricRegistry) GetAuthServerEvaluatorDurationMetric() *DynamicHistogram {
+	return r.GetOrCreateDynamicHistogram(
+		"auth_server_evaluator_duration_seconds",
+		"Response latency of individual authconfig rule evaluated by the auth server (in seconds).",
+	)
+}
+
+var Registry = NewMetricRegistry()
```
guicassolato, line 135: We used to define AuthPipeline-related metrics in the `service` package, for example. This made more sense to me than having some metric definitions widespread and others contained within the packages they are used.

pkg/metrics/metrics.go
```diff
@@ -18,44 +20,130 @@ func Register(metrics ...prometheus.Collector) {
...
 }
 
-func ReportMetric(metric *prometheus.CounterVec, labels ...string) {
-	metric.WithLabelValues(labels...).Inc()
+func (dc *DynamicCounter) Inc(labels map[string]string) {
+	dc.mu.Lock()
+	defer dc.mu.Unlock()
+
+	key := fmt.Sprintf("%v", labels)
+	if c, ok := dc.counters[key]; ok {
```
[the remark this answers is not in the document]
KevFan: Yeah, sounds good to me 👍

pkg/metrics/registry.go
KevFan: Yeah, fair. I can move this to `service` package

adam-cattermole (approved):
The changes look good from my point of view and the verification steps are passing. Only thing I guess to note is something we spoke about before; perhaps some follow up to handle cleanup of unused metrics to prevent them growing indefinitely

adam-cattermole (dismissed):
Nice, new changes look great and I've verified the label remains unchanged

guicassolato (dismissed):
/lgtm

PR #566, opened 2025-10-15
Claude Code Slash command `/protect-with-authorino`

guicassolato:
Add Anthropic's [Claude Code](https://github.com/anthropics/claude-code) Slash command `/protect-with-authorino`

### Usage

❶ Install [Claude Code](https://docs.claude.com/en/docs/claude-code/quickstart#step-1%3A-install-claude-code)

> **Note:** Recommended way for RedHatters is [configuring Claude Code through Google Vertex AI](https://docs.claude.com/en/docs/claude-code/google-vertex-ai#setup)

❷ Run claude CLI from the Authorino repo

```sh
cd $(ls -d -- "$GOPATH"/src/github.com/*/authorino(N/) | head -n1) && claude
```

❸ Run the command

E.g.:

```sh
/protect-with-authorino talker-api.127.0.0.1.nip.io "oidc authentication (issuer url: http://keycloak.keycloak.svc.cluster.local/realms/my-realm)" "'aud' jwt claim must be equal to 'talker-api'"
```

Claude should then:
1. Verify that a Kubernetes cluster is in context and accessible
2. Check if Authorino is installed and running in the cluster, otherwise install and deploy it from a local build using the Makefile targets
3. Parse the user input to extract the hostname, authentication rules, and authorization rules
4. Craft an AuthConfig resource based on the parsed arguments
5. Apply the AuthConfig to the cluster and check its status afterwards

### DIFF
--- a/.claude/commands/protect-with-authorino.md
+++ b/.claude/commands/protect-with-authorino.md
@@ -1,0 +1,163 @@
+---
+description: Protect a service with Authorino by generating and applying an AuthConfig CR
+---
+
+You are an expert at working with Authorino, a Kubernetes-native authorization service. Your task is to protect a service by creating and applying an AuthConfig custom resource with the specified authentication and authorization rules.
+
+## Task Overview
+
+1. **Check Kubernetes cluster**: Verify that a Kubernetes cluster is accessible via `kubectl`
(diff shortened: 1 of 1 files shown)


KevFan (approved):
Nice! Verified working as expected 💯

Issue #580, opened 2026-03-18
[X.509] Research and Design XFCC Support


guicassolato:
## Description

Research and design the API for X.509 client certificate authentication via XFCC (X-Forwarded-Client-Cert) header extraction in Authorino.

## Tasks

- [x] Analyze XFCC header format in Envoy documentation
- [x] Test XFCC header with Istio gateway in local environment
- [x] Test XFCC header with Envoy Gateway (if applicable)
- [x] Verify wasm-shim forwards XFCC header without modification
- [x] Identify certificate chain maximum size in XFCC
- [x] Design API types for `X509CertificateSource`

## Related

- RFC: https://github.com/Kuadrant/architecture/blob/main/rfcs/0015-x509-client-cert-authpolicy.md
- Epic: https://github.com/Kuadrant/architecture/issues/140

---
*Issue created by Claude Code*

guicassolato:
## Research Findings: XFCC Support in Authorino

### 1. XFCC Header Format Analysis

**Source:** [Envoy HTTP Header Manipulation Documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers)

#### Header Structure
- **Format:** Comma-separated XFCC elements, where each element represents information added by a single proxy
- **Element Structure:** Semicolon-separated key-value pairs (key=value)
- **Keys:** Case-insensitive | **Values:** Case-sensitive
- **Escaping:** Values containing `,`, `;`, or `=` must be double-quoted; internal `"` must be escaped as `\"`

#### Supported Keys in XFCC Header
1. **By** – Subject Alternative Name (URI type) of the current proxy's certificate
2. **Hash** – SHA 256 digest of the current client certificate
3. **Cert** – Entire client certificate in URL-encoded PEM format
4. **Chain** – Complete client certificate chain (including leaf) in URL-encoded PEM format
5. **Subject** – Subject field of the current client certificate (always double-quoted)
6. **URI** – URI type Subject Alternative Name field
7. **DNS** – DNS type Subject Alternative Name field

#### Processing Configuration
Envoy's XFCC handling is controlled by `forward_client_cert_details` and `set_current_client_cert_details`. Default: XFCC header is sanitized.

---

### 2. RFC Review Summary

**Source:** [RFC 0015: X.509 Client Certificate Authentication](https://raw.githubusercontent.com/Kuadrant/architecture/refs/heads/main/rfcs/0015-x509-client-cert-authpolicy.md)

[...]

PR #585, opened 2026-03-23
Add X.509 client certificate extraction from client cert headers (XFCC and RFC 9440)

### LINKED ISSUE #581: [X.509] Implement X.509 XFCC Header Support
## Description

Implement X.509 client certificate authentication via XFCC (X-Forwarded-Client-Cert) header extraction in Authorino's core authentication logic.

## Dependencies

- Blocked by: #580 (Research and Design XFCC Support)

## Tasks

- [x] Implement XFCC header parser
- [x] Implement certificate extraction from XFCC
- [x] Add backward compatibility handling
- [x] Write unit tests for XFCC parsing
- [x] Write integration tests with mocked headers
- [x] Update API documentation
- [x] Update AuthConfig examples
- [x] Submit PR to Authorino repository
- [x] Address code review feedback

## Acceptance Criteria

- XFCC header is correctly parsed following Envoy format
- Certificates are properly extracted and validated
- Backward compatibility is maintained for existing AuthConfigs
- All tests pass with >80% code coverage for new code
- Documentation is updated with examples

## Related

- RFC: https://github.com/Kuadrant/architecture/blob/main/rfcs/0015-x509-client-cert-authpolicy.md
- Epic: https://github.com/Kuadrant/architecture/issues/140
- Research Issue: #580

---
*Issue created by Claude Code*

guicassolato:
## Summary

Implements X.509 client certificate authentication via HTTP header extraction in Authorino's core authentication logic, as described in issue #581.

This enhancement enables Authorino to extract and validate client certificates from HTTP headers populated by upstream proxies (e.g., Envoy, Istio Gateway), in addition to the existing TLS connection attribute extraction. This is essential for deployment scenarios where the gateway terminates TLS and forwards certificate information via headers.

Supports three certificate extraction methods:
- **XFCC (X-Forwarded-Client-Cert)**: Envoy's standard format for forwarding client certificate information
- **Client-Cert (RFC 9440)**: Standardized HTTP header field for TLS client certificate forwarding
- **CEL Expression**: Flexible extraction from any field in the authorization JSON

## Changes

**API (v1beta3):**
- Added `X509CertificateSource` struct with `xfccHeader`, `clientCertHeader`, and `expression` fields
- Extended `X509ClientCertificateAuthenticationSpec` with optional `source` field
- Added CEL validation to ensure mutual exclusivity between source options (optimized for cost budget)
- Generated updated CRDs and manifests

**Core Implementation:**
- Implemented robust XFCC header parser following Envoy format specification (`pkg/evaluators/identity/xfcc.go`)
  - Handles semicolons in quoted values (e.g., `Subject="CN=test;OU=dev"`)
  - Handles equals signs in quoted values
  - Case-insensitive key matching
- Implemented RFC 9440 Client-Cert header parser (`pkg/evaluators/identity/client_cert.go`)
  - Converts DER to PEM for internal consistency
  - Validates base64 encoding and colon delimiters
- Extended MTLS evaluator to support multiple certificate sources with full backward compatibility (`pkg/evaluators/identity/mtls.go`)
  - Priority: XFCC → Client-Cert → Expression → Default (TLS attributes)
- Updated controller to wire source configuration through reconciliation loop

**Testing:**
- Added comprehensive unit tests for XFCC parsing logic (handles quoted values correctly)
- Added unit tests for RFC 9440 Client-Cert parsing
- Added integration tests for all three certificate sources
- Added tests for error scenarios (invalid base64, missing headers, untrusted CA, expired certs)
- All 41 tests passing with 100% backward compatibility verified

**Documentation:**
- Updated mTLS authentication user guide with all three certificate source examples
- Updated feature documentation with certificate source configuration details
- Added security considerations for HTTP header usage vs TLS attributes
- Documented Envoy/Istio configuration requirements for XFCC forwarding

## Key Features

✅ **Three Certificate Sources**: XFCC header, RFC 9440 Client-Cert header, or CEL expression  
✅ **Backward Compatible**: Existing AuthConfigs work without changes (defaults to TLS attributes)  
✅ **Robust Parsing**: Handles special characters in quoted values, case-insensitive keys  
✅ **Format Flexibility**: Supports Envoy XFCC, RFC 9440 DER, and custom PEM formats  
✅ **Security-Focused**: Validates only leaf certificate, includes security warnings  
✅ **Well-Tested**: Comprehensive test coverage with 100% passing rate

## Design Decisions

Following [RFC 0015](https://github.com/Kuadrant/architecture/blob/main/rfcs/0015-x509-client-cert-authpolicy.md) and research from issue #580:
- Validates only the **leaf certificate** (not full chain) per design decision
- Prefers `Cert` field over `Chain` field in XFCC (header size optimization)
- Uses first certificate if multiple are present
- Maintains full backward compatibility when source is not specified
- XFCC keys are case-insensitive per HTTP header standards
- RFC 9440 Client-Cert header is treated as a regular HTTP header by Envoy (not affected by `forward_client_cert_details`)
- **API field naming**: Uses `xfccHeader` (not `xfcc`) to clarify that this field specifies the HTTP header name containing the XFCC data

## Certificate Formats

**XFCC (Envoy format):**
```
By=spiffe://cluster.local/ns/default/sa/httpbin;Hash=SHA256:abc...;Subject="CN=client";URI=spiffe://cluster.local/ns/default/sa/sleep;Cert="-----BEGIN%20CERTIFICATE-----%0A...%0A-----END%20CERTIFICATE-----%0A"
```

**Client-Cert (RFC 9440):**
```
:MIIDDTCCAfWgAwIBAgIUJW4XTwuD...KRWZT0=:
```
(DER format, base64-encoded, colon-delimited)

**CEL Expression:**
```yaml
expression: 'request.http.headers["x-custom-cert"]'
```

## Example Configuration

**Using XFCC header:**
```yaml
spec:
  authentication:
    "mtls":
      x509:
        selector:
          matchLabels:
            app: my-ca
        source:
          xfccHeader: x-forwarded-client-cert
```

**Using RFC 9440 Client-Cert header:**
```yaml
spec:
  authentication:
    "mtls":
      x509:
        selector:
          matchLabels:
            app: my-ca
        source:
          clientCertHeader: client-cert
```

**Using CEL expression:**
```yaml
spec:
  authentication:
    "mtls":
      x509:
        selector:
          matchLabels:
            app: my-ca
        source:
          expression: source.certificate
```

## Verification Steps

Tested in local Kubernetes environment:
1. Created CA certificate and stored in labeled Secret
2. Applied AuthConfig with `source.clientCertHeader: "Client-Cert"`
3. Generated client certificate signed by trusted CA with `clientAuth` extended key usage
4. Converted certificate to RFC 9440 format
5. Verified successful authentication with valid certificate (HTTP 200)
6. Verified rejection of invalid base64 (HTTP 401)
7. Verified rejection of missing header (HTTP 401)
8. Verified rejection of untrusted CA (HTTP 401)

Full verification: Run `make local-setup` from this branch and follow the "Authentication with X.509 certificates and Mutual Transport Layer Security (mTLS)" user guide starting from step ❹.

## Related Issues

Closes #581  
Related to #580 (research/design)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * X.509 client certificate authentication now supports configurable extraction sources: Envoy XFCC headers, RFC 9440 Client‑Cert headers, or CEL expressions.

* **Documentation**
  * Updated feature docs and user guides with configuration examples, header formatting requirements and security notes for the new extraction options.

* **Tests**
  * Added comprehensive tests covering XFCC parsing, RFC 9440 extraction, CEL expression extraction and related error cases.


pkg/evaluators/identity/mtls.go
```diff
@@ -98,6 +120,44 @@ func (m *MTLS) Call(pipeline auth.AuthPipeline, ctx context.Context) (interface{
...
+
+	return pemCert, nil
+}
+
+// extractFromAttribute extracts the certificate from a custom attribute path
+func (m *MTLS) extractFromAttribute(_ auth.AuthPipeline) (string, error) {
+	// For future implementation: use AttributePath to navigate CheckRequest
+	// For now, this is a placeholder for advanced use cases
+	return "", fmt.Errorf("custom attribute path extraction not yet implemented")
+}
```
mikenairn, line 180: Should this be implemented? If not, would anyone trying to use "X509CertificateSource.Attribute" not always get runtime errors?

guicassolato: > Should this be implemented?

It's not part of the scope, but yes, we can implement it. In fact, I can make it part of this PR. It's relatively easy.

> If not, would anyone trying to use "X509CertificateSource.Attribute" not always get runtime errors?

Yes, it was added for future use.

guicassolato: Implemented in https://github.com/Kuadrant/authorino/pull/585/changes/4754970e61deb00abed0c5b0a586bdd21d7a2732.

I also made the following changes:
- `source.header` renamed as `source.xfcc` → more clear it refers to Envoy's proprietary XFCC format and keeps the API open for extension to support [RFC-9440](https://datatracker.ietf.org/doc/rfc9440/) `Client-Cert` HTTP header in the future
- `source.attribute` renamed as `source.expression` → CEL provides users with more options and it's consistent with other features in the API (e.g. `authentication.plain`)

Let me know what you think.

BTW, now we can also add RFC 9440 `Client-Cert` HTTP header if we want. It would probably make more sense for Tier 3 anyway. cc @averevki

pkg/evaluators/identity/xfcc.go
```diff
@@ -0,0 +1,196 @@
...
+	elements, err := parseXFCCHeader(headerValue)
+	if err != nil {
+		return "", err
+	}
+
+	// Per design decision: if multiple certificates present, use the first one
+	if len(elements) == 0 {
+		return "", fmt.Errorf("no XFCC elements found")
+	}
+
```
averevki, line 214: There is already the same check inside the `parseXFCCHeader` function called few lines above, it will never get to this return.

And I think the comment about first certificate fits better to the line with `certValue := elements[0].Cert`

pkg/evaluators/identity/xfcc.go
```diff
@@ -0,0 +1,196 @@
...
+	return elements
+}
+
+// parseXFCCElement parses a single XFCC element (one certificate)
+func parseXFCCElement(elementStr string) (xfccElement, error) {
+	element := xfccElement{}
+
+	// Split by semicolon to get key=value pairs
+	pairs := strings.Split(elementStr, ";")
```
averevki: > Each XFCC element is a semicolon `;` separated string. If `,`, `;` or `=` appear in a value, the value should be double-quoted

I think it won't correctly parse quoted `;` in the value. Don't really see the use case for this symbol except maybe the `Subject` field, but I guess needs to be supported by the XFCC header format

guicassolato: Good catch. I will fix it.

pkg/evaluators/identity/xfcc.go
guicassolato: I could argue it's for treating each function as independent from another, but I can remove the safeguard from `parseXFCCHeader` if it results in simpler code. Not a big deal IMO.

pkg/evaluators/identity/xfcc.go
guicassolato: Fixed. Thanks!

pkg/evaluators/identity/xfcc.go
guicassolato: Fixed.

pkg/evaluators/identity/mtls.go
guicassolato: Added support for RFC 9440 `Client-Cert` HTTP header in https://github.com/Kuadrant/authorino/pull/585/changes/b5540f4297e7cc6ba0d03c346ad75843e5f4a8f4.

api/v1beta3/auth_config_types.go
```diff
@@ -415,6 +415,37 @@ type X509ClientCertificateAuthenticationSpec struct {
...
+	// Name of the HTTP header containing the client certificate as defined in RFC 9440.
+	// Use this option when the certificate is forwarded in the Client-Cert header format (https://datatracker.ietf.org/doc/rfc9440/).
+	// The certificate must be in DER format, base64-encoded, and delimited by colons (:base64_cert:) as specified in the RFC.
+	// One of: xfccHeader, clientCertHeader, expression
+	// +optional
+	ClientCertHeader string `json:"clientCertHeader,omitempty"`
+
+	// CEL expression that resolves to the certificate in PEM format, usually extracted from the Envoy CheckRequest attributes.
+	// Use this option for advanced use cases where the certificate is available in a non-standard attribute path.
+	// The certificate must be in PEM format and URL-encoded.
```
mikenairn, line 444: Does the certificate _**have**_ to be URL-encoded, or is it more it will handle both?

mikenairn: Reason I'm asking is presumably the idea here is that you could for instance get the cert (in PEM format) from any header, or other attribute for that matter, but would those sources contain URL-encoded certificates or plain PEM?

mikenairn (dismissed):
Question about the "expression" source and the expected encoding of the value. Does it need to be URL-Encoded? I think it might work with either. 

Changes look good to me though.

api/v1beta3/auth_config_types.go
guicassolato: Some encoding of the certificate in PEM format is needed due to containing line breaks, such as with the `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----` delimiters being on their own lines.

Besides, `source.certificate`, the (well-known) attribute from the `CheckRequest` where client certificates are provided by default (if not specified otherwise), is URL-encoded. By requiring `authconfigs.authentication.x509.source.expression` to adopt the same format, one can do:

```yaml
authentication:
  "my-x509-client-cert-rule":
    x509:
      source:
        expression: source.certificate
```

Although the above is not an option for Kuadrant – due to the wasm-shim not filling `source.certificate` – it's still nice that it works for standalone integrations, as a way for the user to be explicit about intending the default behaviour.

That said, imagine we didn't expect PEM/URL-encoded, but plain base64-encoded DER. Being CEL, we could technically introduce a function to make URL-decoding a user decision. E.g.:

```yaml
authentication:
  "my-x509-client-cert-rule":
    x509:
      source:
        expression: |
          url.decode(source.certificate)
            .replace("-----BEGIN CERTIFICATE-----", "")
            .replace("-----END CERTIFICATE-----", "")
            .replace("\n", "")
            .trim()
```

or with two extension functions:

```yaml
authentication:
  "my-x509-client-cert-rule":
    x509:
      source:
        expression: pem.parse(url.decode(source.certificate))
```

Unfortunately, we don't have such functions today, and I'd rather not make this change about extending CEL if we can.

david-martin:
(test)

@coderabbitai review

mikenairn (approved):

PR #595, opened 2026-04-19
controllers/auth_config_controller: nil-guard optional user on kubernetesSubjectAccessReview

### LINKED ISSUE #494: Panic in authorino pod when defining a k8s SAR rule without the user field
**Describe the bug**

The authorino process panics when the `user` field in a k8s SAR authorization rule is not defined.

**Help us Reproduce it**

1. Create the following Kuadrant AuthPolicy

```yaml
apiVersion: kuadrant.io/v1beta2
kind: AuthPolicy
metadata:
  name: sar-protected-api
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: some-route
  routeSelectors:
    - matches:
        - path:
            type: PathPrefix
            value: /some-path
  rules:
    authentication:
      "service-accounts":
        kubernetesTokenReview:
          audiences:
            - "https://example.com"
    authorization:
      "k8s-rbac":
          groups:
            - "some-group"
```
2. Check the logs of the authorino pod to see the following panic

```
{"level":"info","ts":"2024-10-10T14:30:50Z","logger":"authorino","msg":"Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference","controller":"authconfig","controllerGroup":"authorino.kuadrant.io","controllerKind":"AuthConfig","AuthConfig":{"name":"ap-3scale-saas-backend-internal-api","namespace":"3scale-saas"},"namespace":"3scale-saas","name":"ap-3scale-saas-backend-internal-api","reconcileID":"23813405-f48b-4cda-a0f9-ef21c5d6aa8c"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1b5cd9c]
goroutine 180 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:116 +0x1e5
panic({0x1de0b00?, 0x36d5b30?})
/usr/lib/golang/src/runtime/panic.go:914 +0x21f
github.com/kuadrant/authorino/controllers.(*AuthConfigReconciler).translateAuthConfig(0xc00067e680, {0x25020a8, 0xc000c40db0}, 0xc00027f4a0)
/usr/src/authorino/controllers/auth_config_controller.go:424 +0x1bdc
github.com/kuadrant/authorino/controllers.(*AuthConfigReconciler).Reconcile(0xc00067e680, {0x25020a8, 0xc000afdf80}, {{{0xc0003edaa0?, 0x5?}, {0xc00043ae70?, 0xc000736d08?}}})
/usr/src/authorino/controllers/auth_config_controller.go:114 +0x4ee
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x2507490?, {0x25020a8?, 0xc000afdf80?}, {{{0xc0003edaa0?, 0xb?}, {0xc00043ae70?, 0x0?}}})
/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:119 +0xb7
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0007c14a0, {0x25020e0, 0xc0007b1270}, {0x1ebbba0?, 0xc0000a87a0?})
/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:316 +0x3cc
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0007c14a0, {0x25020e0, 0xc0007b1270})
/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266 +0x1af
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227 +0x79
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2 in goroutine 93
/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:223 +0x565
```

**Expected behavior**

Don't panic and report the error somehow (logs, conditions or even reject the AuthPolicy/AuthConfig). 

- Openshift 4.16
- kuadrant v0.11.0 & authorino-operator v0.13.0

**Additional context**

Slack conversation https://kubernetes.slack.com/archives/C05J0D0V525/p1728554107621359

SAY-5:
## What

`translateAuthConfig` in `controllers/auth_config_controller.go` handles `KubernetesSubjectAccessReviewAuthorization` with:

```go
case api.KubernetesSubjectAccessReviewAuthorization:
    user := authorization.KubernetesSubjectAccessReview.User
    authorinoUser, err := valueFrom(user)
    if err != nil {
        return nil, err
    }
    ...
```

`KubernetesSubjectAccessReview.User` is a `*ValueOrSelector` — a pointer — and the AuthConfig spec treats it as optional. A group-only SAR rule (only `groups:` set) or an implicit self-SAR doesn't have to spell out `user:`. When the caller omits it the pointer is `nil`, and the first deref inside `valueFrom` (`user.Expression != ""`) panics.

```go
// line 814
if user.Expression != "" {  // nil pointer deref when user == nil
```

Per #494: `sigs.k8s.io/controller-runtime` recovers the panic but then re-enqueues the reconcile, so the controller crash-loops on every AuthConfig event. The user-visible symptom is an otherwise-valid `AuthPolicy` like:

```yaml
rules:
  authorization:
    "k8s-rbac":
        groups:
          - "some-group"
```

taking the whole authorino pod into a restart loop.

## Fix

The sibling optional fields on the same evaluator (`ResourceAttributes`, `AuthorizationGroups`) are already guarded by nil-checks at the caller level. Match that pattern for `User`: declare `authorinoUser` as a zero `expressions.Value` and only populate it when the caller supplied a `User` spec.

```go
var authorinoUser expressions.Value
if user := authorization.KubernetesSubjectAccessReview.User; user != nil {
    var err error
    authorinoUser, err = valueFrom(user)
    if err != nil {
        return nil, err
    }
}
```

Downstream, `NewKubernetesAuthz` + `jsonValueToStr` already treat a nil `expressions.Value` as "no user":

```go
jsonValueToStr := func(value expressions.Value) (string, error) {
    if value == nil {
        return "", nil
    }
    ...
}
```

which becomes `SubjectAccessReviewSpec.User == ""` on the wire, and the Kubernetes API evaluates it as expected (group-only / self-SAR). No behaviour change for any `AuthConfig` that does set `user:`.

Fixes #494

## Bug Fixes

* Resolved a crash in Kubernetes Subject Access Review authorisation when the User field was not provided in the configuration. The system now correctly handles configurations where the User field is optional, improving stability and allowing administrators to define authorisation policies with greater flexibility without requiring explicit User specifications in all cases.

### DIFF
--- a/controllers/auth_config_controller.go
+++ b/controllers/auth_config_controller.go
@@ -578,10 +578,21 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
 			}
 
 		case api.KubernetesSubjectAccessReviewAuthorization:
-			user := authorization.KubernetesSubjectAccessReview.User
-			authorinoUser, err := valueFrom(user)
-			if err != nil {
-				return nil, err
+			// The `user` field on kubernetesSubjectAccessReview is optional:
+			// omitting it is a valid way to request a group-only or self-SAR
+			// check. Guard against the nil pointer here so we don't segfault
+			// the controller's Reconcile (authorino then keeps crash-looping
+			// on every AuthConfig event). NewKubernetesAuthz + jsonValueToStr
+			// already treat a nil expressions.Value as "no user", which maps
+			// to SubjectAccessReviewSpec.User == "" and the Kubernetes API
+			// evaluates it as expected.
+			var authorinoUser expressions.Value
+			if user := authorization.KubernetesSubjectAccessReview.User; user != nil {
+				var err error
+				authorinoUser, err = valueFrom(user)
+				if err != nil {
+					return nil, err
+				}
 			}
 
 			var authorinoResourceAttributes *authorization_evaluators.KubernetesAuthzResourceAttributes


guicassolato:
Thanks for the PR, @SAY-5!

guicassolato (approved):
Ideally, we would have tests to ensure `groups` can be specified without `user` and vice-versa, but that's fine given the poor coverage on these specific controller rules today. It's not like it would be only a small marginal effort, so I'm happy with deferring those to a separate PR.

The case for self-SAR (i.e., nor `user` nor `groups` specified in the AuthConfig) isn't clear to me. I actually think it won't even work anyway. The API server will likely reject the request with `spec.user: Invalid value: "": at least one of user or group must be specified` every time, and Authorino will instruct a 403 response.

**If you hit this error in your logs**, ensure your AuthConfig specifies at least one of `user`, `groups`, or `authorizationGroups`.

We should mitigate that by adding a CEL validation `has(self.user) || size(self.groups) > 0 || has(self.authorizationGroups)` to the AuthConfig. However, not only I think this is slightly out of scope for this PR, but also it wouldn't fully prevent errors on the wire anyway, given that expressions to dynamically fill `user` and `groups` could still evaluate to empty values for specific requests. Whereas the focus of this PR is to fix the bug when the `user` field is missing altogether – and it does exactly that! –, one could still have the field present and set it to an empty string. So I'm also happy with working on this validation separately.

**Final note:** I get the feeling this an AI-generate PR, but we're not picky 🙂 Thanks for the fix! I'm happy to count your contribution, @SAY-5.

PR #606, opened 2026-05-06
fix: convert ParseFloat result to float32 before type assertion

pierDipi:
strconv.ParseFloat always returns float64 regardless of bitSize parameter. The float32 case in EnvVar was asserting the float64 result directly as float32, causing a runtime panic when the KUBE_CLIENT_QPS env var is set.

Adds test coverage for float32 and float64 EnvVar paths.

### DIFF
--- a/pkg/utils/envvar.go
+++ b/pkg/utils/envvar.go
@@ -29,7 +29,7 @@ func EnvVar[T envVar](key string, def T) T {
 			return any(v).(T)
 		case reflect.Float32:
 			v, _ := strconv.ParseFloat(val, 32)
-			return any(v).(T)
+			return any(float32(v)).(T)
 		case reflect.Float64:
 			v, _ := strconv.ParseFloat(val, 64)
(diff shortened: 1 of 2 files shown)


bartoszmajsak:
Good catch, I think I only used flags when testing QPS.

/lgtm

guicassolato (approved):
Nice catch indeed. Thanks!

pierDipi:
@guicassolato could this be included in the next path release of RHCL? how can I help? a cherry-pick somewhere?

guicassolato:
> @guicassolato could this be included in the next path release of RHCL? how can I help? a cherry-pick somewhere?

You mean next _patch_ release. I'll have to check if there's one in the pipeline happening soon, including other fixes like CVEs. In case there isn't any, it will for sure be in the next release of Kuadrant (v1.5) and therefore also in the cut for RHCL v1.4 in a couple of weeks from now.

PR #455, opened 2024-01-19
fix: bump OpenTelemetry Schema URL to 1.21.0

guicassolato:
Fixes an error in the initialization of Authorino when tracing services are enabled.

Error message:

```
cannot merge resource due to conflicting Schema URL
```

Stacktrace:

```
main.setupTelemetryServices
	/usr/src/authorino/main.go:409
main.setup
	/usr/src/authorino/main.go:387
main.runAuthorizationServer
	/usr/src/authorino/main.go:225
github.com/spf13/cobra.(*Command).execute
	/opt/app-root/src/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:944
github.com/spf13/cobra.(*Command).ExecuteC
	/opt/app-root/src/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068
github.com/spf13/cobra.(*Command).Execute
	/opt/app-root/src/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992
github.com/spf13/cobra.(*Command).ExecuteContext
	/opt/app-root/src/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:985
main.main
	/usr/src/authorino/main.go:149
runtime.main
	/usr/lib/golang/src/runtime/proc.go:250
```

This was likely caused after upgrading otel deps to v0.46.x (#448) and it's a known issue of otel's Resource API: https://github.com/open-telemetry/opentelemetry-go/issues/2341.

PR #611, opened 2026-05-14
Update trace schema to v1.40.0

adam-cattermole:
I think #593 bumped the SDK without bumping the schema, and so we have an error due to version mismatch:

```js
ERROR authorino.telemetry unable to set trace provider {"error": "conflicting Schema URL: https://opentelemetry.io/schemas/1.40.0 and https://opentelemetry.io/schemas/1.39.0"}
```

Also I've bumped the otlptracegrpc as that was missed with the other libraries.

See related:
- #532
- #455

Wonder if there'd be a nice way to automate this update, or at least test for it so that dependabot merges don't break tracing support.

## Summary by CodeRabbit

* **Chores**
  * Updated OpenTelemetry tracing and semantic conventions libraries to latest versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/authorino/pull/611)


guicassolato (approved):
If only it were the first time it had happened: https://github.com/Kuadrant/authorino/pull/577, https://github.com/Kuadrant/authorino/pull/532, https://github.com/Kuadrant/authorino/pull/455 😂

Thanks, @adam-cattermole!

PR #613, opened 2026-05-15
fix: unwrap string-backed ValueOrSelector values in controller translation

sphr2k:
## Summary

I hit this error with an `AuthConfig` using Generic HTTP metadata like:

```yaml
http:
  method: GET
  url: https://some.url
  headers:
    Accept:
      value:
        string: application/json
```

This value round-trips through the Kubernetes API server as:

```json
{"string":"application/json"}
```

Without unwrapping this form during controller translation, Generic HTTP metadata headers can be sent with the serialized wrapper object instead of the intended primitive string value.

In practice, the `Accept` header above was effectively sent as:

```text
{"string":"application/json"}
```

instead of:

```text
application/json
```

This can cause metadata fetches that expect JSON responses to fail, depending on how strictly the upstream server validates request headers.

## Changes

- unwrap string-backed `ValueOrSelector.value` payloads in `valueFrom()`
- add a regression test covering the wrapped string case

## Verification

- `go test ./controllers -run 'TestValueFromUnwrapsStringValue'`
- `go test ./pkg/evaluators/metadata -run 'TestGenericHttpCallWithCustomHeaders'`

## Release Notes

* **Bug Fixes**
  * Improved static value processing in configurations to correctly unmarshal JSON data and unwrap nested string values.

* **Tests**
  * Added test coverage to verify proper unwrapping of string values in configuration handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/authorino/pull/613?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/controllers/auth_config_controller.go
+++ b/controllers/auth_config_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	gojson "encoding/json"
 	"fmt"
 	"sort"
 	"sync"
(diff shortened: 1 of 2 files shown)


sphr2k:
@coderabbitai review

guicassolato:
Unwrapping `value: {string: application/json}` as `{"string":"application/json"}` is the exact intended behaviour.

One who wants `"application/json"` should instead use `value: "application/json"`.

Please see the description of the [`ValueOrSelector`](https://pkg.go.dev/github.com/kuadrant/authorino/api/v1beta2#ValueOrSelector) type, typed as [`k8sruntime.RawExtension`](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#RawExtension).

PR #624, opened 2026-06-11
Add MetadataSpec and AuthorizationSpec CEL validation rules

### LINKED ISSUE #599: Add CRD XValidation rules for mutually exclusive fields in AuthConfig
## Context

Several types in `auth_config_types.go` have mutually exclusive field sets where only one should be set, but there's no CRD-level validation enforcing this. Users can set both fields without an API server error, leading to silent failures or undefined behavior at runtime.

Some types already have `XValidation` rules (e.g., `JwtAuthenticationSpec`, `X509CertificateSource`, `UserInfoMetadataSpec`). This issue tracks adding similar rules to the remaining candidates.

## Candidates

The following types have field sets that appear mutually exclusive based on comments, `GetMethod()` dispatchers, or documentation. Not all may require a strict one-of rule — listing as candidates for review.

### `HttpEndpointSpec` (line 491)
- `url` vs `urlExpression` — static URL vs CEL expression
- `body` vs `bodyParameters` — comments explicitly say "use either one or the other"

### `ValueOrSelector` (line 200)
- `value` vs `selector` vs `expression` — three-way

### `PlainIdentitySpec` (line 452)
- `selector` vs `expression`

### `AuthenticationMethodSpec` (line 329)
- `apiKey` vs `jwt` vs `oauth2Introspection` vs `kubernetesTokenReview` vs `x509` vs `plain` vs `anonymous`

### `MetadataMethodSpec` (line 481)
- `http` vs `userInfo` vs `uma`

### `AuthorizationMethodSpec` (line 621)
- `patternMatching` vs `opa` vs `kubernetesSubjectAccessReview` vs `spicedb`

### `AuthResponseMethodSpec` (line 805)
- `plain` vs `json` vs `wristband`

### `Credentials` (line 295)
- `authorizationHeader` vs `customHeader` vs `queryString` vs `cookie`

### `OpaAuthorizationSpec` (line 637)
- `rego` vs `externalPolicy`

## Reference

Types that already have XValidation rules and can serve as a pattern:
- `JwtAuthenticationSpec` (line 362): `jwksUrl` vs `issuerUrl`
- `X509CertificateSource` (line 426): `xfccHeader` vs `clientCertHeader` vs `expression`
- `UserInfoMetadataSpec` (line 581): `identitySource` vs `userInfoUrl`
- Legacy patch file: [`install/crd/patches/oneof_in_authconfigs.yaml`](https://github.com/Kuadrant/authorino/blob/main/install/crd/patches/oneof_in_authconfigs.yaml)

averevki:
## Summary

Replaces oneOf patches with `+kubebuilder:validation:XValidation` CEL annotations for `MetadataSpec` and `AuthorizationSpec`, and simplifies the existing rules on `JwtAuthenticationSpec` and `UserInfoMetadataSpec`. Also adds comprehensive CEL validation tests for all four changed rules.

  Closes #599 — all candidates are now addressed between #602 and this PR. Migrating the remaining rules require further investigation as fitting them into the CRD cost budget would require simplifying existing rules. It is not worth spending time on in my opinion.

  ### Changes

  | Type | Fields | Change |
  |------|--------|--------|
  | `MetadataSpec` | `http` / `userInfo` / `uma` | New CEL rule (exactly one) |
  | `AuthorizationSpec` | `patternMatching` / `opa` / `kubernetesSubjectAccessReview` / `spicedb` | New CEL rule (exactly
  one) |
  | `JwtAuthenticationSpec` | `jwksUrl` / `issuerUrl` | Simplified CEL rule |
  | `UserInfoMetadataSpec` | `identitySource` / `userInfoUrl` | Simplified CEL rule |

  ## Verification

  ```bash
  make test-cel
```

## Summary by CodeRabbit

* **Improvements**
  * Strengthened CRD validation for `AuthConfig` to enforce correct “exactly one” combinations and mutual exclusivity across JWT authentication, metadata sources, and authorisation methods.
  * Updated validation logic and messages to make misconfigurations easier to diagnose.
* **Tests**
  * Added comprehensive CEL validation tests for `v1beta3` `AuthConfig`, covering missing required fields and invalid combinations for the updated configuration rules.

### DIFF
--- a/api/v1beta3/auth_config_types.go
+++ b/api/v1beta3/auth_config_types.go
@@ -368,7 +368,7 @@ type ApiKeyAuthenticationSpec struct {
 
 // Settings to fetch the JSON Web Key Set (JWKS) for the JWT authentication.
 //
-//	+kubebuilder:validation:XValidation:rule="!has(self.jwksUrl) || !has(self.issuerUrl)",message="Use one of: jwksUrl, issuerUrl"
+//	+kubebuilder:validation:XValidation:rule="has(self.jwksUrl) != has(self.issuerUrl)",message="Use one of: jwksUrl, issuerUrl"
 type JwtAuthenticationSpec struct {
 	// URL of the JSON Web Key Set (JWKS) endpoint.
 	// Use it for non-OpenID Connect (OIDC) JWT authentication, where the JWKS URL is known beforehand.
@@ -613,7 +613,7 @@ type OAuth2ClientAuthentication struct {
 
 // Settings of the OpendID Connect UserInfo linked to an OIDC-enabled JWT authentication config of this same AuthConfig.
 //
-//	+kubebuilder:validation:XValidation:rule="!has(self.identitySource) || !has(self.userInfoUrl)",message="Use one of: identitySource, userInfoUrl"
+//	+kubebuilder:validation:XValidation:rule="has(self.identitySource) != has(self.userInfoUrl)",message="Use one of: identitySource, userInfoUrl"
 type UserInfoMetadataSpec struct {
 	// Name of an OIDC JWT authentication rule whose obtained configuration includes an "userinfo_endpoint" claim.
 	// One of: identitySource, userInfoUrl
--- a/api/v1beta3/auth_config_types.go
+++ b/api/v1beta3/auth_config_types.go
@@ -368,7 +368,7 @@ type ApiKeyAuthenticationSpec struct {
 
 // Settings to fetch the JSON Web Key Set (JWKS) for the JWT authentication.
 //
-//	+kubebuilder:validation:XValidation:rule="!(has(self.jwksUrl) && self.jwksUrl != '' && has(self.issuerUrl) && self.issuerUrl != '')",message="Use one of: jwksUrl, issuerUrl"
+//	+kubebuilder:validation:XValidation:rule="!has(self.jwksUrl) || !has(self.issuerUrl)",message="Use one of: jwksUrl, issuerUrl"
 type JwtAuthenticationSpec struct {
 	// URL of the JSON Web Key Set (JWKS) endpoint.
 	// Use it for non-OpenID Connect (OIDC) JWT authentication, where the JWKS URL is known beforehand.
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: api/v1beta3/auth_config_types.go [hunk 1]
```diff
@@ -368,7 +368,7 @@ type ApiKeyAuthenticationSpec struct {
 
 // Settings to fetch the JSON Web Key Set (JWKS) for the JWT authentication.
 //
-//	+kubebuilder:validation:XValidation:rule="!(has(self.jwksUrl) && self.jwksUrl != '' && has(self.issuerUrl) && self.issuerUrl != '')",message="Use one of: jwksUrl, issuerUrl"
+//	+kubebuilder:validation:XValidation:rule="!has(self.jwksUrl) || !has(self.issuerUrl)",message="Use one of: jwksUrl, issuerUrl"
```
guicassolato: I think this may validate an `authentication.jwt` spec where both `jwksUrl` and `issuerUrl` are missing, no?

averevki: Yes, I believe this is the same behavior how the validation was working before. The same case is with the new rule for `UserInfoMetadataSpec`. 

I didn't want to introduce a breaking change, but the empty `authentication.jwt`, with how it is set right now, could block [the existing oneOf rule](https://github.com/Kuadrant/authorino/blob/c8fde6d394ffe10c73a98d7450116ec862d4d6ca/install/crd/patches/oneof_in_authconfigs.yaml#L218-L248) for `authentication.*` fields. Should I fix it by changing them from `one of` to `exactly one from`? This makes sense to me to do

guicassolato: > I believe this is the same behavior how the validation was working before.

True. It's wrong though. Either `jwksUrl` or `issuerUrl` should be specified.

> Should I fix it by changing them from `one of` to `exactly one from`?

The message is good and consistent with the other one-of rules. But the actual behaviour should be _exactly one of_, yes.

Thanks!

averevki: I've kept the `one of` message, but changed the both rules operation to the real XOR now

guicassolato (approved):
Thank you @averevki!

PR #634, opened 2026-06-18
Add two-phase release workflow

Boomatang:
# Summary

Introduces an automated two-phase release process for Authorino, replacing the previous manual git-tag-and-build approach. The new workflow splits every release into a **Pre-release** phase (branch creation, version bump, code generation, PR) and a **Release** phase (smoke tests, tagging, image build, GitHub Release), with a mandatory human review gate between them.

This change implements the design specified in [RFC: Two-Phase Release Workflow](https://github.com/Kuadrant/architecture/pull/178).

# Motivation

The previous release process was entirely manual: a maintainer checked out a ref, created a signed tag, pushed it, manually triggered the image build, and hand-wrote release notes. This was error-prone, undocumented in automation, and lacked pre-release validation.

The two-phase approach automates the mechanical steps while preserving human oversight at the critical review point before a release is finalized.

# What changed

## New workflows

- **Pre-release** (`pre-release.yaml`) — triggered manually with a target version. Creates the `release-X.Y` branch (if needed), opens a `pre-release-vX.Y.Z` branch that bumps `release.yaml`, runs `make generate` and `make manifests`, and opens a PR against the release branch.
- **Release** (`release.yaml`) — triggered manually with a release branch name. Reads the version from `release.yaml`, validates the branch matches, runs smoke tests (lint, unit tests, CEL tests), creates an annotated tag, builds and pushes the multi-arch container image via `build-images.yaml`, and creates the GitHub Release.
- **Version Gate** (`version-gate.yaml`) — runs on PRs targeting `release-**` branches when `release.yaml` changes. Validates that the version is not `0.0.0` on non-main branches and that any declared dependencies have published releases.

## New scripts

- `parse-version.sh` — extracts and validates semver components from `release.yaml`, used by the release workflow.
- `validate-release-yaml.sh` — validates `release.yaml` content and dependency versions, used by the version gate.

## Modified workflows

- **Build and push image** (`build-images.yaml`) — now supports `workflow_call` with `version` and `ref` inputs so the release workflow can invoke it directly. Image tagging logic updated to use the version from inputs when provided.

## New files

- `release.yaml` — version source-of-truth file at the repository root. Set to `0.0.0` on `main` (active development) and updated to the target version on release branches.

## Updated documentation

- `RELEASE.md` — rewritten to document the new two-phase process, including step-by-step instructions, artifact locations.

# Examples

- Pre-release workflow: https://github.com/Boomatang/authorino/actions/runs/28017001760
- Opened PR: https://github.com/Boomatang/authorino/pull/2
- Release workflow: https://github.com/Boomatang/authorino/actions/runs/28017819516
- GitHub release: https://github.com/Boomatang/authorino/releases/tag/v0.1.0

## Release Notes

* **Documentation**
  * Rewrote the release guide to describe a structured two-phase process (Pre-release and Release), including artefacts, version rules, and required secrets.

* **Chores**
  * Added automation to validate `release.yaml`, parse and gate versioning for release branches, and verify dependency release tags.
  * Introduced pre-release and release workflows with consistent version/tag handling and image build/publish.
  * Updated `release.yaml` to set `authorino.version` to `0.0.0`.

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -1,0 +1,139 @@
+name: Pre-release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Release version (semver, e.g. 1.5.0)"
+        required: true
+        type: string
+      source-branch:
+        description: "Branch to base the pre-release changes on (default: main)"
+        required: false
+        type: string
+        default: "main"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.validate.outputs.version }}
+      release-branch: ${{ steps.validate.outputs.release-branch }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.source-branch }}
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Validate version format
+        id: validate
+        run: |
+          VERSION="${{ inputs.version }}"
+          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$ ]]; then
+            echo "::error::Invalid semver version: $VERSION"
+            exit 1
+          fi
+          MAJOR=$(echo "$VERSION" | cut -d. -f1)
+          MINOR=$(echo "$VERSION" | cut -d. -f2)
+          RELEASE_BRANCH="release-${MAJOR}.${MINOR}"
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "release-branch=$RELEASE_BRANCH" >> "$GITHUB_OUTPUT"
+
+      - name: Create or verify release branch
+        run: |
+          RELEASE_BRANCH="${{ steps.validate.outputs.release-branch }}"
+          if git ls-remote --exit-code origin "refs/heads/${RELEASE_BRANCH}" >/dev/null 2>&1; then
+            echo "Release branch '${RELEASE_BRANCH}' already exists"
+          else
+            echo "Creating release branch '${RELEASE_BRANCH}' from '${{ inputs.source-branch }}'"
+            git checkout -b "${RELEASE_BRANCH}"
+            git push origin "${RELEASE_BRANCH}"
+          fi
+
+  prepare-release:
+    needs: setup
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.setup.outputs.release-branch }}
+          fetch-depth: 0
+
+      - name: Create pre-release branch
+        run: |
+          PRE_RELEASE_BRANCH="pre-release-v${{ needs.setup.outputs.version }}"
+          if git ls-remote --exit-code origin "refs/heads/${PRE_RELEASE_BRANCH}" >/dev/null 2>&1; then
+            echo "::error::Pre-release branch '${PRE_RELEASE_BRANCH}' already exists. Delete it first or use a different version."
+            exit 1
+          fi
+          git checkout -b "${PRE_RELEASE_BRANCH}"
+
+      - name: Update release.yaml
+        run: |
+          VERSION="${{ needs.setup.outputs.version }}"
+          yq -i ".authorino.version = \"${VERSION}\"" release.yaml
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Run code generation
+        run: |
+          make generate
+          make manifests
+
+      - name: Commit and push changes
+        run: |
+          VERSION="${{ needs.setup.outputs.version }}"
+          git config user.name "github-actions[bot]"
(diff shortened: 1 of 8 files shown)


.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,139 @@
+name: Pre-release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Release version (semver, e.g. 1.5.0)"
+        required: true
+        type: string
+      source-branch:
```
guicassolato, line 10: Every workflow_dispatch already has a source branch field. Can't that one be used instead?

Boomatang: Possible yes, but also maybe not. The scenario I am thinking that it is required is doing back ports to older releases. We will want to make sure that the workflow is ran from the branch that the backport is targeted at, but that would be different than the source-branch to run the workflow against. 

Does that make sense?

guicassolato: It makes sense.

For the workflow to run, I think it needs to be defined in the branch we pick in that field from the workflow_dispatch form. It would be hard to pre-release a patch on a version from when the workflow originally didn't exist yet. We would have to back port the workflow itself to all maintained versions.

Having the not-so-redundant field also gives us the option to run the latest version of the workflow (as in the one in `main`) against any arbitrary source branch, regardless of the version of the workflow in that branch.

I guess having to type the name of the branch again isn't the end of the world. We just need to be careful about typos – which arguably we could validate in a step of the workflow (if not already).

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+          ### Checklist
+          - [ ] Version numbers are correct
+          - [ ] All pre-release modifications look correct
+          - [ ] CI checks pass
+          - [ ] Version gate check passes
+
+          ### Next steps
+          After merging this PR, run the **Release** workflow with branch \`${RELEASE_BRANCH}\`.
+          EOF
```
guicassolato, line 138: Won't this indentation be a problem for the body of the PR? I wonder if the text won't end up as all wrapped in a code block in markdown.

.github/workflows/version-gate.yaml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  validate-release-yaml:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # ratchet:actions/checkout@v4
+
+      - name: Install yq
+        run: |
+          sudo wget -qO /usr/local/bin/yq \
+            https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
```
guicassolato, line 22: Maybe pin yq to a specific version instead of `latest`, to avoid possibly breaking on external updates?

guicassolato (commented):
Process for Authorino releases today is pretty _simple_ – though arguably requires attention due being significantly manual.

With automation, it will probably become _easier_, though less simple – more steps and more elements to handle (e.g. a `release.yaml` file) – and more abstract.

I wonder if we shouldn't include a "rollback guidance" section to RELEASE.md.

.github/workflows/pre-release.yaml
Boomatang: The indentation is not a problem. You can see a PR that was open with this formatting here: https://github.com/Boomatang/authorino/pull/2.

.github/workflows/version-gate.yaml
Boomatang: Sure I can pin it to a version. I will check the rest of the code base to see what  version is defined and use the same that.

Boomatang:
> I wonder if we shouldn't include a "rollback guidance" section to RELEASE.md.

@guicassolato I think this would be a good idea, but I am not sure if I am understanding what you are wanting to roll back. I am going to assume it would be the reverting of the commit the is generated during the pre-release workflow. 

Can you explain what you intend by roll back.

> With automation, it will probably become easier, though less simple – more steps and more elements to handle (e.g. a release.yaml file) – and more abstract.

I see this in a few different ways. It does become easier to release in the global context of kuadrant. We are work towards every repo having the same release workflow (pre-release pipeline, PR review, release pipeline). I am not sure if you could call it less simple. I know authorino isn't the worse of our projects for this but it still does. When a release was made other workflows would get triggered. The person doing the release has very little context into how many workflows were, or should be triggered. And knowing if any of  those workflows had fail is a different story. So do think it is less simple, it's more where the complicity is has being shifted. 

The `release.yaml` I guess strictly is not need, but it does introduce a common pattern across all our repos, Looking at this from kuadrant as a whole, it becomes simpler to know all the version of components included in a given commit of the kuadrant-operator. A script can follow the trail of dependencies in the release.yaml files till there is none left.  

> Process for Authorino releases today is pretty simple – though arguably requires attention due being significantly manual.

I agree it is one of our simpler projects to release, but the manual steps do make it harder. And harder to automated a global release.

.github/workflows/version-gate.yaml
Boomatang: As branch protections are now working correctly at the org level I can push to this branch any more. This change is covered in this PR https://github.com/Kuadrant/authorino/pull/658

Boomatang:
Because of branch protections I have reopen this PR under a different branch. That PR can be seen here. https://github.com/Kuadrant/authorino/pull/666

PR #640, opened 2026-06-22
Add authorino-operator manifests sync triggering workflow

averevki:
Adds a GitHub Actions workflow that sends a `repository_dispatch` event to [authorino-operator](https://github.com/Kuadrant/authorino-operator) when manifests are changed on the `main` branch (`install/manifests.yaml` or `install/crd/**`).

This triggers the operator's [sync-authorino-manifests](https://github.com/Kuadrant/authorino-operator/blob/sync-manifests-workflow/.github/workflows/sync-authorino-manifests.yaml) (https://github.com/Kuadrant/authorino-operator/pull/328) workflow, which regenerates and opens a PR with the updated operator manifests..

### DIFF
--- a/.github/workflows/sync-operator-manifests.yaml
+++ b/.github/workflows/sync-operator-manifests.yaml
@@ -1,0 +1,24 @@
+name: Sync operator manifests
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'install/manifests.yaml'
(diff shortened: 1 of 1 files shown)


guicassolato (approved):
https://github.com/Kuadrant/authorino-operator/pull/328 needs to go first.

PR #644, opened 2026-06-30
Add CodeQL workflow for SAST scanning

averevki:
## Summary
  - Enable CodeQL static analysis to scan Go code for security vulnerabilities and coding errors
  - Runs on pushes and PRs to `main`, plus a weekly scheduled scan
  - Addresses the [OpenSSF Scorecard SAST check](https://github.com/ossf/scorecard/blob/c395761df6afe1a69e476bc60a013a94bcbc153f/docs/checks.md#sast) (currently 0/10)

Part of the work on https://github.com/Kuadrant/kuadrant-operator/issues/2038

## Summary by CodeRabbit

* **Chores**
  * Added automated CodeQL security scanning for pushes, pull requests, merge-group checks, and weekly scheduled runs.
  * Expanded scan coverage across multiple languages with language-specific build handling.
  * Improved scan reporting by enabling publication of security results through GitHub Actions.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,49 @@
+# generated from the template https://github.com/actions/starter-workflows/blob/1035244887e26fbbd4f1017d919fb5995cc521c4/code-scanning/codeql.yml
+name: CodeQL
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+  merge_group:
+    types: ["checks_requested"]
+  schedule:
+    - cron: '40 15 * * 6' # random schedule to avoid hitting the rate limit of the CodeQL and GitHub Actions API
+  workflow_dispatch:
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      packages: read
+      actions: read
+      contents: read
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - language: actions
+            build-mode: none
+          - language: go
+            build-mode: autobuild
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,105 @@
...
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  merge_group:
+    types: [ "checks_requested" ]
+  schedule:
+    - cron: '40 15 * * 6'
```
guicassolato: Every Saturday at 3:40pm, is that it? What's the rationale behind running on a schedule vs only on pushes to `main` and PRs? Is it because of externalities such as updates to the database of vulnerabilities?

Should we (also) have `workflow_dispatch`?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,105 @@
...
+    - name: Run manual build steps
+      if: matrix.build-mode == 'manual'
+      shell: bash
+      run: |
+        echo 'If you are using a "manual" build mode for one or more of the' \
+          'languages you are analyzing, replace this with the commands to build' \
+          'your code, for example:'
+        echo '  make bootstrap'
+        echo '  make release'
+        exit 1
```
guicassolato: I understand this probably came from a template. It's a bit odd though that we keep a step whose condition will never match – until somebody changes line 51 at least. Should we disable it by commenting the lines altogether perhaps?

.github/workflows/codeql.yml
averevki: Yes I would say so. This is all coming from the template, I only added the merge_group. Added the workflow_dispatch now as well :+1: 

The random weekly cron time was generated from the template, and should help to avoid waiting for GitHub runners to become available, by distributing the load.

.github/workflows/codeql.yml
averevki: You are right, I cleaned up the comments from the template, and also removed the boilerplate code we don't need from there

guicassolato (approved):

PR #647, opened 2026-07-07
Add TLS min version and cipher suite CLI flags

philbrookes:
## Summary
- Adds `--tls-min-version` and `--tls-cipher-suites` CLI flags to Authorino
- New `pkg/tlsconfig` package handles version string parsing and IANA cipher suite name lookup
- All three TLS-serving ports (gRPC 50051, HTTP ext-auth 5001, OIDC 8083) use the configured settings
- Defaults remain unchanged (TLS 1.2, Go default cipher suites) when flags are not provided

## Context
Part of OCP5 TLS profile compliance. The kuadrant-operator reads the cluster's `APIServer` CR and propagates TLS settings through the Authorino CR to these flags.

## Test plan
- [ ] Unit tests for `pkg/tlsconfig` (ParseMinVersion, ParseCipherSuites, BuildTLSConfig)
- [ ] Verify default behavior unchanged when flags omitted
- [ ] Verify explicit flags are applied to all TLS servers

## Summary by CodeRabbit

* **New Features**
  * Added CLI options to configure TLS minimum/maximum versions and cipher suites for external auth, OIDC discovery, and gRPC endpoints.
* **Bug Fixes**
  * Replaced hardcoded TLS settings with configurable TLS configs, including validation that the minimum TLS version does not exceed the maximum.
* **Tests**
  * Added unit tests covering TLS version/cipher-suite parsing and TLS config defaults and validation errors.


main.go
```diff
@@ -185,6 +188,8 @@ func authServerCmd(opts *authServerOptions) *cobra.Command {
 	cmd.PersistentFlags().StringVar(&opts.tlsCertPath, "tls-cert", utils.EnvVar("TLS_CERT", ""), "Path to the public TLS server certificate file in the file system - authorization server")
 	cmd.PersistentFlags().StringVar(&opts.tlsCertKeyPath, "tls-cert-key", utils.EnvVar("TLS_CERT_KEY", ""), "Path to the private TLS server certificate key file in the file system - authorization server")
 	cmd.PersistentFlags().IntVar(&opts.oidcHTTPPort, "oidc-http-port", utils.EnvVar("OIDC_HTTP_PORT", 8083), "Port number of OIDC Discovery server for Festival Wristband tokens")
+	cmd.PersistentFlags().StringVar(&opts.tlsMinVersion, "tls-min-version", utils.EnvVar("TLS_MIN_VERSION", ""), "Minimum TLS version for all TLS servers (VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13)")
```
guicassolato: I guess "VersionTLSXY" follows Kubernetes API Server format for specifying the TLS version, which also matches the names of the constants in the Golang package. I wonder if UX-wise it's as intuitive as a plain "X.Y".

Additionally, should we also support max version? The marginal effort seems small. WDYT?

philbrookes: Do you think it's likely that these command line arguments would be used outside of the reason we are adding them? This impacts my answers here:

If not: Better to match the version string with what is used in the Config CRD, and as that CRD also has no max version, not add the max version command line argument.

If so: I don't have a strong opinion then, I suppose `--tls-min-version=11` reads a little better to a human. Regarding the max version, what is the use-case for restricting new TLS versions?

guicassolato: > Do you think it's likely that these command line arguments would be used outside of the reason we are adding them?

I think we have to assume so, even if that's not how _we_ are planning to use it.

> as that CRD also has no max version, not add the max version command line argument

I would actually flip this the other way around. If there's a use case for it – I believe so – then, we need it at the operand (as a command-line arg). Once it's there, then we bubble it up to the Operator-provided CRD _because_ it's an optional we want to enable.

> `--tls-min-version=11` reads a little better to a human.

Do you think `--tls-min-version=1.1` could work? I'm thinking to avoid possible confusion between "1.1" and "11(.0)".

philbrookes: > Do you think `--tls-min-version=1.1` could work? I'm thinking to avoid possible confusion between "1.1" and "11(.0)".

Yes, this was just me being stupid.

Ok, this sounds sensible, I'll change it to work this way.

pkg/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,68 @@
...
+	for _, cs := range tls.CipherSuites() {
+		m[cs.Name] = cs.ID
+	}
+	for _, cs := range tls.InsecureCipherSuites() {
+		m[cs.Name] = cs.ID
+	}
+	return m
+}
+
+var cipherLookup = buildCipherLookup()
```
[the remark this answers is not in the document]
philbrookes: The inclusion of tls.InsecureCipherSuites() is intentional. This parser is used to apply TLS profiles configured by the cluster admin via the OpenShift APIServer CR. The "Old" profile deliberately includes ciphers like TLS_RSA_WITH_3DES_EDE_CBC_SHA which Go classifies as insecure. If we reject those here, Authorino would silently drop ciphers from the admin's chosen profile rather than respecting their configuration.

main.go
```diff
@@ -530,14 +537,14 @@ func startExtAuthServerGRPC(authConfigIndex index.Index, opts authServerOptions)
 }
 
 func startExtAuthServerHTTP(authConfigIndex index.Index, opts authServerOptions) {
-	startHTTPService("auth", opts.extAuthHTTPPort, service.HTTPAuthorizationBasePath, opts.tlsCertPath, opts.tlsCertKeyPath, service.NewAuthService(authConfigIndex, timeoutMs(opts.timeout), opts.maxHttpRequestBodySize))
+	startHTTPService("auth", opts.extAuthHTTPPort, service.HTTPAuthorizationBasePath, opts.tlsCertPath, opts.tlsCertKeyPath, opts.tlsMinVersion, opts.tlsCipherSuites, service.NewAuthService(authConfigIndex, timeoutMs(opts.timeout), opts.maxHttpRequestBodySize))
 }
 
 func startOIDCServer(authConfigIndex index.Index, opts authServerOptions) {
-	startHTTPService("oidc", opts.oidcHTTPPort, service.OIDCBasePath, opts.oidcTLSCertPath, opts.oidcTLSCertKeyPath, &service.OidcService{Index: authConfigIndex})
+	startHTTPService("oidc", opts.oidcHTTPPort, service.OIDCBasePath, opts.oidcTLSCertPath, opts.oidcTLSCertKeyPath, opts.tlsMinVersion, opts.tlsCipherSuites, &service.OidcService{Index: authConfigIndex})
```
guicassolato: We have separate options for the Authz server TLS config and the OIDC discovery server TLS. Even though TLS min/max versions and Cipher suites do sound like a security requirement that would typically span across all listeners, I wonder if we shouldn't separate them here as well thinking on backward compatibility. One could want to bump the configs for the main listener while still works on migrating clients to support bumping the other listener. Make sense?

main.go
guicassolato: > Regarding the max version, what is the use-case for restricting new TLS versions?

Sorry, I realised I didn't answer this other question of yours.

A few use cases for setting a maximum TLS version typically are:
- gradual rollout of newer versions
- to avoid an intermediate (e.g. proxies, load balancers, firewalls) to break the connection when it doesn't support a newer version
- to momentarily disable a new version known to contain unresolved bugs
- for testing and validation

main.go
```diff
@@ -161,6 +169,21 @@ func main() {
 	}
 }
 
+func envVarStringSlice(key string) []string {
```
guicassolato: Maybe move this func to [`pkg/utils/envvar.go`](https://github.com/Kuadrant/authorino/blob/main/pkg/utils/envvar.go)?

guicassolato: Also OK to drop the env vars altogether if you prefer. We've been favouring command-line flags, keeping the env vars for backwards compatibility anyway. `log-redact-*` flags, for example, are string array vars which we never defined env vars for.

guicassolato (dismissed):
LGTM. Thanks, @philbrookes!

maksymvavilov (commented):
Should we also update the doc? https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/observability.md

philbrookes:
> Should we also update the doc? https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/observability.md

good catch @maksymvavilov I'll do that.

philbrookes:
> > Should we also update the doc? https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/observability.md
> 
> good catch @maksymvavilov I'll do that.

@maksymvavilov I've updated the docs (not much to update though, really) here: https://github.com/Kuadrant/authorino/compare/959015124d3bd2f4a3e8794bd0747ac8180e3dfc..cbbc2211fd3bb0114ffeb076c7a00b73d4bba261

If you can review / approve.

maksymvavilov (approved):

Issue #656, opened 2026-07-15
Add `checkIssuerClaim` to enforce the JWT `iss` claim on the `issuerUrl` path


Aman-Cool:
While reading through the JWT authentication code, I noticed that when you configure `issuerUrl`, Authorino verifies the token's signature and expiry.., but it never checks that the token's `iss` claim actually matches the issuer it discovered. I asked about it, and it was confirmed this is intentional: issuer-checking is opt-in today, done via an authorization rule (CEL / pattern-matching / OPA), and kept consistent with the `jwksUrl` path on purpose.

That makes sense. The one place it can bite is when two different issuers happen to **share signing keys**; e.g. a multi-tenant IdP, or Festival Wristbands issued by two AuthConfigs pointing at the same signing-key Secret. In that case a token minted by issuer A can satisfy a config trusting issuer B, because it's only the signature being checked and the keys match. And it's quiet... the token lands as a valid identity, and nothing in the logs hints at it. Someone who hasn't added the authorization rule wouldn't know.

So this isn't a bug report: it's the RFE that came out of that discussion: give people a way to enforce the issuer directly, without needing a separate authorization rule.

**The proposal:**

```yaml
authentication:
  my-oidc:
    jwt:
      issuerUrl: https://issuer.example.com
      checkIssuerClaim: true   # new field, defaults to false
```

- `checkIssuerClaim` defaults to `false`, so nothing changes for anyone on upgrade.
- When `true`, Authorino rejects any token whose `iss` doesn't match `issuerUrl`, right at the authentication phase. (The discovered OIDC provider already knows the expected issuer, so it's basically letting the library do the check it was going to do anyway.)
- No effect on `jwksUrl`, which has no expected issuer to compare against.
- To make the opt-in nature discoverable, an INFO log at reconciliation when an `issuerUrl` config is used without `checkIssuerClaim`, plus a note in the docs.

I've got this working locally already (field + verifier change + docs + the reconciliation log), with tests covering both settings and a passing `make e2e` against real Keycloak confirming normal tokens still authenticate. Happy to open the PR right after this.

One thing I left out on purpose: there was also an idea for `checkClientID` / `clientID` for audience validation. That felt like a separate enough concern that I kept it out of this to keep the change focused.., glad to do it as a follow-up if that's wanted.

Thanks for @guicassolato for his help on this :)

guicassolato:
[...]

PR #657, opened 2026-07-15
feat(jwt): add `jwt.issuer` to enforce the `iss` claim on both `issuerUrl` and `jwksUrl`

Aman-Cool:
Implements #656.

Adds an optional `jwt.issuer` field to JWT authentication.., the expected value of the `iss` claim. Leave it unset (the default) and nothing changes on upgrade; set it and Authorino rejects, right at the authentication phase, any token whose `iss` doesn't equal it, with no separate authorization rule needed.

Following the review, this uses an explicit `issuer` string rather than a boolean, so it: enforces on the `jwksUrl` path too (not just `issuerUrl`); lets the operator choose the expected value rather than deriving it from discovery; and supports the cluster-internal-discovery / external-issuer split via `oidc.InsecureIssuerURLContext` (discovery/JWKS still fetched from `issuerUrl`, `iss` pinned to `issuer`). It also leaves room for `audiences: [...]` (the `aud` claim) later without piling on booleans.

Also adds an INFO log at reconciliation... uniformly for both `issuerUrl` and `jwksUrl`.., when `issuer` is omitted, plus a note in the docs.

Tests cover both paths (foreign issuer rejected when set, accepted when unset, matching issuer accepted, and the internal/external split enforced). Verified with `make test` and a full `make e2e` against real Keycloak; normal tokens still authenticate.

The `aud`-claim / audience validation is intentionally left out to keep this focused; tracked as a follow-up.

## Summary by CodeRabbit

- **New Features**
  - Added configurable JWT issuer validation for both OpenID Connect and JWKS authentication.
  - Tokens are rejected when their `iss` claim does not match the configured issuer.
  - Supports issuer validation when discovery and token issuer URLs differ.

- **Documentation**
  - Updated configuration schema and documentation with issuer settings and default behaviour.

- **Bug Fixes**
  - Improved JWT verifier consistency and handling of issuer-specific validation.


Aman-Cool:
@guicassolato, Following the thread from #656..., while wiring up `checkIssuerClaim`, I kept looking at the `aud` claim sitting right next to `iss`, unchecked for the same reason (`SkipClientIDCheck: true` in both verifier configs). It's the same story as the issuer one: a token minted for service A authenticates fine against service B when both trust the same issuer, and today the only fix is a hand-written CEL rule in the authorization phase... which trips over `aud` being a string-or-array, returns 403 instead of 401, and fires metadata calls on an identity that should've been rejected.

The interesting part: Authorino already validates this exactly one evaluator over; `kubernetesTokenReview.audiences`, just not for JWTs.

So the natural follow-up feels like an `audiences []string` on `JwtAuthenticationSpec`: opt-in, empty = no check (backward compatible), "any match" semantics, enforced post-verify so it covers both `issuerUrl` and `jwksUrl`.

That's a slightly different shape than the `checkClientID` / `clientID` you originally sketched.., I leaned toward the list because it matches the sibling `kubernetesTokenReview.audiences` field and handles multi-valued `aud` naturally, but I might be missing why you preferred the single-`clientID` form. Which shape would you rather see? Happy to build whichever, same opt-in + INFO-log-on-unsafe-default pattern as this PR.

Aman-Cool:
Good callout on the frequency, but I'd like to keep this at `Info` on purpose. This was a specifically requested signal — the whole point is that operators *see* the unsafe default without turning on debug logging (`V(1)` is debug here and off by default, so it'd effectively hide it). It also mirrors the existing Rego-v0 deprecation warning a few functions down (`translateAuthConfig` → `.Info(...)`), which is the same "your AuthConfig relies on a deprecated/unsafe default" case at the same level. Reconciliation is event-driven rather than a hot loop, so it's one line per change, consistent with that precedent. Happy to revisit if it turns out noisier in practice than the Rego warning.

controllers/auth_config_controller.go
```diff
@@ -381,7 +381,10 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
 		case api.JwtAuthentication:
 			var jwtVerifier identity_evaluators.JWTVerifier
 			if identity.Jwt.IssuerUrl != "" {
-				jwtVerifier = identity_evaluators.NewOIDCProviderVerifier(ctx, identity.Jwt.IssuerUrl, identity.Jwt.TTL, identity.Jwt.Timeout)
+				if !identity.Jwt.CheckIssuerClaim {
+					log.FromContext(ctxWithLogger).Info("JWT authentication with issuerUrl does not verify the token issuer claim by default; set checkIssuerClaim to true or enforce the issuer via an authorization rule", "authentication", identityCfgName, "issuerUrl", identity.Jwt.IssuerUrl)
+				}
```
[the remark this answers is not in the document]
Aman-Cool: Good callout on the frequency, but I'd like to keep this at `Info` on purpose. This was a specifically requested signal.., the whole point is that operators *see* the unsafe default without turning on debug logging (`V(1)` is debug here and off by default, so it'd effectively hide it). It also mirrors the existing Rego-v0 deprecation warning a few functions down (`translateAuthConfig` -> `.Info(...)`), which is the same "your AuthConfig relies on a deprecated/unsafe default" case at the same level. Reconciliation is event-driven rather than a hot loop, so it's one line per change, consistent with that precedent. Happy to revisit if it turns out noisier in practice than the Rego warning.

pkg/evaluators/identity/jwt.go
```diff
@@ -24,7 +24,19 @@ const (
...
+// held by the provider, which binds the verifier to it. When false (the default, kept for
+// backwards compatibility) the check is skipped and callers may enforce the issuer via an
+// authorization rule instead.
+func oidcTokenVerifierConfig(checkIssuerClaim bool) *oidc.Config {
+	return &oidc.Config{SkipClientIDCheck: true, SkipIssuerCheck: !checkIssuerClaim}
+}
+
+// jwksTokenVerifierConfig is used when only a raw jwksUrl is configured. There is no
+// expected issuer in that API shape, so the issuer check must always be skipped.
+var jwksTokenVerifierConfig = &oidc.Config{SkipClientIDCheck: true, SkipIssuerCheck: true}
```
guicassolato: These two functions could probably be simplified to a `oidcConfig(checkIssuerClaim bool) *oidc.Config` method of `JWTVerifier`. While `oidcProviderVerifier` can pass `v.checkIssuerClaim` in the call, `jwksVerifier` will pass a static `false`.

Or, depending on API redesign proposal, `oidcConfig(issue string) *oidc.Config`. Then, both verifiers would pass `v.issuer` in the call.


WDYT?

api/v1beta3/auth_config_types.go
```diff
@@ -384,6 +384,14 @@ type JwtAuthenticationSpec struct {
 	// +optional
 	IssuerUrl string `json:"issuerUrl,omitempty"`
 
+	// Whether to verify that the "iss" (issuer) claim of the JWT matches the issuer discovered from issuerUrl.
+	// Only applies when issuerUrl is set; it has no effect with jwksUrl, which has no expected issuer to compare against.
+	// Defaults to false for backwards compatibility, in which case the issuer claim can still be checked via an authorization rule (CEL, pattern-matching or OPA).
+	// Enable it to have Authorino reject, at the authentication phase, any token whose issuer differs from the configured issuerUrl.
+	// +optional
+	// +kubebuilder:default:=false
+	CheckIssuerClaim bool `json:"checkIssuerClaim,omitempty"`
```
guicassolato: A boolean `checkIssuerClaim` that implicitly derives its comparison value from `issuerUrl` has two problems:
1. It doesn't work when `issuerUrl` and the token's `iss` legitimately differ. A common deployment pattern is fetching the OIDC discovery document from within the cluster (`http://keycloak.keycloak.svc.cluster.local:8080/realms/demo`) while the IdP stamps tokens with the external URL (`https://keycloak.example.com/realms/demo`). Enabling `checkIssuerClaim` here would reject perfectly valid tokens — the user enables it expecting protection and gets false rejections with no recourse.
2. It doesn't extend to `jwksUrl`. The field is silently ignored on the `jwksUrl` path, which creates an inconsistency. A user who sets `checkIssuerClaim: true` with `jwksUrl` would believe they have issuer enforcement when they don't.

**Alternative:** a separate `issuer` field. Instead of a boolean, consider an explicit `issuer string` field — presence implies enforcement, absence preserves current behavior:

```yaml
# cluster-local discovery, external issuer in tokens
jwt:
  issuerUrl: "http://keycloak.keycloak.svc.cluster.local:8080/realms/demo"
  issuer: "https://keycloak.example.com/realms/demo"

# jwksUrl — works the same way
jwt:
  jwksUrl: "https://example.com/.well-known/jwks.json"
  issuer: "https://example.com"
```

This also extends naturally to other claim checks in the future (e.g., `audiences: [...]` for the `aud` claim) without accumulating booleans.

controllers/auth_config_controller.go
```diff
@@ -381,7 +381,10 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
...
+				jwtVerifier = identity_evaluators.NewOIDCProviderVerifier(ctx, identity.Jwt.IssuerUrl, identity.Jwt.CheckIssuerClaim, identity.Jwt.TTL, identity.Jwt.Timeout)
```
guicassolato: With the `issuer`-as-a-string-field approach described above, this call would pass the explicit expected issuer value (or empty string for "don't check") rather than a boolean. That naturally resolves the question of _what_ to compare against — the caller decides, not the OIDC discovery document.

pkg/evaluators/identity/jwt.go
```diff
@@ -120,7 +134,7 @@ func (v *oidcProviderVerifier) Verify(ctx gocontext.Context, rawIDToken string)
 	v.mu.RLock()
 	defer v.mu.RUnlock()
 
-	idToken, err := provider.Verifier(tokenVerifierConfig).Verify(ctx, rawIDToken)
+	idToken, err := provider.Verifier(oidcTokenVerifierConfig(v.checkIssuerClaim)).Verify(ctx, rawIDToken)
```
guicassolato: nit: `oidcTokenVerifierConfig(…)` allocates a new `*oidc.Config` on every `Verify()` call (i.e., every inbound request). Since the config is immutable after construction, consider precomputing and storing the config on the `oidcProviderVerifier` struct.

guicassolato (commented):
@Aman-Cool, I would like to challenge the API shape a little bit before this merges – since CRD fields are hard to change once released.

Leaving for your consideration a few comments throughout the code explaining the idea.

Aman-Cool:
@guicassolato, Really glad you pushed on this.., "CRD fields are forever" is exactly the right reason to slow down here, and honestly your shape is better than mine🧐. I went through each note and even ran the go-oidc bits I wasn't 100% sure about, and I'm sold on the `issuer` string. Thanks for taking the time :)

pkg/evaluators/identity/jwt.go
Aman-Cool: +1, and the second option is the one I'd take. Once both verifiers carry an `issuer`, they collapse into a single `oidcConfig(issuer string)` with each just passing v.issuer. Best part: handing `jwksVerifier` an issuer of its own is the very thing that kills the silent no-op you flag below.., so this isn't just tidier, it's the same fix wearing a different hat.

api/v1beta3/auth_config_types.go
Aman-Cool: You got me on this one😳, `checkIssuerClaim`: true with `jwksUrl` really is a silent no-op today, and a user would absolutely think they're covered when they aren't. The string fixes it cleanly, because the jwks path can enforce a plain issuer directly. 
**On #1 I went and ran it before agreeing**: today's `NewProvider(issuerUrl)` actually rejects that internal/external split at construction (`issuer did not match`), so it fails at reconcile rather than false-rejecting tokens, and `oidc.InsecureIssuerURLContext(issuer)` turns out to be the tidy way to make the split work while still checking **iss**. So... sold: absence = no check, presence = enforce, one field on both paths, and `audiences []string` slots in right beside it later. **I'll rework to this.**

controllers/auth_config_controller.go
Aman-Cool: Yep... this just forwards the `issuer` string (or `""` for "don't check") and lets the verifier decide, so the controller stays out of the discovery-vs-token weeds. Small bonus I didn't see coming: the "unsafe default" log now fires the same way for both `issuerUrl` and `jwksUrl` when issuer is empty, whereas the boolean could only ever warn on the OIDC path.

pkg/evaluators/identity/jwt.go
Aman-Cool: agreed. I'll resolve the config once at construction and hang it on the struct; it's immutable and independent of the provider, so the periodic JWKS/OIDC refresh doesn't care, and `Verify()` stops allocating on every request.

Aman-Cool:
@guicassolato, Reworked the whole thing around your suggestion. The boolean is gone... it's now a single `issuer` string: unset means no check (unchanged on upgrade), set means reject any token whose `iss` doesn't match. That one field does everything the review asked for.., it enforces on `jwksUrl` too (no more silent no-op), the operator picks the value instead of it being derived from discovery, and when `issuer` differs from `issuerUrl` I pin it via `oidc.InsecureIssuerURLContext` so cluster-internal discovery with an external issuer just works. Collapsed the two config funcs into one `oidcConfig(issuer)` as you sketched, and the config is precomputed on the struct now (no per-request alloc). Two CodeRabbit nits folded in on the way: the jwks verifier is built once too, and the pointless `RLock` around `Verify` is gone.

Tested end to end: unit tests for both paths: matching/foreign/unset plus the internal/external split; passing under `-race`, full `make test`, lint clean, and a real-Keycloak `make e2e` confirming legit tokens still authenticate. All CI green.

The `aud`/audiences follow-up is intentionally left for its own PR. Ready when you are :)

api/v1beta3/auth_config_types.go
```diff
@@ -384,6 +384,14 @@ type JwtAuthenticationSpec struct {
 	// +optional
 	IssuerUrl string `json:"issuerUrl,omitempty"`
 
+	// The expected value of the "iss" (issuer) claim of the JWT.
+	// If set, Authorino rejects, at the authentication phase, any token whose "iss" claim does not equal this value.
+	// If omitted, the issuer claim is not verified — it can still be checked via an authorization rule (CEL, pattern-matching or OPA).
+	// Applies to both issuerUrl and jwksUrl. It usually matches issuerUrl, but may differ when the OpenID Connect discovery
+	// endpoint is reached at a different URL than the issuer stamped into the tokens (e.g. cluster-internal discovery vs external issuer).
```
guicassolato: I would probably skip this part of the explanation, to avoid confused users thinking the new field has any relation with `issuerUrl` when the two values are actually completely independent from each other.


```suggestion
	// Applies to both issuerUrl and jwksUrl.
```

pkg/evaluators/identity/jwt_issuer_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/kuadrant/authorino/pkg/httptest"
+
+	jose "github.com/go-jose/go-jose/v4"
+	"github.com/golang-jwt/jwt/v5"
+	"go.uber.org/mock/gomock"
+	"gotest.tools/assert"
+)
+
+const (
+	issuerTestServerHost = "127.0.0.1:9007"
```
guicassolato: I'm afraid this port number may conflict with our OPA unit tests 😞.

In fact, `github.com/kuadrant/authorino/pkg/auth/httptest.NewHttpServerMock` is basically a wrapper of Go's `net/http/httptest` that takes as input a network address (to bind the listener to) and a `HTTP path → HTTP response data` map that dictates the behaviour of the handler. We used this approach in the past to multiplex the requests essentially based on the port number instead of dynamic server URLs.

More recently though, we've been favouring using `net/http/httptest.NewServer` directly instead. It binds the listener to `:0` (OS-assigned ephemeral port), thus preventing port conflicts. E.g.: https://github.com/Kuadrant/authorino/blob/58fecc6cdec38376fa7dba5638f1f7ecb6964cd0/pkg/evaluators/metadata/generic_http_test.go#L207-L213

Your `newIdPMockWithIssuer` func would probably change to something like this:

```go
func newIdPMockWithIssuer(advertisedIssuer, jwks string) *gohttptest.Server {
    var server *gohttptest.Server
    server = gohttptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        switch r.URL.Path {
        case "/.well-known/openid-configuration":
            fmt.Fprintf(w, `{"issuer":%q,"jwks_uri":"%s/certs"}`, advertisedIssuer, server.URL)
        case "/certs":
            w.Header().Set("Content-Type", "application/json")
            fmt.Fprint(w, jwks)
        }
    }))
    return server
}
```

Of course, the tests would have to pass `server.URL` as the `issuerUrl` instead of the constant `trustedIssuer`.

pkg/evaluators/identity/jwt_issuer_test.go
```diff
@@ -0,0 +1,199 @@
...
+	authCredMock.EXPECT().
+		GetCredentialsFromAuthReq(jwtAuthenticationRequestMock.GetAttributes().GetRequest().GetHttp()).
+		Return(rawToken, nil)
+
+	return evaluator.Call(pipelineMock, context.TODO())
+}
+
+// issuerUrl path, issuer set: a token signed by a key in the configured provider's JWKS
+// but whose `iss` names a different issuer must NOT authenticate.
+func TestOIDCProviderVerifier_IssuerSet_RejectsForeignIssuer(t *testing.T) {
```
guicassolato: Do you mind moving the tests in this file into `pkg/evaluators/identity/jwt_test.go` please?

pkg/evaluators/identity/jwt_issuer_test.go
```diff
@@ -0,0 +1,199 @@
...
+	assert.NilError(t, err)
+	assert.Equal(t, obj.(map[string]any)["iss"].(string), externalIssuer)
+
+	obj, err = callWithToken(t, verifier, signToken(t, key, foreignIssuer))
+	assert.Check(t, obj == nil, "wrong-issuer token accepted in the internal/external split configuration")
+	assert.ErrorContains(t, err, "issued by a different provider")
+}
+
+// jwksUrl path, issuer set: `iss` is enforced here too — the exact behavior the boolean design
+// could never provide on the raw-JWKS path.
```
guicassolato: We probably don't need the mention to a previous design that never actually landed.

```suggestion
// jwksUrl path, issuer set: `iss` is enforced
```

guicassolato (commented):
Thanks for putting this together @Aman-Cool! It's really an excellent work! 🏅

Leaving here just a few nits, but it looks like we're very close to getting this one in.

Meanwhile, for the people interested in verifying the PR manually, these are the steps I followed roughly:

### Verification steps

```sh
make local-setup FF=1 DEPLOY_KEYCLOAK=1

```

**issuerUrl = internal, issuer = ∅, token → external ⇒ 200**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
EOF
```

```sh
ACCESS_TOKEN=$(curl http://keycloak.127.0.0.1.nip.io:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=john' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

```sh
# 200
```

**issuerUrl = internal, issuer = external, token → external ⇒ 200**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        issuer: http://keycloak.127.0.0.1.nip.io:8080/realms/kuadrant
EOF
```

```sh
# 200
```

**issuerUrl = internal, issuer = internal, token → external ⇒ 401**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        issuer: http://keycloak.default.svc.cluster.local:8080/realms/kuadrant
EOF
```

```sh
# 401
```

**issuerUrl = internal, issuer = internal, token → internal ⇒ 200**

```sh
ACCESS_TOKEN=$(kubectl run token --attach --rm --restart=Never -q --image=curlimages/curl -- http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=john' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

```sh
# 200
```

**jwksUrl = internal, issuer = ∅, token → external ⇒ 200**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        jwksUrl: http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/certs
EOF
```

```sh
ACCESS_TOKEN=$(curl http://keycloak.127.0.0.1.nip.io:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=john' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

```sh
# 200
```

**jwksUrl = internal, issuer = external, token → external ⇒ 200**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        jwksUrl: http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/certs
        issuer: http://keycloak.127.0.0.1.nip.io:8080/realms/kuadrant
EOF
```

```sh
# 200
```

**jwksUrl = internal, issuer = internal, token → external ⇒ 401**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        jwksUrl: http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/certs
        issuer: http://keycloak.default.svc.cluster.local:8080/realms/kuadrant
EOF
```

```sh
# 401
```

**jwksUrl = internal, issuer = external, token → internal ⇒ 200**

```sh
ACCESS_TOKEN=$(kubectl run token --attach --rm --restart=Never -q --image=curlimages/curl -- http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=john' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

```sh
# 200
```

Aman-Cool:
Thanks for the review @guicassolato😄 applied both doc/comment suggestions.

pkg/evaluators/identity/jwt_issuer_test.go
Aman-Cool: good catch, and it's a real one: `pkg/evaluators/authorization/opa_test.go:24` already binds `127.0.0.1:9007`, so with packages running in parallel this would have flaked in CI rather than locally. Switching to `httptest.NewServer` with the OS-assigned port as you suggested, following the `generic_http_test.go` precedent, so the tests stop depending on a hand-maintained port registry entirely.

pkg/evaluators/identity/jwt_issuer_test.go
Aman-Cool: will do. I'll keep the existing tests in that file on their current `oidcServerHost` setup rather than converting them too, so the diff stays scoped to this PR; happy to migrate them in a follow-up if you'd like the whole file on ephemeral ports.

Aman-Cool:
@guicassolato..,in your proposed signature, `newIdPMockWithIssuer(advertisedIssuer, jwks string)` takes the advertised issuer as a parameter, but for the tests where the advertised issuer is meant to *match* the discovery URL, that value is `server.URL`... which doesn't exist until after the constructor returns. The `jwks_uri` in your snippet sidesteps this by reading `server.URL` from the closure at request time, but `advertisedIssuer` can't be passed the same way.

Options I see:
1. Treat `advertisedIssuer == ""` as "advertise `server.URL`", and pass an explicit value only for the mismatch test (`TestOIDCProviderVerifier_IssuerDiffersFromIssuerUrl_EnforcesConfiguredIssuer`, which needs the discovery URL and the advertised `iss` to genuinely differ; that's the case exercising the `oidc.InsecureIssuerURLContext` path in `getOpenIdProvider`).
2. Two helpers: one that self-advertises, one that takes an explicit issuer.

I'd lean towards (1) since it keeps a single helper and the mismatch case stays explicit at the call site, but it's your test conventions... let me know which you prefer and I'll push the refactor.

Aman-Cool:
Done: moved the issuer tests into `jwt_test.go`.

Still open: the `httptest.NewServer` port fix — waiting on your call re the `newIdPMockWithIssuer` signature (my question above).

pkg/evaluators/identity/jwt_issuer_test.go
guicassolato: Continuing in the thread wrt https://github.com/Kuadrant/authorino/pull/657#issuecomment-5026835233:

Nice catch, @Aman-Cool! Indeed, I hadn't tested that snipped. It was something of the top of my head. Thanks for spotting the limitation.

I guess option 1 is reasonable. The func would then probably go:

```go
func newIdPMockWithIssuer(advertisedIssuer, jwks string) *gohttptest.Server {
    var server *gohttptest.Server
    server = gohttptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        iss := advertisedIssuer
        if iss == "" {
            iss = server.URL // self-advertise
        }
        switch r.URL.Path {
        case "/.well-known/openid-configuration":
            fmt.Fprintf(w, `{"issuer":%q,"jwks_uri":"%s/certs"}`, iss, server.URL)
        case "/certs":
            w.Header().Set("Content-Type", "application/json")
            fmt.Fprint(w, jwks)
        }
    }))
    return server
}
```

Then, called as follows:
- **Common case (self-advertises):** `newIdPMockWithIssuer("", jwks)`
- **Mismatch test only:** `newIdPMockWithIssuer(externalIssuer, jwks)`

Again, I haven't tested it, but make sense?

BTW, since we won't need the `trustedIssuer` constant anymore, if you want to name the func just `newIdPMock`, it works for me. But it's your call.

Aman-Cool:
@coderabbitai , have a look

pkg/evaluators/identity/jwt_issuer_test.go
guicassolato: @Aman-Cool, do you still plan to address this one? I see the [test](https://github.com/Aman-Cool/authorino/blob/d62d2e119be9db2dfbe7182e4f81d7f4cf657cc0/pkg/evaluators/identity/jwt_test.go#L320) still uses Authorino `httptest`'s `NewHttpServerMock` with the listener of the test server bound to `127.0.0.1:9007`, conflicting with the OPA tests. As @coderabbit has [pointed out](https://github.com/Kuadrant/authorino/pull/657#issuecomment-5091076393) as well, using Go `httptest`'s `NewServer` would prevent that.

jasonmadigan:
👀

jasonmadigan:
think this looks good to merge, once the port collision thing discussed in https://github.com/Kuadrant/authorino/pull/657#discussion_r3615115766 is resolved

pkg/evaluators/identity/jwt_issuer_test.go
Aman-Cool: missed this @guicassolato 🥲; onto it rn...

Aman-Cool: Done.., switched the issuer-test mock to httptest.NewServer with the OS-assigned port, as you and CodeRabbit suggested, so it no longer binds 9007 and can't clash with the OPA tests. Went with the single newIdPMock (self-advertising server.URL as the issuer, or a different one for the split case), and the tests pass server.URL as the issuerUrl now.., no more fixed-port constant. Verified by running the identity and authorization packages together under -race. Thanks for the patience on this one 🙏

jasonmadigan (approved):

jasonmadigan:
thanks @Aman-Cool!

PR #666, opened 2026-07-29
Add two-phase release workflow

Boomatang:
# Summary

Introduces an automated two-phase release process for Authorino, replacing the previous manual git-tag-and-build approach. The new workflow splits every release into a **Pre-release** phase (branch creation, version bump, code generation, PR) and a **Release** phase (smoke tests, tagging, image build, GitHub Release), with a mandatory human review gate between them.

This change implements the design specified in [RFC: Two-Phase Release Workflow](https://github.com/Kuadrant/architecture/pull/178).

# Motivation

The previous release process was entirely manual: a maintainer checked out a ref, created a signed tag, pushed it, manually triggered the image build, and hand-wrote release notes. This was error-prone, undocumented in automation, and lacked pre-release validation.

The two-phase approach automates the mechanical steps while preserving human oversight at the critical review point before a release is finalized.

# What changed

## New workflows

- **Pre-release** (`pre-release.yaml`) — triggered manually with a target version. Creates the `release-X.Y` branch (if needed), opens a `pre-release-vX.Y.Z` branch that bumps `release.yaml`, runs `make generate` and `make manifests`, and opens a PR against the release branch.
- **Release** (`release.yaml`) — triggered manually with a release branch name. Reads the version from `release.yaml`, validates the branch matches, runs smoke tests (lint, unit tests, CEL tests), creates an annotated tag, builds and pushes the multi-arch container image via `build-images.yaml`, and creates the GitHub Release.
- **Version Gate** (`version-gate.yaml`) — runs on PRs targeting `release-**` branches when `release.yaml` changes. Validates that the version is not `0.0.0` on non-main branches and that any declared dependencies have published releases.

## New scripts

- `parse-version.sh` — extracts and validates semver components from `release.yaml`, used by the release workflow.
- `validate-release-yaml.sh` — validates `release.yaml` content and dependency versions, used by the version gate.

## Modified workflows

- **Build and push image** (`build-images.yaml`) — now supports `workflow_call` with `version` and `ref` inputs so the release workflow can invoke it directly. Image tagging logic updated to use the version from inputs when provided.

## New files

- `release.yaml` — version source-of-truth file at the repository root. Set to `0.0.0` on `main` (active development) and updated to the target version on release branches.

## Updated documentation

- `RELEASE.md` — rewritten to document the new two-phase process, including step-by-step instructions, artifact locations.

# Examples

- Pre-release workflow: https://github.com/Boomatang/authorino/actions/runs/28017001760
- Opened PR: https://github.com/Boomatang/authorino/pull/2
- Release workflow: https://github.com/Boomatang/authorino/actions/runs/28017819516
- GitHub release: https://github.com/Boomatang/authorino/releases/tag/v0.1.0

## Summary by CodeRabbit

- **New Features**
  - Added a guided two-phase pre-release and final release process with review checkpoints.
  - Added automated version validation, dependency checks, testing, image builds, smoke tests, tagging and release creation.
  - Added reusable workflows supporting configurable versions, branches and container images.
  - Added automated version gating for release branches.

- **Documentation**
  - Updated release guidance with the new process, artefacts, versioning rules and build behaviour.

- **Bug Fixes**
  - Improved safeguards against invalid versions, unavailable dependencies, draft releases and duplicate release tags.

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -1,0 +1,141 @@
+name: Pre-release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Release version (semver, e.g. 1.5.0)"
+        required: true
+        type: string
+      source-branch:
+        description: "Branch to base the pre-release changes on (default: main)"
+        required: false
+        type: string
+        default: "main"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.validate.outputs.version }}
+      release-branch: ${{ steps.validate.outputs.release-branch }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # ratchet:actions/checkout@v4
+        with:
+          ref: ${{ inputs.source-branch }}
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Validate version format
+        id: validate
+        run: |
+          VERSION="${{ inputs.version }}"
+          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$ ]]; then
+            echo "::error::Invalid semver version: $VERSION"
+            exit 1
+          fi
+          MAJOR=$(echo "$VERSION" | cut -d. -f1)
+          MINOR=$(echo "$VERSION" | cut -d. -f2)
+          RELEASE_BRANCH="release-${MAJOR}.${MINOR}"
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "release-branch=$RELEASE_BRANCH" >> "$GITHUB_OUTPUT"
+
+      - name: Create or verify release branch
+        run: |
+          RELEASE_BRANCH="${{ steps.validate.outputs.release-branch }}"
+          if git ls-remote --exit-code origin "refs/heads/${RELEASE_BRANCH}" >/dev/null 2>&1; then
+            echo "Release branch '${RELEASE_BRANCH}' already exists"
+          else
+            echo "Creating release branch '${RELEASE_BRANCH}' from '${{ inputs.source-branch }}'"
+            git checkout -b "${RELEASE_BRANCH}"
+            git push origin "${RELEASE_BRANCH}"
+          fi
+
+  prepare-release:
+    needs: setup
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # ratchet:actions/checkout@v4
+        with:
+          ref: ${{ needs.setup.outputs.release-branch }}
+          fetch-depth: 0
+
+      - name: Create pre-release branch
+        run: |
+          PRE_RELEASE_BRANCH="pre-release-v${{ needs.setup.outputs.version }}"
+          if git ls-remote --exit-code origin "refs/heads/${PRE_RELEASE_BRANCH}" >/dev/null 2>&1; then
+            echo "::error::Pre-release branch '${PRE_RELEASE_BRANCH}' already exists. Delete it first or use a different version."
+            exit 1
+          fi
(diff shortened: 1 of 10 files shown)


.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,141 @@
...
+      - name: Commit and push changes
+        run: |
+          VERSION="${{ needs.setup.outputs.version }}"
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add -A
+          if git diff --cached --quiet; then
+            echo "No changes to commit"
+          else
+            git commit -m "chore: prepare release v${VERSION}"
```
guicassolato, line 102: Missing the sign-off. Won't this fail the DCO check?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,129 @@
...
+            echo "::error::Branch '${{ inputs.release-branch }}' does not match version ${{ steps.parse.outputs.version }} (expected branch: $EXPECTED_BRANCH)"
+            exit 1
+          fi
+
+      - name: Check for existing release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          VERSION="${{ steps.parse.outputs.version }}"
+          if gh release view "v${VERSION}" &>/dev/null; then
```
guicassolato, line 51: I remember a question raised about what to do in the case a *draft* release exists. I guess we want to want to discourage using those and fail the workflow just the same?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,129 @@
...
+            exit 1
+          fi
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git tag -a "$TAG" -m "Release $TAG"
+          git push origin "$TAG"
+
+  build-image:
+    needs: [read-version, tag]
+    uses: ./.github/workflows/build-images.yaml
```
guicassolato, line 105: Authorino has a ["Smoke Test"](https://github.com/Kuadrant/authorino/blob/main/.github/workflows/smoke-test.yaml) workflow that will trigger after building the image in CI. It deploys the image just built to a Kubernetes cluster and runs the e2e tests on it.

Unfortunately, it always deploys the latest image built from the `main` branch, [statically specified](https://github.com/Kuadrant/authorino/blob/58fecc6cdec38376fa7dba5638f1f7ecb6964cd0/.github/workflows/smoke-test.yaml#L60) in the workflow. It would be nice to make that work for release builds too in the future.

.github/workflows/pre-release.yaml
Boomatang: Could find documentation around this, but with the help of cluade found it the actions source. The check is skipped of merge commits, and for accounts that are marked as bots. Could there be a question over should that be case, possible, but that is why it wont fail the check. 

https://github.com/dcoapp/app/blob/main/lib/dco.js#L21-L25

.github/workflows/release.yaml
Boomatang: Yes, this is one thing that should be added. These hidden/triggered workflows are one of the reasons for doing this work. You need to know they exist in order to know to even check for them. I will work in so that workflow most pass in before making the GitHub release.

.github/workflows/release.yaml
Boomatang: Some times I have memory like a gold fish. Yes, we did say that I will update that to.

Boomatang: I had a second look at this. The current logic is not easy to update to support update a draft release. It would need a lot of rework to create the two branching logic's.  For now I have a message to say the draft release are unsupported.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,143 @@
...
+      - name: Check for existing release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          VERSION="${{ steps.parse.outputs.version }}"
+          if gh release view "v${VERSION}" &>/dev/null; then
+            echo "::error::GitHub Release v${VERSION} already exists"
+            if [ "$(gh release view "v${VERSION}" --json isDraft --jq '.isDraft')" = "true" ]; then
+              echo "Updating of draft releases is current not supported. Please remove existing drafts"
+            fi
```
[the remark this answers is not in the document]
Boomatang: done

PR #668, opened 2026-08-10
perf: improve authorino caching with coalescing concurrent requests

hanna836:
...

## Summary by CodeRabbit

* **Performance Improvements**
  * Concurrent authorization and metadata requests are now consolidated when they request the same data, reducing duplicate evaluations and external requests.
  * Results from consolidated requests continue to populate the cache, improving response times for subsequent requests.
  * Requests for different data remain independently processed.

* **Reliability**
  * Added coverage for concurrent request handling, separate cache keys, and cache population behaviour.

### DIFF
--- a/pkg/evaluators/authorization.go
+++ b/pkg/evaluators/authorization.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"golang.org/x/sync/singleflight"
+
 	"github.com/kuadrant/authorino/pkg/auth"
 	"github.com/kuadrant/authorino/pkg/evaluators/authorization"
 	"github.com/kuadrant/authorino/pkg/jsonexp"
@@ -23,6 +25,7 @@ type AuthorizationConfig struct {
 	Conditions jsonexp.Expression `yaml:"conditions"`
 	Metrics    bool               `yaml:"metrics"`
 	Cache      EvaluatorCache
+	flight     singleflight.Group
 
 	OPA             *authorization.OPA                 `yaml:"opa,omitempty"`
 	JSON            *authorization.JSONPatternMatching `yaml:"json,omitempty"`
@@ -48,35 +51,49 @@ func (config *AuthorizationConfig) GetAuthConfigEvaluator() auth.AuthConfigEvalu
 // impl:AuthConfigEvaluator
 
 func (config *AuthorizationConfig) Call(pipeline auth.AuthPipeline, ctx context.Context) (interface{}, error) {
-	if evaluator := config.GetAuthConfigEvaluator(); evaluator == nil {
+	evaluator := config.GetAuthConfigEvaluator()
+	if evaluator == nil {
 		return nil, fmt.Errorf("invalid authorization config")
-	} else {
-		logger := log.FromContext(ctx).WithName("authorization")
+	}
 
-		cache := config.Cache
-		var cacheKey interface{}
+	logger := log.FromContext(ctx).WithName("authorization")
 
-		if cache != nil {
-			cacheKey, _ = cache.ResolveKeyFor(pipeline.GetAuthorizationJSON())
-			if cacheKey != nil {
-				if cachedObj, err := cache.Get(cacheKey); err != nil {
-					logger.V(1).Error(err, "failed to retrieve data from the cache")
-				} else if cachedObj != nil {
-					return cachedObj, nil
+	cache := config.Cache
+	var cacheKey interface{}
+
+	if cache != nil {
+		cacheKey, _ = cache.ResolveKeyFor(pipeline.GetAuthorizationJSON())
+		if cacheKey != nil {
+			if cachedObj, err := cache.Get(cacheKey); err != nil {
+				logger.V(1).Error(err, "failed to retrieve data from the cache")
+			} else if cachedObj != nil {
+				return cachedObj, nil
+			}
+		}
+	}
+
+	if cache != nil && cacheKey != nil {
+		flightKey := fmt.Sprintf("%s/%v", config.Name, cacheKey)
+		result, err, shared := config.flight.Do(flightKey, func() (interface{}, error) {
+			if cachedObj, _ := cache.Get(cacheKey); cachedObj != nil {
+				return cachedObj, nil
+			}
+			obj, err := evaluator.Call(pipeline, log.IntoContext(ctx, logger))
+			if err == nil {
+				if setErr := cache.Set(cacheKey, obj); setErr != nil {
+					logger.V(1).Info("unable to store data in the cache", "err", setErr)
 				}
 			}
+			return obj, err
+		})
+		if shared {
+			logger.V(1).Info("singleflight: coalesced duplicate authorization call", "key", flightKey)
 		}
-
-		obj, err := evaluator.Call(pipeline, log.IntoContext(ctx, logger))
-
-		if err == nil && cacheKey != nil {
-			if err := cache.Set(cacheKey, obj); err != nil {
-				logger.V(1).Info("unable to store data in the cache", "err", err)
-			}
-		}
-
-		return obj, err
+		return result, err
 	}
+
+	obj, err := evaluator.Call(pipeline, log.IntoContext(ctx, logger))
+	return obj, err
 }
 
 // impl:NamedEvaluator
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 	go.uber.org/mock v0.5.2
 	go.uber.org/zap v1.25.0
 	golang.org/x/oauth2 v0.35.0
+	golang.org/x/sync v0.22.0
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260401024825-9d38bb4040a9
 	google.golang.org/grpc v1.80.0
 	google.golang.org/protobuf v1.36.11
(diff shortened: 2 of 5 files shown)


pkg/evaluators/authorization.go
```diff
@@ -48,35 +51,49 @@ func (config *AuthorizationConfig) GetAuthConfigEvaluator() auth.AuthConfigEvalu
...
+	}
 
-		obj, err := evaluator.Call(pipeline, log.IntoContext(ctx, logger))
-
-		if err == nil && cacheKey != nil {
-			if err := cache.Set(cacheKey, obj); err != nil {
-				logger.V(1).Info("unable to store data in the cache", "err", err)
+	if cache != nil && cacheKey != nil {
+		flightKey := fmt.Sprintf("%s/%v", config.Name, cacheKey)
+		result, err, shared := config.flight.Do(flightKey, func() (interface{}, error) {
```
guicassolato: What sets `config.flight`?

pkg/evaluators/authorization.go
```diff
@@ -48,35 +51,49 @@ func (config *AuthorizationConfig) GetAuthConfigEvaluator() auth.AuthConfigEvalu
...
 		}
+	}
 
-		obj, err := evaluator.Call(pipeline, log.IntoContext(ctx, logger))
-
-		if err == nil && cacheKey != nil {
-			if err := cache.Set(cacheKey, obj); err != nil {
-				logger.V(1).Info("unable to store data in the cache", "err", err)
+	if cache != nil && cacheKey != nil {
+		flightKey := fmt.Sprintf("%s/%v", config.Name, cacheKey)
```
guicassolato, line 76: Each config instance owns its own flight group. I wonder if using the evaluator's name as the prefix isn't redundant here.

guicassolato (commented):
First and foremost, thanks for the PR @hanna836. I imagine this may be more than an opportunity to contribute – which in itself we very much appreciate – but hopefully backed by a real use case we can document.

I do have a couple things I started wondering while reading the PR and the problem it proposes to solve, in face of the real life scenarios we know.

Cache keys are more often than not unique per request. In this situation, concurrent requests rarely collide on a key, so `flight.Do` almost never finds a duplicate and coalescing almost never fires. The feature is then mostly inert: near-zero cost, but also near-zero benefit. It only pays off under a specific condition — a coarse/shared cache key (tenant-scoped, static, low-cardinality) plus bursty concurrent traffic against a cold entry. Whether that condition occurs in real deployments is exactly the motivation question this PR should answer with a benchmark. I don't see one though. In other words, the gain in performance must be more than theoretical and I would love to see numbers on that. 

A second thought regards introducing cross-request coupling, something the plain cache does not suffer from. IIUC, the coalesced function runs under the leader's `ctx` (`log.IntoContext(ctx, logger)` captures the first caller's context). Followers block regardless of their own context, but based strictly on the matching flight key. So if the leader's request is cancelled mid-flight, the evaluator call can abort and the error/cancellation is propagated to every coalesced follower,  even though their own contexts were healthy. The followers share the leader's error and result indiscriminately. This is a new request-coupling failure mode we don't have to deal with today, traded for a benefit that only materializes in a narrow traffic pattern. And this coupling appears only when coalescing actually fires, which is the same narrow condition from my first point — so the feature is effectively either inert or introducing cross-request failure coupling, with little middle ground.

hanna836:
@guicassolato  thanks for the input. Yes, I agree that the solution only address same or low cardinality API keys that expire all at once, and then overflow MaaS-api server and DB. I agree that this is not a real world scenario, and that the solution provided here maybe, as you said, inert. So I am ok if we do not push this, however, we still need to discuss and address the underlying issue here where API keys expire at 60s mark, creating a flood of requests that MaaS-api cannot handle (results in requests being dropped). There needs to be a caching solution that works for high cardinality API keys, at least  if it expires in batches, to allow MaaS-api to keep up.

guicassolato:
@hanna836, thank you for bringing this one to light!

By "API key" here — this being a Red Hat OpenShift AI use case — we mean an external HTTP metadata call to a server that validates keys Authorino extracts from the context, not the API key authentication feature. That's a real, documentable shape. But I want to separate two problems:
- (a) a single hot key hit by many concurrent requests, which is the only thing coalescing can collapse, and
- (b) the flood you're describing — many distinct keys expiring together at the 60s mark.

singleflight can't touch (b): distinct keys never coalesce, so its benefit scales with concurrency-per-key, not flood size. For (b), the usual levers are TTL jitter/staggered expiry and stale-while-revalidate so refreshes don't stampede MaaS-api at once.

If there are hot keys in the real traffic, then (a) is worth pursuing — but it needs a benchmark measuring per-key concurrency at expiry, and I would like to see the coupling closed properly first (detached execution context and per-caller cancellation), so we're not trading a flood for cross-request failure coupling.

hanna836:
@guicassolato  this request was created as a result of such benchmark:

"but it needs a benchmark measuring per-key concurrency at expiry, and I would like to see the coupling closed properly first (detached execution context and per-caller cancellation), so we're not trading a flood for cross-request failure coupling."

Where TTL expiration of hot keys caused massive flood to MAAS-api. As a result of that, these requests were not able to be served within 200ms timeout, and so they got dropped. I think the discussion should focus around the caching techniques on Authorino regardless of low/high cardinality keys, so as to prevent such disruption. Given default is 60s this will be too frequent and quite disruptive under high load.

guicassolato:
Thanks @hanna836 — the 200ms drops against MaaS-api at TTL expiry are a useful, concrete signal.

Before we let "hot keys" steer us back to coalescing, I want to be careful about one thing: whether that hot-key concentration is a property of the real traffic or an artifact of how the load was modeled. A test that fires many concurrent requests at a single fixed cache key will always make coalescing look like a win — by construction there's exactly one upstream call to collapse to — while telling us little about production, where each consumer typically carries a distinct key and coalescing would be inert.

So for this to be decision-useful, the scenario needs to preserve the real key distribution: the number of *distinct* keys in flight around the expiry mark and the concurrency *per key*, not an aggregate rate against one key. If real traffic genuinely shows a few keys absorbing most of the concurrency (e.g. a shared service-account key), coalescing is on the table. If it's many distinct keys expiring together, the lever is different — desynchronizing expiry / serving stale while revalidating — and coalescing won't move the needle.

Proposal: close this PR and open an issue on caching techniques in Authorino to prevent TTL-mark floods — coalescing as one candidate alongside TTL jitter / staggered expiry and stale-while-revalidate — anchored on that key-distribution data.

One clarification on the TTL itself: Authorino doesn't cache by default and enforces no 60s TTL — that value is a choice on your side, so if the direction is tuning that number there's nothing for Authorino to do. Where Authorino *can* help is the expiry behavior (jitter, stale-while-revalidate), independent of whatever TTL you pick.

hanna836:
@guicassolato I am a little confused as to why the type of key matters here. When 60s is hit, all keys are gone from the cache. So this leaves us to the mercy of the workload itself: if at that time the burst of concurrent requests is very high, high enough to overload maas-api, then we are at the risk of dropping requests. This is exactly what I observed in my tests. So at this point does it matter what types of keys are in the cache if they get all wiped out? If they are high cardinality, they will all be referenced to the DB to populate. If low cardinality, this fix helps to keep requests with same keys waiting. So regardless of types and distribution of keys, the problem still exists, and yes 60s is configurable default, but it does not address any of these issues, it simply pushes them to later time. Currently this is the behavior observed:

Cache TTL=60s expires
    → ALL in-flight requests see cache miss (same key: Bearer token)
    → ALL call evaluator.Call() → HTTP POST to maas-api simultaneously
    → maas-api (200m CPU, 10 threads) gets overwhelmed
    → CFS scheduler throttles pod (200ms CPU budget per 1000ms period)
    → Tail requests take >200ms to resolve
    → WASM timeout (200ms) fires
    → failureMode: deny → HTTP 500 to client

We improved WASM timeout (it is now 5s) so that will help to keep the requests but will not help with latency.

I am ok to close this requests if we don't even want to improve this for low cardinality keys. Regardless of production scenarios, I don't think this fix hurts anything if added. But a separate request should discuss optimization on Authorino caching. Here I am pasting proposed mitigations from my report:

Screenshot 2026-09-17 at 1 07 55 PM

guicassolato:
@hanna836, thanks — the CFS-throttling breakdown is the most useful part of this thread, and it points the way.

Your trace answers the "why does key type matter" question itself: the flow assumes "same key: Bearer token" — one token shared across all concurrent requests, which is exactly why coalescing looks effective (one key → one upstream call to collapse to). In production each consumer carries its own bearer token, so at expiry you get N distinct keys, N distinct DB hits, and coalescing has nothing to collapse. The gain you measured is a property of the single-token test model, not the workload.

More to the point, your trace names the real failure: maas-api at 200m CPU getting CFS-throttled. That's a downstream capacity problem, and right-sizing maas-api (more CPU / replicas) is the simplest fix that holds regardless of key distribution — I'm surprised it isn't top of the list in the report.

On the four mitigations:

- Cache jitter (mitigation 4): my first choice on the Authorino side. It targets the flood directly by de-synchronizing expiry, so only a fraction of keys miss at any instant. Cheap and low-risk. Caveat: jitter spreads load; it doesn't stop an individual cold-miss request from blocking on the populate — it just reduces how many do so at once.

- Single-flight coalescing (mitigation 2): I'm closing this PR. Beyond the cardinality point above, where it does fire it introduces cross-request coupling — a cancelled or failed leader poisoning otherwise-healthy followers — unless reworked with a detached execution context and per-caller cancellation. So "doesn't hurt anything if added" doesn't hold even for the low-cardinality case.

- Stale-while-revalidate (mitigation 3): more than we need right now, mainly because you've already raised the WASM timeout to 5s, which removes the acute drop-at-200ms failure SWR would have prevented. It's the right tool if latency SLOs later demand never blocking on a miss, but it's a bigger lift (soft vs hard TTL, background refresh) and with synchronized soft-expiry the background refreshes still burst maas-api unless paired with jitter — so jitter comes first regardless.

- Longer default TTL + explicit invalidation (mitigation 1): reasonable for steady state, but it trades freshness and adds a maas-api→Authorino invalidation dependency, plus the security consideration that revoked keys stay honored until invalidated. Plus, as agreed before, the TTL value is configuration, so nothing to change in Authorino itself.

Proposal: close this PR and open a dedicated issue on Authorino caching, with jitter as the lead candidate – still framed as secondary to right-sizing maas-api, though.

PR #618, opened 2026-06-01
Fix credential leakage in debug logs

guicassolato:
## Summary

Fixes #617 - Prevents sensitive credentials from leaking into debug logs by implementing comprehensive redaction across all log statements.

## Problem

When debug logging is enabled (`--log-level=debug`), Authorino logged sensitive information in plain text, including:

- OAuth2 client credentials embedded in URLs
- Bearer tokens in form data during token introspection
- User credentials in Authorization headers and cookies
- API keys
- Kubernetes service account tokens
- **Complete Kubernetes Secret objects** (both base64-encoded data and plaintext in annotations)

This exposed credentials in centralized logging systems during troubleshooting, incident response, and security audits.

## Solution

Implemented automatic redaction that replaces sensitive data with `***REDACTED***` while preserving useful debugging information.

### What Gets Redacted

**Default redacted form fields:**
- `token`, `access_token`, `refresh_token`, `id_token`
- `client_secret`, `password`, `secret`
- `api_key`, `apikey`

**Default redacted HTTP headers:**
- `Authorization`, `Cookie`
- `X-API-Key`, `APIKey`

**Additionally redacted:**
- URL credentials (userinfo in URLs)
- Kubernetes TokenReview tokens
- Kubernetes Secret data (all keys in `data` and `stringData`)
- Secrets in `kubectl.kubernetes.io/last-applied-configuration` annotations
- Request bodies (form-encoded fields or complete body)
- Authorization pipeline evaluation data

### Configurable Redaction

New command-line flags allow customization:

```bash
--log-redact-add-field=custom_token          # Add custom sensitive field
--log-redact-remove-field=api_key            # Remove from defaults
--log-redact-add-header=X-Custom-Auth        # Add custom sensitive header
--log-redact-remove-header=cookie            # Remove from defaults
```

### Core Redaction Engine
- **pkg/log/redact.go** - Thread-safe redaction utilities with configurable sensitive fields/headers
- **pkg/log/redact_test.go** - Comprehensive test coverage including Kubernetes Secret redaction

### Log Statement Updates
- **pkg/service/auth_pipeline.go** - Redact identity validation and authorization evaluation logs
- **pkg/evaluators/identity/oauth2.go** - Redact OAuth2 token introspection requests
- **pkg/evaluators/metadata/uma.go** - Redact UMA PAT requests
- **pkg/evaluators/identity/kubernetes_auth.go** - Redact Kubernetes TokenReview tokens
[...]

PR #671, opened 2026-08-11
feat(service): enrich auth decision logs with configurable fields from filter metadata

andresllh:
## Summary

- Read `io.kuadrant.logging.fields` from the CheckRequest's filter metadata and append resolved key-value pairs to the info-level "outgoing authorization response" log line
- Mirrors the existing `metricLabels()` pattern for `io.kuadrant.metrics.labels` — supports plain strings, numbers, booleans, and deferred CEL expressions (e.g. `auth.identity.sub`) resolved against the post-auth pipeline context
- Identity fields resolve on both allow and deny decisions, and are gracefully omitted when authentication fails

## Motivation

Kuadrant's TelemetryPolicy currently configures metrics via `io.kuadrant.metrics.labels` filter metadata. Extending this to a `logging` domain (`io.kuadrant.logging.fields`) enables platform engineers to declare additional fields — such as `auth.identity.sub` — that appear in Authorino's per-request info-level logs without enabling debug verbosity.

This is the Authorino-side consumer for a planned TelemetryPolicy logging stanza (see [CONNLINK-1384](https://issues.redhat.com/browse/CONNLINK-1384)). The wasm plugin's domain-agnostic `buildMetadataContext` already groups bindings by domain, so a `logging.fields` domain in TelemetryPolicy would automatically produce the `io.kuadrant.logging.fields` metadata key with zero wasm plugin changes.

## Changes

- `pkg/service/auth_pipeline.go`: Add `loggingFields()` method that reads `io.kuadrant.logging.fields` from filter metadata (mirrors `metricLabels()`)
- `pkg/service/auth.go`: Extend `logAuthResult` to accept and append logging fields; extract fields via type assertion after `pipeline.Evaluate()` in `Check()`
- `pkg/service/auth_pipeline_test.go`: 5 new tests covering value types (string, number, bool, CEL, null), empty metadata, identity resolution on allow, identity resolution on authorization deny, and graceful handling on authentication failure

## Test plan

- [x] `TestPipelineLoggingFields` — all value types: CEL expressions, static strings, numbers, booleans, unresolvable CEL, null
- [x] `TestPipelineLoggingFieldsEmpty` — no filter metadata present → empty map
- [x] `TestLoggingFieldsResolvesIdentityOnAllow` — `auth.identity.anonymous` resolves to `"true"` after allowed request
- [x] `TestLoggingFieldsResolvesIdentityOnAuthzDeny` — identity still resolves after authorization deny (the key audit property)
- [x] `TestLoggingFieldsGracefulOnAuthnFailure` — identity field gracefully omitted when authentication fails; `request.*` fields still resolve
- [x] All 500 existing tests pass unchanged
- [x] `go build ./...` compiles cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Authentication results can include optional pipeline-provided logging fields.
  * Logging fields support static text, numbers, booleans and expressions evaluated against redacted authorisation data.
  * Logging-field support can be enabled through the `--enable-logging-fields` option or `ENABLE_LOGGING_FIELDS` environment variable.
  * Logging field values can be limited with `--logging-fields-max-value-bytes` or `LOGGING_FIELDS_MAX_VALUE_BYTES` (default: 1024; set to `0` to disable truncation).
  * Field names use the `custom.` prefix, and unsafe characters are escaped before values are recorded.

* **Bug Fixes**
  * Authentication logging consistently includes fields for successful, unavailable and not-found results.
  * Invalid or unsupported metadata is safely skipped without disrupting authentication.

### DIFF
--- a/pkg/service/auth.go
+++ b/pkg/service/auth.go
@@ -73,14 +73,15 @@ func init() {
 
 // AuthService is the server API for the authorization service.
 type AuthService struct {
-	Index                  index.Index
-	Timeout                time.Duration
-	MaxHttpRequestBodySize int64
-	EnableLoggingFields    bool
+	Index                      index.Index
+	Timeout                    time.Duration
+	MaxHttpRequestBodySize     int64
+	EnableLoggingFields        bool
+	LoggingFieldsMaxValueBytes int
 }
 
-func NewAuthService(index index.Index, timeout time.Duration, maxHttpRequestBodySize int64, enableLoggingFields bool) *AuthService {
-	return &AuthService{Index: index, Timeout: timeout, MaxHttpRequestBodySize: maxHttpRequestBodySize, EnableLoggingFields: enableLoggingFields}
+func NewAuthService(index index.Index, timeout time.Duration, maxHttpRequestBodySize int64, enableLoggingFields bool, loggingFieldsMaxValueBytes int) *AuthService {
+	return &AuthService{Index: index, Timeout: timeout, MaxHttpRequestBodySize: maxHttpRequestBodySize, EnableLoggingFields: enableLoggingFields, LoggingFieldsMaxValueBytes: loggingFieldsMaxValueBytes}
 }
 
 // ServeHTTP invokes authorization check for a simple GET/POST HTTP authorization request
@@ -312,7 +313,7 @@ func (a *AuthService) Check(parentContext gocontext.Context, req *envoy_auth.Che
 	var loggingFields map[string]string
 	if a.EnableLoggingFields {
 		if p, ok := pipeline.(*AuthPipeline); ok {
-			loggingFields = p.loggingFields()
+			loggingFields = p.loggingFields(a.LoggingFieldsMaxValueBytes)
 		}
 	}
 	a.logAuthResult(result, ctx, loggingFields)
--- a/pkg/service/auth.go
+++ b/pkg/service/auth.go
@@ -284,13 +284,13 @@ func (a *AuthService) Check(parentContext gocontext.Context, req *envoy_auth.Che
 	// If we couldn't find the AuthConfig in the config, we return and deny.
 	if authConfig == nil {
 		result := auth.AuthResult{Code: rpc.NOT_FOUND, Message: RESPONSE_MESSAGE_SERVICE_NOT_FOUND}
-		a.logAuthResult(result, ctx)
+		a.logAuthResult(result, ctx, nil)
 		return a.deniedResponse(result), nil
 	}
 
 	if err := context.CheckContext(ctx); err != nil {
 		result := auth.AuthResult{Code: rpc.UNAVAILABLE}
-		a.logAuthResult(result, ctx)
+		a.logAuthResult(result, ctx, nil)
 		context.Cancel(ctx)
 		span.RecordError(err)
 		span.SetStatus(otel_codes.Error, err.Error())
@@ -308,7 +308,11 @@ func (a *AuthService) Check(parentContext gocontext.Context, req *envoy_auth.Che
 		span.SetStatus(otel_codes.Error, err.Error())
 	}
 
-	a.logAuthResult(result, ctx)
+	var loggingFields map[string]string
+	if p, ok := pipeline.(*AuthPipeline); ok {
+		loggingFields = p.loggingFields()
+	}
+	a.logAuthResult(result, ctx, loggingFields)
 
 	if result.Success() {
 		return a.successResponse(result, ctx), nil
@@ -401,7 +405,7 @@ func (a *AuthService) logAuthRequest(req *envoy_auth.CheckRequest, ctx gocontext
 	}
 }
 
-func (a *AuthService) logAuthResult(result auth.AuthResult, ctx gocontext.Context) {
+func (a *AuthService) logAuthResult(result auth.AuthResult, ctx gocontext.Context, loggingFields map[string]string) {
 	logger := log.FromContext(ctx)
 	success := result.Success()
 	baseLogData := []interface{}{"authorized", success, "response", result.Code.String()}
@@ -415,11 +419,17 @@ func (a *AuthService) logAuthResult(result auth.AuthResult, ctx gocontext.Contex
 		}
 		logData = append(logData, "object", reducedResult)
 	}
+	for k, v := range loggingFields {
+		logData = append(logData, k, v)
+	}
 	logger.Info("outgoing authorization response", logData...) // info
 
 	if logger.V(1).Enabled() {
 		if !success {
 			baseLogData = append(baseLogData, "object", result)
+		}
+		for k, v := range loggingFields {
+			baseLogData = append(baseLogData, k, v)
 		}
 		logger.V(1).Info("outgoing authorization response", baseLogData...) // debug
 	}
--- a/pkg/service/auth.go
+++ b/pkg/service/auth.go
@@ -76,10 +76,11 @@ type AuthService struct {
 	Index                  index.Index
 	Timeout                time.Duration
 	MaxHttpRequestBodySize int64
+	EnableLoggingFields    bool
 }
 
-func NewAuthService(index index.Index, timeout time.Duration, maxHttpRequestBodySize int64) *AuthService {
-	return &AuthService{Index: index, Timeout: timeout, MaxHttpRequestBodySize: maxHttpRequestBodySize}
+func NewAuthService(index index.Index, timeout time.Duration, maxHttpRequestBodySize int64, enableLoggingFields bool) *AuthService {
+	return &AuthService{Index: index, Timeout: timeout, MaxHttpRequestBodySize: maxHttpRequestBodySize, EnableLoggingFields: enableLoggingFields}
 }
 
 // ServeHTTP invokes authorization check for a simple GET/POST HTTP authorization request
@@ -309,8 +310,10 @@ func (a *AuthService) Check(parentContext gocontext.Context, req *envoy_auth.Che
 	}
 
 	var loggingFields map[string]string
-	if p, ok := pipeline.(*AuthPipeline); ok {
-		loggingFields = p.loggingFields()
+	if a.EnableLoggingFields {
+		if p, ok := pipeline.(*AuthPipeline); ok {
+			loggingFields = p.loggingFields()
+		}
 	}
 	a.logAuthResult(result, ctx, loggingFields)
 
--- a/pkg/service/auth_pipeline.go
+++ b/pkg/service/auth_pipeline.go
@@ -556,21 +556,31 @@ func (pipeline *AuthPipeline) metricLabels() map[string]string {
 	return labels
 }
 
-func (pipeline *AuthPipeline) loggingFields() map[string]string {
+const loggingFieldPrefix = "logging."
+
+func truncateValue(s string, maxLen int) string {
+	if maxLen > 0 && len(s) > maxLen {
+		return s[:maxLen] + "...(truncated)"
+	}
+	return s
+}
+
+func (pipeline *AuthPipeline) loggingFields(maxValueBytes int) map[string]string {
 	fields := make(map[string]string)
 
 	filteredMetadata := pipeline.GetRequest().GetAttributes().GetMetadataContext().GetFilterMetadata()
 	if customFields, ok := filteredMetadata["io.kuadrant.logging.fields"]; ok {
 		for k, v := range customFields.Fields {
+			key := loggingFieldPrefix + k
 			switch kind := v.Kind.(type) {
 			case *structpb.Value_StringValue:
-				fields[k] = kind.StringValue
+				fields[key] = truncateValue(kind.StringValue, maxValueBytes)
 
 			case *structpb.Value_NumberValue:
-				fields[k] = fmt.Sprintf("%v", kind.NumberValue)
+				fields[key] = fmt.Sprintf("%v", kind.NumberValue)
 
 			case *structpb.Value_BoolValue:
-				fields[k] = fmt.Sprintf("%v", kind.BoolValue)
+				fields[key] = fmt.Sprintf("%v", kind.BoolValue)
 
 			case *structpb.Value_StructValue:
 				if celExprField, ok := kind.StructValue.Fields["cel_expr"]; ok {
@@ -585,7 +595,7 @@ func (pipeline *AuthPipeline) loggingFields() map[string]string {
 							pipeline.Logger.Error(err, "failed to evaluate CEL expression", "expression", exprStr)
 							continue
 						}
-						fields[k] = fmt.Sprintf("%v", value)
+						fields[key] = truncateValue(fmt.Sprintf("%v", value), maxValueBytes)
 					}
 				}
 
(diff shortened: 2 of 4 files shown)


pkg/service/auth.go
```diff
@@ -415,12 +419,18 @@ func (a *AuthService) logAuthResult(result auth.AuthResult, ctx gocontext.Contex
 		}
 		logData = append(logData, "object", reducedResult)
 	}
+	for k, v := range loggingFields {
+		logData = append(logData, k, v)
+	}
 	logger.Info("outgoing authorization response", logData...) // info
```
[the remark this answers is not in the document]
andresllh: This is intentional by design — the logging fields are configured by the platform operator via TelemetryPolicy (a cluster-scoped CRD), not by end users or request data. The operator explicitly chooses which attributes to surface in auth decision logs for audit purposes (e.g. `auth.identity.sub`, `request.path`). Redacting or gating these behind `V(1)` would defeat the feature's purpose.

This follows the same pattern as the existing `metricLabels()` path in Authorino, which also writes operator-configured values at info level without redaction.

guicassolato (changes requested):
I don't know how I feel about dumping user-defined values to the production logs. It could have serious implications on multi-tenant deployments. I think this needs further discussion.

I would at the very least encourage putting the feature behind an instance-level flag and/or making it opt-in in the AuthConfig (so admins could choose to disable it with an ValidatingAdmissionPolicy).

In case the conversation does not progress timely in the PR itself, please feel free to bring it to the next Kuadrant community call or offline in the `#kuadrant` Slack channel.

andresllh:
@guicassolato there's 2 different ways I can go to make this opt-in.
1.  Add a flag `--enable-logging-fields` (or something like this) with false as the default.
2. AuthConfig opt-in: `spec.logging.enabled` (or something similar)

Which would you prefer?
Option 1 is simpler and doesn't require API changes.

guicassolato:
> @guicassolato there's 2 different ways I can go to make this opt-in.
> 
> 1. Add a flag `--enable-logging-fields` (or something like this) with false as the default.
> 2. AuthConfig opt-in: `spec.logging.enabled` (or something similar)
> 
> Which would you prefer? Option 1 is simpler and doesn't require API changes.

@andresllh, I would start with an instance-level `--enable-logging-fields` flag (option 1), then reiterate with per-AuthConfig configurability (option 2) in a separate if needed.

Option 1 will require changes to the Authorino Operator too, but it's the cleanest way IMO to default to `false` from start without asking users to create an additional resource (which may still come later if we do option 2).

andresllh:
@guicassolato I've added the flag. Can you take another look?

pkg/service/auth_pipeline.go
```diff
@@ -556,6 +556,48 @@ func (pipeline *AuthPipeline) metricLabels() map[string]string {
 	return labels
 }
 
+func (pipeline *AuthPipeline) loggingFields() map[string]string {
+	fields := make(map[string]string)
+
+	filteredMetadata := pipeline.GetRequest().GetAttributes().GetMetadataContext().GetFilterMetadata()
+	if customFields, ok := filteredMetadata["io.kuadrant.logging.fields"]; ok {
```
guicassolato, line 572: We should have a way to block specific custom field names that overlap with fields of the current structured log record which already contains `authorized`, `response` and `object`. We want to make sure that a user won't be able to specify, for example, a custom field `authorised: "true"`.

I believe the current zap JSON encoder does not de-duplicate keys. Rather, it emits duplicates verbatim. If the downstream JSON parsers keep the last occurrence for each duplicate, then a (malicious) user-defined field `authorised: true` could end up deceiving a consumer of the logs into believing that a denied request was actually allowed. For a feature whose entire purpose is auditing authorisation decisions, that'd be a real integrity hole.

Maybe an easy solution could be namespacing all custom field names?

pkg/service/auth_pipeline.go
```diff
@@ -556,6 +556,48 @@ func (pipeline *AuthPipeline) metricLabels() map[string]string {
...
+						}
+						fields[k] = fmt.Sprintf("%v", value)
+					}
+				}
+
+			default:
+				pipeline.Logger.V(1).Info("unexpected value kind", "kind", kind)
+			}
+		}
+	}
```
guicassolato, line 606: I wonder if we should cap the resolved custom values to a maximum length. Especially dynamically resolved values from custom CEL expressions, such as `request.headers['x-custom-header']` and `request.body` (when enabled), could cause STDOUT backpressure and ultimately even be a DoS vector.

pkg/service/auth.go
```diff
@@ -415,12 +422,18 @@ func (a *AuthService) logAuthResult(result auth.AuthResult, ctx gocontext.Contex
 		}
 		logData = append(logData, "object", reducedResult)
 	}
+	for k, v := range loggingFields {
+		logData = append(logData, k, v)
```
guicassolato, line 427: How can we ensure this is compatible with redaction of sensitive data as controlled by the `--log-redact-add-field`, `log-redact-remove-field`, `--log-redact-add-header` and `--log-redact-remove-header` flags?

These flags were [introduced](https://github.com/Kuadrant/authorino/pull/618) to configure the mechanism that mitigates risks of exfiltrating sensitive data to debug logs. With this change, production-level logs are subject to the same risks.

How can a sysadmin control which user-defined customisations are allowed and which ones are not?

pkg/service/auth_pipeline.go
```diff
@@ -556,6 +556,48 @@ func (pipeline *AuthPipeline) metricLabels() map[string]string {
...
+				fields[k] = kind.StringValue
+
+			case *structpb.Value_NumberValue:
+				fields[k] = fmt.Sprintf("%v", kind.NumberValue)
+
+			case *structpb.Value_BoolValue:
+				fields[k] = fmt.Sprintf("%v", kind.BoolValue)
+
+			case *structpb.Value_StructValue:
+				if celExprField, ok := kind.StructValue.Fields["cel_expr"]; ok {
```
guicassolato, line 586: We're currently looking into enforcing a complexity budget for CEL expressions in an AuthConfig. Because these customisations come from request metadata they could end up dodging the limit enforcement.

Perhaps it's out of scope now, but we'll have to find a way to conciliate per-AuthConfig budgets with a complexity that can change per request.

My worry here is the added CEL cost on the hot path.

pkg/service/auth_pipeline.go
```diff
@@ -556,6 +556,48 @@ func (pipeline *AuthPipeline) metricLabels() map[string]string {
...
+						fields[k] = fmt.Sprintf("%v", value)
+					}
+				}
+
+			default:
+				pipeline.Logger.V(1).Info("unexpected value kind", "kind", kind)
+			}
+		}
+	}
+
```
guicassolato, line 607: Another sanitisation concern – In `production` mode, Authorino prints structured log messages (processed as JSON) which comes with some escaping of newlines and control chars in the values. However, in `development` mode, Authorino uses the console encoder, which doesn't escape newlines for example. Even not being the most recommended mode for automated log parsing, lack of sanitisation could still pose a risk of breaking readability for users.

pkg/service/auth_pipeline.go
```diff
@@ -556,6 +556,58 @@ func (pipeline *AuthPipeline) metricLabels() map[string]string {
 	return labels
 }
 
+const loggingFieldPrefix = "logging."
+
+func truncateValue(s string, maxLen int) string {
+	if maxLen > 0 && len(s) > maxLen {
+		return s[:maxLen] + "...(truncated)"
```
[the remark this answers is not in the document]
andresllh: for bullet 1: This is valid, but negligible. The purpose of this is to prevent multi KB values from hitting stdout, not to enforce the exact byte cap. 
for bullet 2: for it to be negative it would most likely be misconfiguration that's very unlikely in practice.

andresllh:
@guicassolato made the updates. PTAL whenever you have some time.

pkg/service/auth_pipeline.go
andresllh: Addressed in e68456be by prefixing all custom fields with `logging.` to prevent collisions with built-in log fields.

pkg/service/auth_pipeline.go
andresllh: Addressed in e68456be by adding the admin-configurable `--logging-fields-max-value-bytes` limit, which defaults to 1024 bytes.

pkg/service/auth_pipeline.go
andresllh: Addressed in c43b75dc by escaping control characters in static and CEL-resolved logging field values before truncation.

pkg/service/auth_pipeline.go
andresllh: Would you prefer restricting logging fields to simple Authorization JSON field selectors (for example, `auth.identity.sub`), which matches the TelemetryPolicy use case and avoids arbitrary CEL cost, or should we retain arbitrary CEL and integrate it with the complexity budget?

pkg/service/auth.go
andresllh: Addressed in 93e20dbf by resolving custom CEL expressions against `RedactedAuthorizationJSON`, so existing admin-configured sensitive-field and header redaction rules apply before values reach production logs.

pkg/service/auth_pipeline.go
guicassolato: Thanks for teeing up the options. My take: let's keep arbitrary CEL (your second option) rather than restricting logging fields to plain selectors. Restricting the syntax would break parity with the metrics-labels feature (same user-defined fields, which is the whole point) and wouldn't actually bound cost — a selector over deep JSON is cheap, and only pathological expressions are expensive, which a syntax restriction doesn't prevent.

That said, the concern is real and the code makes it sharper than "some CEL cost". `NewExpression` → `Compile` builds a fresh `cel.Env` (declarations + `ext.Strings` + `OptionalTypes`), then parses, type-checks, and builds the Program — per field, per request, with no caching. For metrics-labels that cost was only paid when metrics were enabled; here it lands on the always-on INFO decision path.

So I'd like to split this into two:
1. A small merge condition: cache the compiled expression keyed by its source string, in a bounded cache (LRU or capped map). The bound matters because it also prevents a memory-growth vector from many distinct expressions. Since expressions arrive as request-time metadata we can't precompile at reconcile time, but the realistic workload is the same policy emitting the same expressions every request, so caching collapses it to a single compile. This removes the certain, dominant cost.

2. A follow-up (not blocking): We can work on the details later, but I'm thinking a conservative, independent limit (optionally a max-number-of-fields-per-request cap too) that closes the "request metadata dodges the per-AuthConfig budget" gap — logging fields get their own ceiling, decoupled from AuthConfig budgets. In time, I'll open a tracking issue and link it here.

With the caching in (1), I'm comfortable merging and handling the ceiling as a fast-follow.

guicassolato (dismissed):
Tested this locally end-to-end and it works as intended. Thanks for putting all the work @andresllh! 🎉

### Verification steps

❶ Setup

```sh
make cluster install run \
  RUN_EXTRA_ARGS="--enable-logging-fields --logging-fields-max-value-bytes=256 --log-redact-add-field=email"
```

❷ AuthConfig (`plain` identity sourced from the "verified JWT" in the filter metadata, so a we have an extra piece of sensitive data to test redaction)

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: logging-fields-test
spec:
  hosts:
    - example.com
  authentication:
    "jwt-plain":
      plain:
        expression: metadata.filter_metadata["envoy.filters.http.jwt_authn"].verified_jwt
EOF
```

❸ Send a request

```sh
grpcurl -plaintext -d @ localhost:50051 envoy.service.auth.v3.Authorization.Check <<'EOF'
{
  "attributes": {
    "metadata_context": {
      "filter_metadata": {
        "envoy.filters.http.jwt_authn": {
          "verified_jwt": {
            "iss": "https://trusted-idp", "aud": "test",
            "sub": "1c06e06f-3870-48e3-b879-9a627f738ea9",
            "exp": 123456, "email": "user@example.com"
          }
        },
        "io.kuadrant.logging.fields": {
          "client_identity":          { "cel_expr": "auth.identity.sub" },
          "request_method":           { "cel_expr": "request.method" },
          "sensitive_header":         { "cel_expr": "request.headers.authorization" },
          "sensitive_identity_field": { "cel_expr": "auth.identity.email" },
          "static_field": "audit-v1", "numeric_field": 42, "bool_field": true,
          "unresolvable":             { "cel_expr": "auth.nonexistent.value" },
          "null_field": null
        }
      }
    },
    "request": {
      "http": {
        "host": "example.com", "path": "/some-path", "method": "POST",
        "headers": { "authorization": "Bearer <jwt>" }
      }
    }
  }
}
EOF
```

❹ Verify in the Authorino logs – you should spot an "ongoing authorization response" with the following values

```
"authorized": true, "response": "OK",
"logging.client_identity": "1c06e06f-3870-48e3-b879-9a627f738ea9",
"logging.request_method": "POST",
"logging.static_field": "audit-v1",
"logging.numeric_field": "42",
"logging.bool_field": "true",
"logging.sensitive_header": "***REDACTED***",
"logging.sensitive_identity_field": "***REDACTED***"
```

I also tested truncation; omitting the steps here for brevity.

### Verdict

Flipping my previous review to APPROVED with one condition — please cache the compiled CEL expressions (bounded) before merge, as discussed in https://github.com/Kuadrant/authorino/pull/671/changes#r3905926262. Everything else — namespacing, value truncation, redaction integration, control-char sanitisation — looks great 👍 I'm also re-running the tests which we haven't done in a while. I don't expect anything to break but just in case.

pkg/service/auth_pipeline.go
andresllh: Addressed the merge condition in 3012db5c with a bounded 256-entry LRU cache for compiled expressions and compilation errors; the independent runtime and field-count limits are tracked in CONNLINK-1721.

pkg/service/logging_fields_expression_cache.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"container/list"
+	"sync"
+
+	"github.com/kuadrant/authorino/pkg/expressions/cel"
+)
+
+const loggingFieldExpressionCacheCapacity = 256
+
+var loggingFieldExpressions = newLoggingFieldExpressionCache(loggingFieldExpressionCacheCapacity)
```
guicassolato: Hmm. Logging field expression cache is _one_ use case. I wonder if the `ExpressionCache` struct and functionality don't belong to the `github.com/kuadrant/authorino/pkg/expressions/cel` package (renaming off `loggingField`). Then in `pkg/service` we instantiate the `loggingField` one. WDYT?

andresllh: Addressed in 8a2a8896 by moving the reusable bounded `ExpressionCache` into `pkg/expressions/cel` and keeping only the logging-specific instance in `pkg/service`.

pkg/service/auth_pipeline.go
guicassolato: Glad to see the namespacing prefix changed to `.custom`. Other options could be `x-` (borrowing from HTTP headers) and `ext.`, but I'm fine with `custom.`

guicassolato:
cc @coderabbit

pkg/service/auth_pipeline.go
guicassolato: Thanks!

guicassolato (approved):
Good to merge! Thanks so much for the hard work, @andresllh!

PR #675, opened 2026-08-25
fix: make worker.Stop() idempotent so a failed reconcile cannot crash Authorino

### LINKED ISSUE #674: Reconcile crashes the whole Authorino process: `worker.Stop()` is not idempotent and `cleanConfigs()` cleans the same evaluator twice when `translateAuthConfig()` fails
**Describe the bug**

`(*worker).Stop()` isn't idempotent, and `Reconcile()` calls it twice on the same evaluator whenever `translateAuthConfig()` fails for an already-indexed AuthConfig. The second call closes a closed channel and panics in a goroutine `AuthConfig.Clean()` spawned, taking the process with it.

An update is meant to be *clean the old, then replace it*. The error path does the first half only:

```go
// controllers/auth_config_controller.go:190
if err := r.cleanConfigs(ctx, resourceId); err != nil { ... }   // stops the indexed config's workers

translatedAuthConfig, err := r.translateAuthConfig(...)
if err != nil {
    r.StatusReport.Set(resourceId, api.StatusReasonInvalidResource, err.Error(), []string{})
    return ctrl.Result{}, err                                    // cleaned, never replaced
}
```

1. **reconcile #1** — `cleanConfigs()` stops the indexed config's refreshers, `translateAuthConfig()` fails, we return the error *before* `addToIndex()`. The index still holds the instance we just tore down.
2. **the requeue**, ~5ms later.
3. **reconcile #2** — `cleanConfigs()` (`:254`) pulls that same instance back out and cleans it again → `close of closed channel`.

Same object, not a copy: `Index.Set()` stores the struct by value, but `IdentityConfigs` is a slice, so the copy shares its backing array down to the `*worker`.

```go
// pkg/workers/worker.go:74
func (w *worker) Stop() error {
    if w.done != nil {
        close(w.done)   // second call panics
    }
    return nil
}
```

The nil check only covers a worker never started — nothing marks a *stopped* one stopped. `Start()` calls `Stop()` for restarts, likely why it survived there, but `oidcProviderVerifier.Clean()` (jwt.go:158) and `OPAExternalSource.cleanupRefresher()` (opa.go:133) call it directly, and `cleanConfigs()` can't know the instance was already cleaned.

Fatal rather than logged because `AuthConfig.Clean()` fans out into raw goroutines (`config.go:56`) and `recover()` only catches panics in its own goroutine. (Moot anyway — on controller-runtime v0.16.3 `RecoverPanic` is a `*bool` we never set, so the controller re-panics at `internal/controller/controller.go:116`.) Same shape as #651/#652 and #494.

Needs two ordinary things: a refresher (`jwt.ttl > 0` or `opa.externalPolicy.ttl`), and a translate failure *after* the config was indexed — rotated or deleted `credentialsRef` Secret, unreachable OPA registry, Rego precompile error, bad selector, CEL build failure. A cold start is safe, since nothing is indexed yet to clean; it needs the config live and serving when the dependency breaks. Which is exactly the incident you hoped to ride out on cached policy and JWKS.

**Help us Reproduce it**

No cluster — the existing fake-client harness in `controllers/` drives the real `Reconcile()`. AuthConfig with `jwt.ttl: 60` against the mock OIDC server `TestMain` already runs → reconcile (indexes it) → add an `oauth2Introspection` whose `credentialsRef` names a missing Secret → reconcile twice.

```
reconcile #1 ok - authconfig indexed, OIDC refresher running
reconcile #2 failed as expected: secrets "no-such-secret" not found
stale config still indexed with its worker already stopped
panic: close of closed channel

goroutine 82 [running]:
workers.(*worker).Stop                  pkg/workers/worker.go:76
identity.(*oidcProviderVerifier).Clean  pkg/evaluators/identity/jwt.go:158
identity.(*JWTAuthentication).Clean     pkg/evaluators/identity/jwt.go:77
evaluators.(*IdentityConfig).Clean      pkg/evaluators/identity.go:157
evaluators.(*AuthConfig).Clean.func1    pkg/evaluators/config.go:59
created by evaluators.(*AuthConfig).Clean in goroutine 64
```

`created by` — child goroutine, no recover boundary. In-cluster is the same three steps: an AuthConfig with `jwt.ttl` serving traffic, break a translate-time dependency (scale the policy registry to 0, delete the introspection Secret), touch the AuthConfig; every pod dies on the requeue.

**Two related problems on the same path**

*Leaked refreshers.* A partway `translateAuthConfig()` failure drops whatever it already built — nothing cleans a config that never reached the index. If the JWT identity is built before the failing one (map order), its worker is already running and unowned. Measured on the same loop with no worker on the indexed side, so it loops instead of panicking: `goroutines before=3 after=51 failed reconciles (delta=+48 over 50)`. Backoff starts at ~5ms, so dozens within seconds, each polling discovery every `ttl` forever. Fixing `Stop()` doesn't touch it.

*Stale index keys → the wrong config gets cleaned.* `Index.Set()` appends to `c.keys[id]` unconditionally (`index.go:77`) and neither `Delete()` nor `deleteKey()` prunes it. Narrow `A` from `[x.com, y.com]` to `[y.com]`: the tree entry goes, `c.keys["ns/A"]` keeps `x.com`. Create `B` claiming `x.com` and it legitimately gets it. Now every reconcile of `A` resolves `FindKeys("ns/A")[0]` → `Index.Get("x.com")` → **`B`'s config** → `B.Clean()`, stopping `B`'s refreshers and shutting its caches while `B` serves traffic; once more and it's the double `Stop()` on `B`'s worker. That stale map also keeps `Index.Empty()` false forever, defeating the re-bootstrap guard in `bootstrapIndex()`.

**Expected behavior**

A failed translate should keep the last-known-good config indexed, report `InvalidResource`, and retry. Double-cleaning should be a no-op, no cleanup path should kill the process, and reconciling one AuthConfig should never touch another's evaluators.

The blast radius isn't one pod: every replica reconciles every matching AuthConfig, so they panic together and the whole ext_authz service CrashLoopBackOffs. With `failure_mode_allow: true` — a common choice precisely to survive authz blips — that window is every request allowed through unauthenticated; false gives a 503 on everything protected.

The aftermath hides it. `bootstrapIndex()` re-indexes the host with the `denyAll` placeholder (503 `"Busy"`); reconcile still fails so it's never replaced, but the placeholder carries no workers, so **it doesn't panic again**. One crash, a pod that looks recovered, a host stuck on 503, nothing in the panic naming the AuthConfig, and a status reading `InvalidResource` like a benign user error.

- Cluster information: none required — both reproduce under `go test ./controllers/...`. Reconcile-lifecycle bug, platform-independent.
- Authorino version: `main` @ `58fecc6c`; present in **v0.27.0**. Go 1.26.4.
- Link to the Authorino `AuthConfig` resource: inline above — any AuthConfig with `jwt.ttl` or `opa.externalPolicy.ttl`.

**Additional context**

Not a regression: `pkg/workers/worker.go` has one commit since it was written (`714717a7`, the `pkg/cron` → `pkg/workers` rename) and `Stop()` none; `index.go` likewise only `e64de14a`. #572 is close history — it flagged races in `TestOPAExternalUrlWithTTL` and `TestOIDCProviderVerifierRefresh`, these two exact refreshers, and closed on the races.

Proposed, small and independent:

1. `workers/worker.go:74` — nil `done` after closing (guarded; `Start()` also calls `Stop()`). Kills the crash alone.
2. `evaluators/config.go:56` — `recover()` in the per-evaluator goroutine.
3. `index/index.go` — dedupe in `Set()`, prune `c.keys` in `deleteKey()`/`Delete()`.
4. `auth_config_controller.go:196` — clean the partially-built config on the error path.

~30 lines plus regression tests. Happy to split (3) into its own issue — kept here because it's the second trigger for the same panic.

**### I have a fix + regression tests ready and am happy to open a PR.**

Aman-Cool:
#674 has the long version. The short story:

`cleanConfigs()` cleans up whatever AuthConfig is sitting in the index right now. When `translateAuthConfig()` fails we return before `addToIndex()` — so the instance we just cleaned up is still the one in the index. The requeue comes back ~5ms later and cleans that exact same instance again, `worker.Stop()` closes an already-closed channel, and because it happens in a goroutine `AuthConfig.Clean()` spawned there's nothing to recover it. Pod gone, every replica at once.

Four small changes:

- `Stop()` is idempotent now (and takes the mutex, since `Start()` calls it too)
- each cleanup goroutine recovers, so no cleaner can ever take the process with it again
- the index prunes its `id -> keys` map, so `FindKeys()` stops handing back hosts a resource has released — otherwise `cleanConfigs()` happily resolves to whichever AuthConfig owns that host *now* and cleans someone else's evaluators
- a failed translation cleans up what it already built, instead of leaking a refresher worker on every requeue (it was ~1 per attempt, each still polling the discovery endpoint forever)

Every test fails on the unfixed code — revert any one of the four and run it if you want to watch it go.

Fixes #674

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented background refresh workers from leaking when authentication configuration processing fails.
  * Preserved the last valid configuration during failed replacement attempts.
  * Improved index tracking so deleted or reassigned entries are no longer reported incorrectly.
  * Prevented cleanup errors from crashing the process.
  * Ensured startup and cleanup failures, including unexpected evaluator errors, are reported safely.

* **Reliability**
  * Made worker shutdown safe to repeat and enabled workers to restart after stopping.
  * Improved background refresh start-up and lifecycle handling during configuration changes.
  * Prevented duplicate background refresh workers from starting concurrently.


controllers/auth_config_controller.go
```diff
@@ -281,6 +281,31 @@ func (r *AuthConfigReconciler) translateAuthConfig(ctx context.Context, authConf
...
+	interfacedMetadataConfigs := make([]auth.AuthConfigEvaluator, 0)
+	interfacedAuthorizationConfigs := make([]auth.AuthConfigEvaluator, 0)
+	interfacedResponseConfigs := make([]auth.AuthConfigEvaluator, 0)
+	interfacedCallbackConfigs := make([]auth.AuthConfigEvaluator, 0)
+
+	// some evaluators start background workers as soon as they are built. if the translation
+	// fails halfway through, whatever was built so far is discarded and never reaches the index,
+	// so nothing would ever clean it up - and a persistently failing reconcile is requeued over
+	// and over, leaking a worker each time
+	defer func() {
```
guicassolato: Interesting. When I first read #674 – literally a few minutes ago –, I thought only a check on whether the channel had been closed before would suffice. I now realise this case of workers that started prematurely before a mid-process reconciliation error also matters.

I wonder if cleaning the workers when a reconciliation error is detected isn't treating the symptom rather than the (well spotted) design gap. Perhaps a two-phase approach — build everything first, start workers only after translation succeeds – would be cleaner.
1. It eliminates the problem class entirely. The defer cleanup in `translateAuthConfig` is reactive: it creates orphaned workers, then tears them down. With deferred start, orphaned workers never exist in the first place. No cleanup path means no cleanup bugs.
2. It makes the "last known good" story more honest. Right now, `cleanConfigs()` stops the old config's workers before we know the new config is valid. If translation then fails, the old config stays indexed but its refresher is dead — so the "last known good config should stay in the index" assertion in the test is true structurally, but that config is degraded (stale JWKS, no refresh). Ideally, the old config would keep running until the new one is fully ready, a proper swap-on-success.
3. It poses a simpler invariant. "Workers only run for configs that are in the index" is easier to reason about than "workers run for partially-built configs too, but we clean those up on error."

I understand this would require some refactoring of `oidcProviderVerifier` and `OPAExternalSource` (to return signals the reconciler can call on to start the workers), almost extending the scope of this PR whose number 1 issue IIUC was the non-idempotent `Stop()`.

Interested in your thought, @Aman-Cool.

Aman-Cool: I went looking for a hole in this and mostly came up empty.., you're right, and the reorder is the better fix.

Two things did fall out of the search though. First, **a small correction**: the stale JWKS bit isn't quite right. The `ttl` refresher only re-fetches the discovery document; JWKs get pulled again by go-oidc whenever a token's `kid` misses the cache, worker or not (rotation strategy straight from the spec, and our own godoc on `ttl` says as much). So a dead refresher doesn't break key rotation. What actually goes stale is discovery metadata and the external OPA policy. Still degraded, just a slower burn than it sounds.

The second is more a hole in my patch than in your argument. Chasing "no cleanup path means no cleanup bugs", I went looking for a case where translation succeeds and the config still never lands in the index; and there is one: all hosts taken by another AuthConfig. `addToIndex()` hands back `linkedHosts=[]`, `looseHosts=[...]` and a nil error, we return, no requeue, and the workers are already running. +20 goroutines over 20 collision reconciles, same on main as on this branch. My `defer` never fires because translation succeeded, and a deferred start wouldn't catch it either unless it hangs off `linkedHosts` rather than "translate returned nil".

So the invariant is the right one, it just wants anchoring a step later. Which makes the two-phase version worth more than my patch, not less. Offer stands.., keep this PR to the crash and do the refactor properly next, or extend it here, your call.

guicassolato: I'd say do the work in this PR. We've been leaving with a non ideal `worker.Stop()` for a while now. We can wait a little longer to land the best fix we can deliver with all the information we have now.

Thanks cracking this one @Aman-Cool!

Aman-Cool: Done..,two-phase it is.

Refreshers don't start in the constructors any more; there's an `auth.AuthConfigStarter` mirroring the cleaner, called once the config is translated and indexed. Translation now runs before the clean, so a failed one leaves the old config indexed *and* still refreshing; your (2) properly rather than structurally. Start is gated on `linkedHosts`, and the `defer` cleanup is gone.

Heads up that `TestOIDCProviderVerifierRefresh` and `TestOPAExternalUrlWithTTL` both moved.., they asserted the refresher existed right after the constructor, now they assert it doesn't until `Start()`.

Ready for another look.

controllers/auth_config_controller_test.go
```diff
@@ -446,3 +448,177 @@ func BenchmarkReconcileAuthConfig(b *testing.B) {
...
+	assert.Check(t, refresherRunning(indexed), "the indexed config should be refreshing once it is reconciled")
+
+	breakTranslation(t, k8sClient, name)
+
+	_, err = reconciler.Reconcile(context.Background(), req)
+	assert.ErrorContains(t, err, "no-such-secret")
+
+	stillIndexed := authConfigIndex.Get("echo-api")
+	assert.Check(t, stillIndexed != nil, "the last known good config should stay in the index")
+	assert.Check(t, refresherRunning(stillIndexed), "and it should still be refreshing, not left degraded")
```
guicassolato: This test verifies the worker is still running after a failed attempt to reconcile, which is good. However, it builds on the known fact that the failed reconciliation didn't result in the AuthConfig being re-indexed. We should assert `stillIndexed` is the same AuthConfig previously stored as `indexed` and not the newer failed version of the resource. One easy way to do that is, somewhere along with `breakTranslation`, removing from the AuthConfig the identity evaluator. By only having a broken metadata rule, there's no other option but the leftover running worker being from the previous, last-known-good config.

Aman-Cool: Good catch, and you're right😅 the test was leaning on "a failed translation doesn't re-index" without ever checking it. Passing for a reason it doesn't assert.

Tried your suggestion: dropping the identity from the AuthConfig along with `breakTranslation` works nicely. The broken version has no jwt at all, so a running refresher can only have come from the previous config; nothing left to coincidence. Still fails if I put the clean back before the translate, so it keeps the coverage it had.

Will push that shortly.

pkg/evaluators/config.go
```diff
@@ -40,13 +41,35 @@ func (config *AuthConfig) GetChallengeHeaders() []map[string]string {
...
-func (config *AuthConfig) Clean(ctx context.Context) error {
+// Start kicks off whatever the evaluators of this config should only be running once the config is
+// valid and reachable. Unlike Clean, it runs sequentially: there is nothing to parallelise here and
+// it keeps the panic surface of the reconcile path small.
+func (config *AuthConfig) Start(ctx context.Context) error {
+	var errs error
+
+	for _, evaluator := range config.allEvaluators() {
+		if starter, ok := evaluator.(auth.AuthConfigStarter); ok {
+			if err := starter.Start(ctx); err != nil {
```
guicassolato: An individual starter could still panic, right? While `Clean` dispatches wrapped in `recover()`, `Start()` does not, propagating through `startConfigs` and ultimately yielding a reconciliation error while there's already no rolling back of started evaluators before that.

Wondering if there's a way to surface the startup error without leaving garbage that a cleanup (or no cleanup at all) would fail to clean. WDYT?

pkg/evaluators/identity/jwt.go
```diff
@@ -93,24 +105,42 @@ type JWTVerifier interface {
...
 	v.getOpenIdProvider(ctxWithLogger, false)
-	v.setupOpenIdProviderRefresh(ctxWithLogger, ttl)
 	return v
 }
 
+// impl: auth.AuthConfigStarter
+func (v *oidcProviderVerifier) Start(ctx gocontext.Context) error {
+	v.mu.Lock()
+	alreadyRunning := v.refresher != nil
+	v.mu.Unlock()
```
guicassolato: Using the RW mutex to read the refresher here. Should it be `RLock` and `RUnlock`?

Also, it's a preexisting issue, but should `v.setupOpenIdProviderRefresh` also use the lock before setting the refresher?

pkg/evaluators/authorization/opa.go
```diff
@@ -69,34 +69,54 @@ func NewOPAAuthorization(policyName string, rego string, externalSource *OPAExte
...
+// translation of the authconfig; only the periodic refresh waits until the config is indexed.
+// impl: auth.AuthConfigStarter
+func (opa *OPA) Start(ctx context.Context) error {
+	if !opa.pullFromRegistry || opa.ExternalSource == nil {
+		return nil
+	}
+
+	opa.mu.Lock()
+	alreadyRunning := opa.ExternalSource.refresher != nil
+	opa.mu.Unlock()
```
guicassolato: Same as https://github.com/Kuadrant/authorino/pull/675#discussion_r3914125562

pkg/evaluators/config.go
```diff
@@ -55,6 +78,19 @@ func (config *AuthConfig) Clean(ctx context.Context) error {
...
+					logger := log.FromContext(ctx)
+					if named, ok := e.(auth.NamedEvaluator); ok {
+						logger = logger.WithValues("config", named.GetName())
+					}
+					logger.Error(fmt.Errorf("%v", r), "recovered from panic while cleaning up evaluator")
+				}
+			}()
 			if cleaner, ok := e.(auth.AuthConfigCleaner); ok {
 				if err := cleaner.Clean(ctx); err != nil {
 					errors = multierror.Append(errors, err)
```
guicassolato: Another preexisting bug – `errors` comes from the closure, yet handled without a mutex. Two simultaneous non-nil `Clean` errors race the read-modify-write and can drop an error.

pkg/evaluators/config.go
Aman-Cool: Yes, and it's the same class this PR is about, so it should be covered. One difference from `Clean`: `Start()` runs sequentially in the reconcile goroutine, so a single `recover()` in `AuthConfig.Start` covers every starter, rather than one per goroutine.

Small correction on the propagation — `startConfigs` errors are logged today, not returned, so a failed start doesn't fail the reconcile. A panic would though, and controller-runtime re-panics by default (`RecoverPanic` is a `*bool` we never set), so that's the actual hole.

On the garbage, I think the ordering already covers it: `startConfigs` only runs once the config is indexed, so a partial start leaves it degraded but never unreachable. The next `cleanConfigs()` finds it through the index and stops whatever did start, and `Clean` on an evaluator that never started is a no-op. That was rather the point of moving the start after `addToIndex`.

Which leaves surfacing the error. I'd keep it off the reconcile return — a requeue would re-translate, re-clean and re-start the lot, and today `StartWorker` only errors when ttl <= 0, i.e. no refresher configured. Happy to put it on the status report instead if you'd rather it were visible.

pkg/evaluators/identity/jwt.go
Aman-Cool: Right on both, though I think the first one goes the other way: `RLock` would make it worse. The check and the set need to be atomic and currently aren't; the lock is released before `setupOpenIdProviderRefresh` writes, so two Starts could both see nil and both spawn a worker, and one of them leaks. Holding the write lock across both is the fix, which answers your second question too: the refresher gets set under that same lock.

And yes, the unguarded write is pre-existing. It isn't actually racing today, since Start and Clean both run from the reconcile goroutine and controller-runtime won't reconcile one key concurrently, but `Clean` reads and clears the refresher under the lock now, so leaving the write outside makes the contract inconsistent. Will tidy it.

pkg/evaluators/authorization/opa.go
Aman-Cool: Same fix, with one OPA-specific wrinkle worth flagging: `refresher` lives on `OPAExternalSource`, but I guarded it with `opa.mu`; which `Call()` read-locks on every request. Taking the write lock to start a refresher would briefly block policy evaluation for that config, and it's a lock guarding a field it doesn't own.

So rather than widening `opa.mu`, I'd give `OPAExternalSource` its own mutex for its refresher: same atomic check-and-set as the jwt one, and the request path stays out of it entirely. Shout if you'd rather keep it on the single lock.

pkg/evaluators/config.go
Aman-Cool: Yes.., read-modify-write from every cleanup goroutine, so it can drop errors, and `-race` would flag it the moment two cleaners failed at once.

Full disclosure: I spotted this when I added the `recover()` and left it alone as unrelated to the crash. Since we're in here anyway, I'll fix it.

Rather than a mutex I'd give each goroutine its own slot; `errs := make([]error, len(evaluators))`, each writes `errs[i]`, merged after `Wait()`. Distinct elements, so there's no shared write left to guard at all.

pkg/index/index.go
```diff
@@ -133,6 +135,20 @@ func (c *authConfigTree) deleteKey(id, key string) {
 	if node, _ := c.root.longestCommonLabel(revertKey(key)); node != nil && node.entry != nil && node.entry.Id == id {
 		node.entry = nil
 	}
+
+	// the key must go from the id -> keys map as well, otherwise FindKeys() keeps reporting hosts
+	// the resource no longer owns, and callers such as the reconciler's cleanConfigs() end up
+	// resolving to whichever AuthConfig owns that host now
+	if keys, ok := c.keys[id]; ok {
+		if i := slices.Index(keys, key); i >= 0 {
+			keys = slices.Delete(keys, i, i+1)
```
jasonmadigan, line 144: looks like this modifies a slice that FindKeys exposes to concurrent readers. maybe: return a copy under the read lock, and add some regression test coverage?

jasonmadigan:
this'll need a rebase

Aman-Cool:
@guicassolato @jasonmadigan whenever either of you has a moment😁

PR #678, opened 2026-08-27
docs: fixing privilege escalation bug

DaliborD45:
### Summary

Adds a new user guide documenting how to prevent a namespace-to-cluster privilege escalation vector in Authorino `AuthConfig` resources.

Two fields — `spec.authentication.*.apiKey.allNamespaces` and `spec.authentication.*.x509.allNamespaces` trigger cluster-wide `Secret` lookups when set to `true`. This means any subject able to create an `AuthConfig` in a single namespace can reach API-key / trusted-certificate secrets across **every** namespace in the cluster. 

### Changes

- **`docs/user-guides/preventing-privilege-escalation.md`** (new) — Step-by-step guide covering:
  1. Creating `ClusterRole`s that expose custom verbs (`set-apikey-all-namespaces`, `set-x509-all-namespaces`) on `authconfigs`.
  2. Granting those permissions — starting with the Authorino instance's own ServiceAccount (so it isn't locked out), then to specific SAs/Users via `RoleBinding`s.
  3. Applying a `ValidatingAdmissionPolicy` + binding that rejects `allNamespaces: true` unless the requesting subject holds the matching permission (checked via `authorizer.requestResource.check(...)`).
  4. Verifying behavior: unauthorized subjects are blocked, permitted subjects are allowed, unrestricted resources always pass, updates are re-checked (not just creates), and `RoleBinding`-granted permissions stay namespace-scoped.

## Summary by CodeRabbit

* **Documentation**
  * Added a guide explaining how to prevent namespace-to-cluster privilege escalation in Authorino `AuthConfig` resources.
  * Documented how to restrict cluster-wide API key and certificate secret lookups using a ValidatingAdmissionPolicy.
  * Included setup instructions, permission configuration, and verification scenarios for allowed and denied access.
  * Clarified that the guidance applies to cluster-wide Authorino instances and includes checks for resource updates.

### DIFF
--- a/docs/user-guides/preventing-privilege-escalation.md
+++ b/docs/user-guides/preventing-privilege-escalation.md
@@ -3,7 +3,7 @@
 Two fields on `AuthConfig` resources reach beyond the namespace they live in: `spec.authentication.*.apiKey.allNamespaces` and `spec.authentication.*.x509.allNamespaces`. When either is set to `true`, cluster-wide Authorino instances will look up the API-key / trusted-certificate `Secret`s across **every** namespace in the cluster, so anyone allowed to create `AuthConfig`s in a single namespace can use Authorino's elevated privileges to quietly reach secrets at cluster scope.
 This issue does not affect namespaced Authorino instances, but it can be a problem in multi-tenant, shared Authorino instances (aka: cluster-wide deployments).
 
-The [ValidatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) below closes that gap. It blocks those fields unless the user has been given a special permission for them, and you hand that permission only to the subjects that need it to do their job.
+The [ValidatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) below closes that gap. It blocks *enabling* those fields unless the user has been given a special permission for them, and you hand that permission only to the subjects that need it to do their job.
 
 The policy:
 
@@ -117,12 +117,20 @@ spec:
       expression: "has(object.spec.authentication) && object.spec.authentication.exists(k, has(object.spec.authentication[k].apiKey) && has(object.spec.authentication[k].apiKey.allNamespaces) && object.spec.authentication[k].apiKey.allNamespaces)"
     - name: wantsX509AllNamespaces
       expression: "has(object.spec.authentication) && object.spec.authentication.exists(k, has(object.spec.authentication[k].x509) && has(object.spec.authentication[k].x509.allNamespaces) && object.spec.authentication[k].x509.allNamespaces)"
+    # hadApiKeyAllNamespaces / hadX509AllNamespaces capture whether the field was already
+    # enabled before this request (oldObject is null on CREATE). They let the policy
+    # restrict only newly enabling allNamespaces, so a subject without the permission can
+    # still edit unrelated fields of an object that already has it enabled.
+    - name: hadApiKeyAllNamespaces
+      expression: "oldObject != null && has(oldObject.spec.authentication) && oldObject.spec.authentication.exists(k, has(oldObject.spec.authentication[k].apiKey) && has(oldObject.spec.authentication[k].apiKey.allNamespaces) && oldObject.spec.authentication[k].apiKey.allNamespaces)"
+    - name: hadX509AllNamespaces
+      expression: "oldObject != null && has(oldObject.spec.authentication) && oldObject.spec.authentication.exists(k, has(oldObject.spec.authentication[k].x509) && has(oldObject.spec.authentication[k].x509.allNamespaces) && oldObject.spec.authentication[k].x509.allNamespaces)"
   validations:
-    - expression: "!variables.wantsApiKeyAllNamespaces || variables.isExemptApiKey"
-      message: "apiKey allNamespaces: true (cluster-wide secret lookup) can only be set by a subject granted the 'set-apikey-all-namespaces' permission on authconfigs"
+    - expression: "!variables.wantsApiKeyAllNamespaces || variables.hadApiKeyAllNamespaces || variables.isExemptApiKey"
+      message: "apiKey allNamespaces: true (cluster-wide secret lookup) can only be enabled by a subject granted the 'set-apikey-all-namespaces' permission on authconfigs"
       reason: Forbidden
-    - expression: "!variables.wantsX509AllNamespaces || variables.isExemptX509"
-      message: "x509 allNamespaces: true (cluster-wide secret lookup) can only be set by a subject granted the 'set-x509-all-namespaces' permission on authconfigs"
+    - expression: "!variables.wantsX509AllNamespaces || variables.hadX509AllNamespaces || variables.isExemptX509"
+      message: "x509 allNamespaces: true (cluster-wide secret lookup) can only be enabled by a subject granted the 'set-x509-all-namespaces' permission on authconfigs"
       reason: Forbidden
 ---
 apiVersion: admissionregistration.k8s.io/v1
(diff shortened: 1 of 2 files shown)


docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,391 @@
+# User guide: Preventing namespace to cluster privilege escalation (AuthConfigs)
+
+Two fields on `AuthConfig` resources reach beyond the namespace they live in: `spec.authentication.*.apiKey.allNamespaces` and `spec.authentication.*.x509.allNamespaces`. When either is set to `true`, Authorino looks up the API-key / trusted-certificate `Secret`s across **every** namespace in the cluster, so anyone allowed to create `AuthConfig`s in a single namespace can quietly reach secrets cluster-wide.
```
guicassolato:
```suggestion
Two fields on `AuthConfig` resources reach beyond the namespace they live in: `spec.authentication.*.apiKey.allNamespaces` and `spec.authentication.*.x509.allNamespaces`. When either is set to `true`, cluster-wide Authorino instances will look up the API-key / trusted-certificate `Secret`s across **every** namespace in the cluster, so anyone allowed to create `AuthConfig`s in a single namespace can use Authorino's elevated privileges to quietly reach secrets at cluster scope.

This issue does not affect namespaced Authorino instances, but it can be a problem in multi-tenant, shared Authorino instances (aka: cluster-wide deployments).
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,391 @@
...
+
+## 4. Verifying the VAP
+
+### A normal user is blocked
+
+Try to create resources that break the rules. Run these as a regular user (one *without* the permissions) and both should be **rejected**:
+
+> [!NOTE]
+> Don't run these as a cluster admin. Anything with wildcard access (`verbs: ["*"]`) — which cluster admins have — satisfies the `set-apikey-all-namespaces` / `set-x509-all-namespaces` checks and is treated as exempt, so the request would be **allowed** and a real cluster-wide secret lookup enabled. Use an ordinary user (or `--as=<unauthorized-subject>`) to see the policy block.
```
guicassolato:
```suggestion
> Do not run these as a cluster administrator. Anything with wildcard access (`verbs: ["*"]`) — which cluster admins have — satisfies the `set-apikey-all-namespaces` / `set-x509-all-namespaces` checks and is treated as exempt, so the request would be **allowed** and a real cluster-wide secret lookup enabled. Use an ordinary user (or `--as=<unauthorized-subject>`) to see the policy block.
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,391 @@
...
+  kind: ClusterRole
+  name: authorino-set-x509-all-namespaces
+subjects:
+  - kind: ServiceAccount
+    name: <sa-name>
+    namespace: <namespace-of-sa>
+EOF
+```
+
+## 3. Create the ValidatingAdmissionPolicy
```
guicassolato:
```suggestion
## 3. Create the ValidatingAdmissionPolicy (VAP)
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,391 @@
+# User guide: Preventing namespace to cluster privilege escalation (AuthConfigs)
```
guicassolato:
```suggestion
# User guide: Preventing namespace-to-cluster privilege escalation (AuthConfigs)
```

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,391 @@
...
+
+You should get an error like this instead of the resource being created:
+
+```text
+... is forbidden: ValidatingAdmissionPolicy 'authconfig-restrict-all-namespaces' ... denied request: apiKey allNamespaces: true (cluster-wide secret lookup) can only be set by a subject granted the 'set-apikey-all-namespaces' permission on authconfigs
+```
+
+### A permitted subject is allowed
+
+Now run the same requests as a subject that holds the matching permission (granted in steps 1–2). Both should be **admitted**. Replace `<authorized-subject>` with the subject you granted the permission to (e.g. `system:serviceaccount:<namespace>:<sa>`):
```
guicassolato:
```suggestion
Now run the same requests as a subject that holds the matching permission (granted in steps 1–2). Both should be **admitted**. Replace `<authorized-subject>` with the subject you granted the permission to (for example, `system:serviceaccount:<namespace>:<sa>`):
```

guicassolato (commented):
Not from me, but a few more nits from [`/kdt:verify-docs`](https://github.com/Kuadrant/dev-team-plugin#verify-docs):

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -0,0 +1,392 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: authorino-set-x509-all-namespaces
+subjects:
+  - kind: ServiceAccount
+    name: <authorino-sa>
+    namespace: <authorino-namespace>
+EOF
+```
+
```
guicassolato: Hmmm. I've just tried the procedure again and I might be wrong about this. It turns out Authorino does NOT need the permission while it only touches `authconfigs/status` resources (not `authconfigs`).

I think we're safe to remove this part.

DaliborD45: Yes, correct observation, removed it

guicassolato (dismissed):

docs/user-guides/preventing-privilege-escalation.md
```diff
@@ -117,12 +117,20 @@ spec:
...
+    # restrict only newly enabling allNamespaces, so a subject without the permission can
+    # still edit unrelated fields of an object that already has it enabled.
+    - name: hadApiKeyAllNamespaces
+      expression: "oldObject != null && has(oldObject.spec.authentication) && oldObject.spec.authentication.exists(k, has(oldObject.spec.authentication[k].apiKey) && has(oldObject.spec.authentication[k].apiKey.allNamespaces) && oldObject.spec.authentication[k].apiKey.allNamespaces)"
+    - name: hadX509AllNamespaces
+      expression: "oldObject != null && has(oldObject.spec.authentication) && oldObject.spec.authentication.exists(k, has(oldObject.spec.authentication[k].x509) && has(oldObject.spec.authentication[k].x509.allNamespaces) && oldObject.spec.authentication[k].x509.allNamespaces)"
   validations:
-    - expression: "!variables.wantsApiKeyAllNamespaces || variables.isExemptApiKey"
-      message: "apiKey allNamespaces: true (cluster-wide secret lookup) can only be set by a subject granted the 'set-apikey-all-namespaces' permission on authconfigs"
+    - expression: "!variables.wantsApiKeyAllNamespaces || variables.hadApiKeyAllNamespaces || variables.isExemptApiKey"
```
guicassolato, line 129: I think this change adds back the grandfathered `allNamespaces` issue mentioned in this [other review](https://github.com/Kuadrant/authorino-operator/pull/351#discussion_r3842936966).

I understand the goal is to make it possible for unauthorised users to modify resources that contain a privileged value (set by an authorised user in the past). However, in practice it has implications for the revocation of privileges. A user who's once given permission to set a restricted value and then has the permission revoked can continue to modify the resource perpetuating the restricted value indefinitely until the value explicitly reset to a non-restricted one. This may confuse admins that think that, by revoking the RBAC permissions from a user, the user will be forced to reset the resources to non-privileged values when actually that is not the case.

I suppose the trade-off we're facing it:

1. **VAP does not check old state** – Unauthorised users cannot at all touch resources that contain any restricted values in it. It implies that, once a restricted value is set, all unauthorised users will have to reach out to authorised ones to modify that resource from that moment on, until the resource is modified back to no longer contain restricted values (if ever).

2. **VAP makes exception to authorised the operation based on the old state of the resource** – Users inherit privileges from others (and from previous selfs, in case of permission once granted then revoked) in order to be able to modify resources with restricted values. It implies that, once a user is granted the privilege to modify a restricted value (i.e. it's granted authorisation), the privilege cannot ever be revoked; even by deleting the binding that once granted the permission will do nothing to prevent the user from keep using the restricted value in the resource, until the value is reset back to a non restricted one (if ever).

cc @coderabbit

DaliborD45: Thats true, I removed this feature. The decision made is that VAP does not check old state.
I was also looking at the option of checking every key separately if it changed, hence user without permission would not be able to add new .allNamespaces attribute, but it seemed a bit as workaround a not ideal solution.

jasonmadigan (approved):
Tested the guide on kind cluster,API-key and X.509 requests were allowed or denied as expected. revoking either custom permission blocked and subsequent restricted updates while ordinary patch access remained

PR #683, opened 2026-09-03
docs: add user guide for restricting Authorino's outbound HTTP routes

DaliborD45:
### What

Adds a new user guide, **[Restricting the outbound HTTP routes Authorino can reach](docs/user-guides/restrict-outbound-http-routes.md)**, and links it from the user-guides index.

### What the guide covers

Two `ValidatingAdmissionPolicy` (VAP) resources that turn every outbound destination into an **explicit, RBAC-gated 
allowlist**:

### Changes

- `docs/user-guides/restrict-outbound-http-routes.md` (new)
- `docs/user-guides.md` (index entry)

## Summary by CodeRabbit

- **Documentation**
  - Added a comprehensive guide for restricting outbound HTTP routes from Authorino authentication configurations.
  - Documented a unified RBAC allowlist for statically identified destinations, inline Rego `http.send`, and externally sourced OPA policies.
  - Added Kubernetes prerequisites, setup instructions, policy and binding examples, verification scenarios, limitations, and re-validation behaviour for configuration changes.
  - Updated the user-guide index to highlight RBAC requirements for outbound HTTP access.

### DIFF
--- a/docs/user-guides/restrict-outbound-http-routes.md
+++ b/docs/user-guides/restrict-outbound-http-routes.md
@@ -9,12 +9,14 @@ host reachable from the pod and have Authorino make the request on their behalf.
 This creates a possible situation where attacker can create malicious AuthConfig
 in order to reach cloud metadata endpoints, internal-only services, or to exfiltrate data.
 
-The two ValidatingAdmissionPolicies below close that gap. They turn every
+The three ValidatingAdmissionPolicies below close that gap. They turn every
 outbound destination into an **explicit, RBAC-gated allowlist**: an `AuthConfig`
-may only name a host that the requesting subject has been granted access to, and
+may only name a host that the requesting subject has been granted access to,
 inline Rego may only use `http.send` if the subject has been granted a dedicated
-permission. You hand those permissions only to the subjects that need them to do
-their job.
+permission, and an OPA policy may only be loaded from an external source if the
+subject has been granted a dedicated permission (its Rego is fetched at runtime
+and cannot be scanned for `http.send` at admission time). You hand those
+permissions only to the subjects that need them to do their job.
 
 The policies:
 
@@ -23,13 +25,24 @@ The policies:
 | `authorino-restrict-http-route` | `authconfigs` | any outbound endpoint whose hostname the subject has not been granted (JWT `jwksUrl` / `issuerUrl`, OAuth2 introspection `endpoint`, UserInfo `userInfoUrl`, UMA `endpoint`, metadata/callback `http.url` + `http.oauth2.tokenUrl`, OPA `opa.externalPolicy` URL, SpiceDB `endpoint`) | `access` on `http-resource/<hostname>` |
 | `authorino-restrict-http-route` | `authconfigs` | endpoints whose host cannot be statically verified: a dynamic `urlExpression`, or a URL with a templated `{...}` hostname | *(none — always denied; use a literal hostname instead)* |
 | `authorino-deny-rego-httpsend` | `authconfigs` | inline OPA/Rego (`spec.authorization.*.opa.rego`) that uses the `http.send` builtin | `use` on `authconfig-httpsend` |
+| `authorino-deny-external-opa` | `authconfigs` | OPA policies loaded from an external source (`spec.authorization.*.opa.externalPolicy`), whose Rego is fetched at runtime and cannot be scanned for `http.send` at admission time | `use` on `authconfig-external-opa` |
 
+
+## Prerequisites
+
+> **Important**
+>
+> These manifests require **Kubernetes v1.30 or newer**. They use
+> `ValidatingAdmissionPolicy` via the stable `admissionregistration.k8s.io/v1`
+> API, which is only available from v1.30 (where the feature graduated to GA),
+> along with the CEL `authorizer` library the policies rely on. On older
+> clusters these resources will not apply.
 
 ## 1. Create the Roles
 
 Create one `ClusterRole` per host you want to be able to allow (the RBAC resource
 name is `http-resource/<hostname>`), plus one `ClusterRole` for the inline-Rego
-`http.send` permission.
+`http.send` permission and one for loading OPA policies from an external source.
 
 ```bash
 kubectl apply -f - <<'EOF'
@@ -50,15 +63,24 @@ rules:
   - apiGroups: ["authorino.kuadrant.io"]
     resources: ["authconfig-httpsend"]
     verbs: ["use"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: authorino-external-opa
+rules:
+  - apiGroups: ["authorino.kuadrant.io"]
+    resources: ["authconfig-external-opa"]
+    verbs: ["use"]
 EOF
 ```
 
 ## 2. Grant the access to the restricted destinations
 
-Grant the host-access and `http.send` permissions to your own ServiceAccounts and
-Users. Use the RoleBindings below as a template. Replace the placeholders
-(`<sa-name>`, `<namespace-of-sa>`, `<authconfig-namespace>`) with the appropriate
-values.
+Grant the host-access, `http.send`, and external-OPA permissions to your own
+ServiceAccounts and Users. Use the RoleBindings below as a template. Replace the
+placeholders (`<sa-name>`, `<namespace-of-sa>`, `<authconfig-namespace>`) with the
+appropriate values.
 
 
 
@@ -91,12 +113,26 @@ subjects:
   - kind: ServiceAccount
     name: <sa-name>
     namespace: <namespace-of-sa>
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: rb-external-opa
+  namespace: <authconfig-namespace>
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: authorino-external-opa
+subjects:
+  - kind: ServiceAccount
+    name: <sa-name>
+    namespace: <namespace-of-sa>
 EOF
 ```
 
 ## 3. Create the ValidatingAdmissionPolicies (VAPs)
 
-Apply the two policies below.
+Apply the three policies below.
 
 ```bash
 kubectl apply -f - <<'EOF'
@@ -253,12 +289,52 @@ metadata:
 spec:
   policyName: authorino-deny-rego-httpsend
   validationActions: ["Deny"]
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: authorino-deny-external-opa
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["authorino.kuadrant.io"]
+        apiVersions: ["v1beta3"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["authconfigs"]
+  matchConditions:
+    - name: user-not-external-opa-exempt
+      expression: >-
+        !authorizer.group("authorino.kuadrant.io")
+        .resource("authconfig-external-opa")
+        .namespace(object.metadata.namespace)
+        .check("use")
+        .allowed()
+  variables:
+    - name: usesExternalOpa
+      expression: >-
+        has(object.spec.authorization)
+        && object.spec.authorization.exists(k,
+        has(object.spec.authorization[k].opa)
+        && has(object.spec.authorization[k].opa.externalPolicy))
+  validations:
+    - expression: "!variables.usesExternalOpa"
+      reason: Forbidden
+      message: "OPA policies loaded from an external source (spec.authorization[*].opa.externalPolicy) are not allowed: the Rego is fetched at runtime and cannot be scanned for the 'http.send' builtin at admission time (SSRF). Use an inline 'rego' policy, which is scanned, or ask an admin for the 'authconfig-external-opa' role."
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: authorino-deny-external-opa
+spec:
+  policyName: authorino-deny-external-opa
+  validationActions: ["Deny"]
 EOF
 ```
 
 > **Warning**
 >
-> Unlike a policy that only restricts *newly enabling* a field, both policies
+> Unlike a policy that only restricts *newly enabling* a field, all three policies
 > re-validate the **entire object** on every `CREATE` **and** `UPDATE`, with no
 > comparison to the previous version:
 >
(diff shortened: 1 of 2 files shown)


docs/user-guides/restrict-outbound-http-routes.md
```diff
@@ -0,0 +1,490 @@
...
+        variables.httpEndpoints.filter(e, has(e.oauth2) && has(e.oauth2.tokenUrl) && e.oauth2.tokenUrl != "").map(e, e.oauth2.tokenUrl)
+    - name: jwksUrls
+      expression: >-
+        has(object.spec.authentication)
+        ? object.spec.authentication.map(k, object.spec.authentication[k]).filter(a, has(a.jwt) && has(a.jwt.jwksUrl) && a.jwt.jwksUrl != "").map(a, a.jwt.jwksUrl)
+        : []
+    - name: issuerUrls
+      expression: >-
+        has(object.spec.authentication)
+        ? object.spec.authentication.map(k, object.spec.authentication[k]).filter(a, has(a.jwt) && has(a.jwt.issuerUrl) && a.jwt.issuerUrl != "").map(a, a.jwt.issuerUrl)
```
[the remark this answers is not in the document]
DaliborD45: But the premise of this attack is that admin will grant an attacker a role with a hostname in the whitelist that will contain this malicious url.
Isnt this highly unlikely ?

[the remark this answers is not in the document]
DaliborD45: @guicassolato Do we want to address this issue? At first glance, it seems a bit overengineered to me.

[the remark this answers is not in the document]
guicassolato: I agree. Authorino's trust is primarily on the IdP. The consecutive requests it has to make to fetch the key sets borrow (by transitivity) from the trust that the IdP has put on those other hosts. If we did impose restrictions on the secondary hosts, that could occasionally invert the trust relationship while it motivates the IdP to limit what goes in its own discovery docs because of what's considered valid by the client.

Nothing of this is a problem for an IdP like Keycloak that by default will automatically match the secondary hosts referred by the discovery docs to the primary one used to fetch the doc in the first place. While this might be understood as a counterexample to my second point (about inverting the trust relationship), it reinforces your perception that the concerns raised by @coderabbit go a little beyond the purpose of this PR, which is limited to documentation.

To this final point – the scope of the VAP is limited to the control plane and the requests to the API server. Authorising hosts that can only be known after this phase – as in a background follow-up process of the reconciliation or already in the data plane – would imply a significant redesign of this feature. Instead of enforcing the policy at the level of the API server, Authorino's code would have to be modified to do that (e.g., from a middleware for its internal HTTP client). I therefore conclude @coderabbit's suggestion should be out of scope for this PR.

docs/user-guides/restrict-outbound-http-routes.md
```diff
@@ -0,0 +1,611 @@
...
+### Updates are re-checked, not just creates
+
+Because the policies match `UPDATE` as well as `CREATE`, and they validate the
+whole object each time, a subject that lacks a host's role cannot edit an
+`AuthConfig` that references that host — even to change an unrelated field. Using
+the `route-allowed-1` `AuthConfig` created above (which points at
+`keycloak.example.com`):
+
+```bash
+# Change an unrelated field (the host) while keeping the same issuer,
```
guicassolato, line 549: Can we please modify this example to updating some other field of the AuthConfig (e.g.: `jwt.ttl`)? Using `hosts` may generate some confusion given the work here is to protect against untrusted _hosts_ used in the AuthConfig, even if they are completely unrelated.

guicassolato (dismissed):
We're definitely in the right track. The example works as expected. I just think we could simplify things out a little bit. Here are a few suggestions:

1. I think we could reduce to a single VAP, named something like `trusted-hostnames-only`. The VAP would have two parts:
    1. authorises setting untrusted hostnames in the AuthConfig, which includes using fields that resolve the hostname dynamically (i.e., templated `url`/`endpoint` fields, `urlExpression`, Regos's `http.send` and `opa.externalPolicy`), and
    2. (if needed) authorises using specific values in the fields that specify a static hostname as value.
  
2. We only need 3 cluster roles IMO:
    - one to exemplify one set of authorised hostnames (`trusted-hostnames-1`);
    - a second one for a second set of authorised hostnames (`trusted-hostnames-2`, with some overlap with `trusted-hostnames-1`); and
    - a third one to allow everything (`untrusted-hostnames`).

3. We can establish the following convention for the roles:
    - for the `trusted-hostnames-*` roles: `resources=authconfigs/<hostname>`, `verb=set-hostname`
    - for the `untrusted-hostnames` role: `resources=authconfigs`, `verb=set-untrusted-hostname`

4. Although the current validation against RBAC, based on counting denied hosts (must be equal to 0 to allow), works, I wonder if it wouldn't be simpler and easier to read/explain if instead we did `requestedHostnames.all(h, authorizer...(h).allowed())`.

After (1) through (4), the VAP validation becomes something like: `authorizer.requestResource().check('untrusted-hostnames').allowed() || requestedHostnames.all(h, authorizer.resource('authconfigs').subresource(h).check('set-hostname').allowed())`

Please let me know what you think.

docs/user-guides/restrict-outbound-http-routes.md
DaliborD45: > authorises setting untrusted hostnames in the AuthConfig, which includes using fields that resolve the  
> hostname dynamically (i.e., templated url/endpoint fields, urlExpression, Regos's http.send and 
> opa.externalPolicy) 

I am not sure, how we want to evaluate the urls dynamically. Do we want to support all levels of dynamic resolution ? For example when user types *.keycloak.example.com, do we want to allow a.b.keycloak.example.com, or just a.keycloak.example.com ? 
Also I am not really experienced with CEL, so I am not sure how i should tackle this problem, since I am not finding any good solution

guicassolato:
(Replying out of the thread since the question is not related to the update operation example.)

> I am not sure, how we want to evaluate the urls dynamically. Do we want to support all levels of dynamic resolution ? For example when user types *.keycloak.example.com, do we want to allow a.b.keycloak.example.com, or just a.keycloak.example.com ? Also I am not really experienced with CEL, so I am not sure how i should tackle this problem, since I am not finding any good solution

I didn't mean authorise the _actual_ hostnames that are resolved dynamically when using those fields/features (e.g.: URL templates), but authorise _using the fields/features_, that result in resolving hostnames dynamically.

So it's actually a simpler problem. The question the VAP should answer is: _"Can this user use any hostname in an AuthConfig or only specific authorised hostnames?"_

If the answer to this question is: "Yes, the user can use _any_ hostnames", then this includes hostnames that are resolved dynamically. Therefore, the user can also use the fields and features (e.g., templating) that cause resolving hostnames dynamically, let alone statically defined hostnames without requiring more specific permission for it.

If the answer is: "No, the user cannot use _any_ hostnames, only hostnames _specifically_ authorised", then the user should not be allowed to use fields and features that resolve hostnames dynamically, but only fields with static hostnames specified, and these static hostnames specified in the AuthConfig must be authorised per RBAC rule.

DaliborD45:
Oh okay, now I understand perfectly. I am thinking, about two options. Have one role, where resources: "*" would mean to resolve any array OR have two separate roles, where in first we define static hostnames and other role automatically allows user to resolve any hostname.

docs/user-guides/restrict-outbound-http-routes.md
```diff
@@ -0,0 +1,600 @@
...
+kubectl apply -f - <<'EOF'
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: authorino-trusted-hostnames
+rules:
+  - apiGroups: ["authorino.kuadrant.io"]
+    resources:
+      - "authconfigs/keycloak.example.com"
+      - "authconfigs/userinfo.example.com"
```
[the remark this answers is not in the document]
DaliborD45: Where it says that it is empty ?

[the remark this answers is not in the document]
DaliborD45: You have probably outdated context, no occurence in the file

docs/user-guides/restrict-outbound-http-routes.md
```diff
@@ -0,0 +1,599 @@
...
+        .check("use")
+        .allowed()
+  variables:
+    - name: usesHttpSend
+      expression: >-
+        has(object.spec.authorization)
+        && object.spec.authorization.exists(k,
+        has(object.spec.authorization[k].opa)
+        && has(object.spec.authorization[k].opa.rego)
+        && object.spec.authorization[k].opa.rego.contains("http.send"))
```
jasonmadigan, line 167: `contains("http.send")` misses the bracket spelling of the same builtin. `resp := http["send"]({...})` gets through for a subject with no hostname grant and OPA runs it. something like

```suggestion
        && object.spec.authorization[k].opa.rego.matches(r'http(\.send|\[\s*.send.\s*\])'))
```

catches both spellings. since this is a syntactic denylist, the stricter option is to treat any inline `opa.rego` as needing the unrestricted role, which is the only way to make check 1 airtight.

docs/user-guides/restrict-outbound-http-routes.md
```diff
@@ -0,0 +1,599 @@
...
+
+- **Exact hostname match only.** There is no wildcard form. `*.example.com` is not
+  supported; list each host.
+- **Case-sensitive.** `KEYCLOAK.example.com` does not match a grant for
+  `keycloak.example.com`. It is denied.
+- **Hostname only.** A grant does not restrict the port, path or scheme on that
+  host.
+- **Dynamic URLs need the bypass role.** `urlExpression` and templated
+  `{selector}` hostnames are always denied by check 3, even in the safe shape
+  where the host is fixed and only the path varies. There is no separate role for
```
jasonmadigan: check 3 only looks at the hostname, so a templated path on a granted host is admitted: `https://keycloak.example.com/{request.path}` passes for a subject granted `keycloak.example.com`. maybe: "templated `{selector}` hostnames are denied by check 3; a template that only varies the path on a granted host is allowed." or deny any `{` in URLs if that was the intent.

docs/user-guides/restrict-outbound-http-routes.md
```diff
@@ -0,0 +1,599 @@
...
+
+Try to create resources that break the rules. Run these as a regular user (one
+without the roles) and each should be rejected.
+
+> **Note**
+>
+> Do not run these as a cluster administrator. Anything with wildcard access
+> (`verbs: ["*"]` on `resources: ["*"]`) — which cluster admins have — satisfies
+> the `access` and `use` checks, so the request would be allowed and a real
+> outbound route enabled. Use an ordinary user (or `--as=<unauthorized-subject>`)
```
jasonmadigan, line 321: worth adding: `--as` needs `impersonate` on users, and the test subjects need ordinary create/update on authconfigs, otherwise the Forbidden comes from RBAC not the policy. also a freshly created VAP takes a few seconds to take effect; the first attempts right after `kubectl apply` of the policy can be admitted.

docs/user-guides/restrict-outbound-http-routes.md
```diff
@@ -0,0 +1,599 @@
...
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["authorino.kuadrant.io"]
+        apiVersions: ["*"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["authconfigs"]
+  matchConditions:
+    - name: is-unrestricted-user
```
jasonmadigan: the condition is true for subjects who are not unrestricted; `is-restricted-user` would read as it evaluates.

jasonmadigan (changes requested):

docs/user-guides/restrict-outbound-http-routes.md
DaliborD45:
```
  - name: hasUnparseableUrl
    expression: >-
      variables.allUrls.exists(u, !isURL(u) || url(u).getHostname() == "" || url(u).getHostname().contains("{"))
      || variables.spicedbHosts.exists(h, h == "" || h.contains("{"))

```
This should not admit templated path, since it checks whether it contains "{", than later if I find unparseableUrl, the VAP rejects the request

docs/user-guides/restrict-outbound-http-routes.md
DaliborD45: The third solution that I came up with is that we can block every rego.http, since send is the only attribute that http object provides.
So we can write something like this:
`object.spec.authorization[k].opa.rego.matches(r'(^|[^A-Za-z0-9_.])http[.\[]'))`
respectively full check
```
 expression: >-
        has(object.spec.authorization)
        && object.spec.authorization.exists(k,
        has(object.spec.authorization[k].opa)
        && has(object.spec.authorization[k].opa.rego)
        && object.spec.authorization[k].opa.rego.matches(r'(^|[^A-Za-z0-9_.])http[.\[]'))
```
WDYT ?

docs/user-guides/restrict-outbound-http-routes.md
DaliborD45: Oh okay I now understand what do you mean.
Well this is correct behavior, isnt it ? We want to block only hostnames, the path can be templated, as it cannot cause any harm. I will update the doc though

jasonmadigan: yea agreed sorry, and the reworded limitations match what it actually does

docs/user-guides/restrict-outbound-http-routes.md
jasonmadigan: works for me!

jasonmadigan: sorry, will resolve when you push this. in pretty good shape then!

DaliborD45: @jasonmadigan I have resolved this, so everything should be good now

jasonmadigan (dismissed):

jasonmadigan:
some CI lint stuff

DaliborD45:
@jasonmadigan fixed, i guess I need new approval though

jasonmadigan (approved):

docs/user-guides/restrict-outbound-http-routes.md
```diff
@@ -0,0 +1,611 @@
...
+      - "authconfigs/userinfo.example.com"
+    verbs: ["set-hostname"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: authorino-unrestricted-hostnames
+rules:
+  - apiGroups: ["authorino.kuadrant.io"]
+    resources: ["unrestricted-hostnames"]
```
guicassolato, line 69: I see we went with `unrestricted-hostnames` for the resource name. Why not using `authconfigs`? Wouldn't it be more consistent with that other documented VAP where the RBAC controls access to `set-apikey-all-namespaces` and `set-x509-all-namespaces` verbs on the `authconfigs` resource?

DaliborD45: Good catch, I overlooked it for some reason. I will fix it in the implementation part, so I dont have to do separate merge for that, or should I just create new PR with this one fix ?

guicassolato: I would lean to a hotfix PR and fix the docs independently.

DaliborD45: created hotfix: #691
Hope everything is correct, but please double-check.
Thanks

PR #684, opened 2026-09-03
chore: ugraded kubernetes version to 1.30.x

DaliborD45:
## What

Upgrade minimal Kubernetes version from 1.28.x to 1.30.x so we can apply VAPs, as they are in GA from version 1.30.0

## Summary by CodeRabbit

- **Compatibility**
  - Kubernetes 1.30 is now the minimum supported version, with Kubernetes 1.35 recommended.
  - Updated test tooling and environments for newer Kubernetes releases.

- **New Features**
  - Added optional limits for HTTP response sizes in authentication and metadata configurations.
  - Improved AuthConfig schema handling for label selectors.

- **Bug Fixes**
  - Added clearer defaults and guidance for credential reference names while preserving backwards compatibility.

- **Documentation**
  - Updated CEL testing, troubleshooting, and platform requirement guidance for current Kubernetes versions.


jasonmadigan (commented):
this looks good to me overall. could we update the platform requirements in docs/getting-started.md too? still says 1.21, and docs.kuadrant.io already imports that page

thomasmaas:
bumping to 1.30 still leaves authorino 2–5 minor versions behind everything else. If the team wants authorino to track OCP 4.20 (the current stable), the target should be v0.33.x (k8s 1.33) to match dns-operator.
Screenshot 2026-09-14 at 20 54 17

DaliborD45:
Yes I was also wondering whether we should not upgrade to higher version. I think upgrading to the versions as authorino-operator would be ideal. 
WDYT ?

DaliborD45:
So what I did is that I upgraded the k8s version to 1.35.x, but kept minimal version to 1.30.x, because VAP is in GAs since 1.30 and thats why I am doing the upgrade in the first place.
However I am not really sure about it, maybe we can also bump MINIMAL version to 1.35.x.
This would also be needed in the authorino-operator.
Any thoughts ?

jasonmadigan:
👀

jasonmadigan (dismissed):
thanks @DaliborD45 !

A small nit but otherwise I think this is fine:

`docs/getting-started.md` still says 1.21 (docs.kuadrant.io imports that page). be good to get in if not too much hassle

DaliborD45:
@jasonmadigan 
I am sorry, but what I wrote in the comment 2 days ago was a bit misleading information, because I forgot to push the changes :D
Can you please have a look on this once again ? If it is actually worse, I can just remove the last commit to the state  where you approved it.
Thanks

jasonmadigan (approved):
thanks @DaliborD45 !

PR #688, opened 2026-09-17
feat(jwt): add `jwt.audiences` to enforce the `aud` claim on both `issuerUrl` and `jwksUrl`

### LINKED ISSUE #664: feat(jwt): validate the aud (audience) claim on authentication.jwt
**Is your feature request related to some problem you are facing? Please describe that problem here.**

JWT authentication (`authentication.jwt`) verifies the token signature and expiry, and; once #657 lands: optionally the `iss` claim via `authentication.jwt.issuer`. It still does not verify the `aud` (audience) claim.

Where several services sit behind the same issuer (the usual Keycloak/Auth0/Entra setup), a token legitimately minted for service A authenticates to service B on signature alone; a cross-service token-replay gap. Authorino already validates audiences for `kubernetesTokenReview` (`audiences`) but not for JWTs, its main authentication method. RFC 7519 §4.1.3 and RFC 8725 (JWT BCP) §3.9 both call for the recipient to reject a token not intended for it.

**Describe the solution you'd like**

Add an optional `authentication.jwt.audiences []string`. When set, a token authenticates only if its `aud` claim intersects the configured list ("any match"; the same semantics as `kubernetesTokenReview.audiences` and Envoy's `jwt_authn`). When omitted (the default), the claim is not verified and existing AuthConfigs behave identically.

- Enforced at the authentication phase, so a wrong-audience token returns `401` (and honours `response.unauthenticated`), not `403`.
- Applies to both the `issuerUrl` and `jwksUrl` paths. go-oidc already parses `aud` (JSON string or array) into `[]string`, so this is a post-verify intersection check.

```yaml
spec:
  authentication:
    "keycloak-users":
      jwt:
        issuerUrl: https://sso.corp.example/realms/prod
        issuer: https://sso.corp.example/realms/prod
        audiences:
        - my-api.io
```

This is the natural next field after #657's `issuer`: same shape (opt-in claim enforcement, default off, enforced in the verifier). It completes the mandatory-claim set (`exp`/`nbf`, `iss`, `aud`) and reaches parity with Envoy `jwt_authn`, Istio `RequestAuthentication`, and Authorino's own `kubernetesTokenReview`.

**Describe alternatives you've considered**

A pattern-matching / OPA authorization rule (e.g. `"my-api" in auth.identity.aud`) can approximate it, but it isn't the same behavior:

- It runs in the authorization phase, so a wrong-audience token returns `403` instead of `401` and bypasses `response.unauthenticated`.
- The CEL is a correctness trap: `aud` may be a JSON string *or* an array, so a naive rule silently passes on one of the two shapes.
- It composes badly across multiple identity sources (needs per-source `when` conditioning to avoid rejecting non-JWT identities).

**Additional context**

- Follow-up to #657. Touches the same files (`api/v1beta3/auth_config_types.go`, `pkg/evaluators/identity/jwt.go`, `controllers/auth_config_controller.go`), so implementation should land after #657 merges to avoid conflicts.
- Not blocked by the open review discussion on #657 (the `newIdPMockWithIssuer` test-mock signature): that is internal to #657's tests; the audience tests would simply follow whichever mock pattern #657 settles on.

jasonmadigan:
Closes #664. Follow-up to #657.

`authentication.jwt` verifies the signature, expiry and, optionally, the issuer, but not the audience. Behind a shared issuer, a token minted for one service authenticates to every other service on signature alone (RFC 7519 section 4.1.3, RFC 8725 section 3.9).

This adds the optional `authentication.jwt.audiences` list. When set, a verified token must carry at least one of the values in its `aud` claim (any match, as Kubernetes TokenReview and Envoy `jwt_authn` do); otherwise authentication fails with `401` and `response.unauthenticated` applies. Both JSON shapes of `aud` (string, array) are handled. When omitted, the claim is not verified and existing AuthConfigs are unchanged; the reconciler logs the unverified default, as it does for `issuer`.

The check lives in `JWTAuthentication`, after verification, so it applies to both `issuerUrl` and `jwksUrl`; go-oidc's client ID check matches a single value only. The CRD rejects an empty list and empty values.

Verification:

- Unit tests: the audience matrix through a mocked verifier (string, array, absent, empty and null `aud`; any match; unset default; static error text), wiring through the real `issuerUrl` and `jwksUrl` verifiers, and the reconciler log pair.
- `make test-cel`: valid list, empty value rejected, and `audiences: []` rejected via an unstructured create (the typed client omits an empty slice).
- Live: real Keycloak 26.4 tokens through Envoy against a kind cluster, `issuerUrl` and `jwksUrl` each with audiences unset / matching / non-matching / any-match / non-matching with `response.unauthenticated` (`418`): 10/10, plus the two schema rejections through `kubectl apply`.
- `tests/v1beta3/authconfig.yaml` sets `audiences: [account]` on the Keycloak source, so the e2e suite covers the field.

Review order: `api/v1beta3/auth_config_types.go` (field and markers), `pkg/evaluators/identity/jwt.go` (`checkAudience`, constructor), `controllers/auth_config_controller.go` (wiring and log), then the tests, `docs/features.md` and the JWT user guide. The generated CRD and manifests follow from the type.

### Verification steps

```sh
make local-setup FF=1 DEPLOY_KEYCLOAK=1

```

```sh
ACCESS_TOKEN=$(curl http://keycloak.127.0.0.1.nip.io:8080/realms/kuadrant/protocol/openid-connect/token -s -d 'grant_type=password' -d 'client_id=demo' -d 'username=john' -d 'password=p' -d 'scope=openid' | jq -r .access_token)
```

The token's `aud` is `["realm-management", "account"]`. None of the cases below set `issuer`, so only the audience varies.

**audiences = ∅ ⇒ 200**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
EOF
```

```sh
# 200
```

**audiences = [talker-api] (not in the token) ⇒ 401**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        audiences:
        - talker-api
EOF
```

```sh
# 401
# www-authenticate: Bearer realm="keycloak-kuadrant-realm"
# x-ext-auth-reason: jwt: token audience not accepted
```

**audiences = [talker-api, account] (any match) ⇒ 200**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        audiences:
        - talker-api
        - account
EOF
```

```sh
# 200
```

**jwksUrl, audiences = [account] ⇒ 200; audiences = [talker-api] ⇒ 401**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        jwksUrl: http://keycloak.default.svc.cluster.local:8080/realms/kuadrant/protocol/openid-connect/certs
        audiences:
        - account
EOF
```

```sh
# 200
```

Change `- account` to `- talker-api`, apply again:

```sh
# 401
```

**audiences = [talker-api] with a custom unauthenticated response ⇒ 418** (the rejection is an authentication failure, not authorization)

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        audiences:
        - talker-api
  response:
    unauthenticated:
      code: 418
      message:
        value: wrong audience
EOF
```

```sh
# 418
# x-ext-auth-reason: wrong audience
```

**audiences = [] or [""] ⇒ rejected by the API server**

```sh
kubectl apply -f -<<EOF
kind: AuthConfig
metadata:
  name: talker-api-protection
spec:
  hosts:
  - talker-api.127.0.0.1.nip.io
  authentication:
    "keycloak-kuadrant-realm":
      jwt:
        audiences: []
EOF
# The AuthConfig "talker-api-protection" is invalid: spec.authentication.keycloak-kuadrant-realm.jwt.audiences: Invalid value: 0: spec.authentication.keycloak-kuadrant-realm.jwt.audiences in body should have at least 1 items
```

With `audiences: [""]` instead:

```sh
# The AuthConfig "talker-api-protection" is invalid: spec.authentication.keycloak-kuadrant-realm.jwt.audiences[0]: Invalid value: "": spec.authentication.keycloak-kuadrant-realm.jwt.audiences[0] in body should be at least 1 chars long
```

## Summary by CodeRabbit

- **New Features**
  - JWT authentication can validate the token’s `aud` claim against one or more configured audiences.
  - Supports audience claims supplied as either a single value or an array.
  - Tokens without a matching audience are rejected during authentication with an unauthenticated response.
  - Audience validation is optional; existing behaviour is preserved when no audiences are configured.
  - Audience configuration requires at least one non-empty value.

- **Documentation**
  - Added configuration guidance and examples for JWT audience validation.


pkg/evaluators/identity/jwt.go
```diff
@@ -71,6 +87,32 @@ func (j *JWTAuthentication) Call(pipeline auth.AuthPipeline, ctx gocontext.Conte
...
+		}
+		if slices.Contains(j.audiences, aud) {
+			return nil
+		}
+		present = append(present, aud)
+	}
+	if len(present) == 0 {
+		return errors.New(msg_jwtAudienceMissing)
+	}
+	return fmt.Errorf(msg_jwtAudienceNotAccepted, strings.Join(present, ", "))
```
guicassolato: The contract is clear. At least one audience present in the JWT must be allow-listed in the policy. If it doesn't find any (i.e., empty overlap between the two sets), all audiences from the JWT are "present" (or should I say "absent"?), therefore the JWT is altogether deemed invalid.

A caller can easily know (and should know) what audiences are in the JWT, therefore I wonder about the need to echo back in the error message this info that is effectively arbitrary user input and thus also an opening for Cross-Site Scripting (XSS) attacks perhaps?

On the other hand, telling back to the user which audiences _are_ allowed would be more useful for the user to take action upon it. However, it may also be considered leaking crucial information for the security of the system. In theory, the parties involved knowing that beforehand is part of the contract and Authorino is probably better off not disclosing the info by default. (I could change my mind if an RFC says otherwise – which I really don't remember in the case of OAuth2, but happens to be the case of Section 11 of RFC 9110 and challenge headers.)

Maybe a static error message is good enough to start with?

jasonmadigan: yeah agreed. now static: jwt: token has no audience (aud) claim and jwt: token audience not accepted, nothing from the token or the allow-list echoed

guicassolato (approved):
LGTM. Thanks @jasonmadigan!

